PR #1, opened 2021-06-03
first draft of gene implementation

dake0795:
First draft of gene implementation into pyrokinetics. 

To be fixed / checked:

**1. Namelist reading and writing:**
- Manually need to change any instances of &species to &species_1, &species_2 etc to convert GENE -> pyro object.
- Manually need to change any instances of $species_1, &species_2 to &species to run GENE output. 

**2. Normalisation**
- To be checked. I believe the normalisation in GENE should be consistent with that of CGYRO when using Miller geometry (Physics of Plasmas 25, 042505 (2018); https://doi.org/10.1063/1.5018741). 

**3. Grids**
- I haven't thought very much about how to sensibly map the grids between the different coordinate systems in GS2/CGYRO vs GENE.

### DIFF
--- a/pyrokinetics/gene.py
+++ b/pyrokinetics/gene.py
@@ -279,7 +279,6 @@ class GENE(GKCode):
 
             species_data.temp = species_data.temp / te
             species_data.dens = species_data.dens / ne
-            print(species_data.temp, species_data.dens, species_data.a_ln, species_data.a_lt)
             pressure += species_data.temp * species_data.dens
             a_lp += species_data.temp * species_data.dens * (species_data.a_lt + species_data.a_ln)
 
--- a/pyrokinetics/gene.py
+++ b/pyrokinetics/gene.py
@@ -127,7 +127,7 @@ class GENE(GKCode):
 
         # Kinetic data
         local_species = pyro.local_species
-        gene_input['box']['nspec'] = local_species.nspec
+        gene_input['box']['n_spec'] = local_species.nspec
 
         pyro_gene_species = self.pyro_to_gene_species()
 
@@ -177,6 +177,19 @@ class GENE(GKCode):
 
         else:
             gene_input['general']['nonlinear'] = False
+
+
+
+        gene_input['box']['nky0'] = numerics.nky
+        gene_input['box']['kymin'] = numerics.ky
+        
+
+        if numerics.kx != 0.0: 
+            gene_input['box']['lx'] = 2 * pi / numerics.kx
+ 
+        gene_input['box']['nz0'] = numerics.ntheta
+        gene_input['box']['nv0'] = 2 * numerics.nenergy
+        gene_input['box']['nw0'] = numerics.npitch 
 
         # Currently forces NL sims to have nperiod = 1
         gene_nml = f90nml.Namelist(gene_input)
@@ -306,9 +319,9 @@ class GENE(GKCode):
             'Rgeo' : ['geometry', 'major_r'],
             'q' : ['geometry', 'q0'],
             'kappa' : ['geometry', 'kappa'],
-            's_kappa' : ['geometry', 's_kappa'],     #checked
+            's_kappa' : ['geometry', 's_kappa'],    
             'delta' : ['geometry', 'delta'],
-            's_delta' : ['geometry', 's_delta'],    #checked
+            's_delta' : ['geometry', 's_delta'],    
             'shat' : ['geometry', 'shat'],  
             'shift' : ['geometry', 'drr'],
             }
@@ -356,14 +369,34 @@ class GENE(GKCode):
         
         numerics = Numerics()
 
-        numerics['nky'] = gene['box']['nky0']
-        numerics['nkx'] = gene['box']['nx0']
-        numerics['nky'] = 1
-        numerics['ky'] = gene['box']['kymin'] 
-        numerics['kx'] = 1.0
+        numerics.nky = gene['box']['nky0']
+        numerics.nkx = gene['box']['nx0']
+        numerics.ky = gene['box']['kymin'] 
+        numerics.kx = 2 * pi / gene['box']['lx']
 
 
         # Velocity grid
+
+
+        try:
+            numerics.ntheta = gene['box']['nz0']
+        except KeyError:
+            numerics.ntheta = 24
+
+        try:
+            numerics.nenergy = 0.5 * gene['box']['nv0']
+        except KeyError:
+            numerics.nenergy = 8
+
+        try: 
+            numerics.npitch = gene['box']['nw0']
+        except KeyError: 
+            numerics.npitch = 16 
+
+
+
+
+
 
         try:
             nl_mode = gene['nonlinear']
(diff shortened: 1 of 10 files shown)


pyrokinetics/gene.py
```diff
@@ -0,0 +1,370 @@
...
+
+        # Load GENE with species data
+        self.load_local_species(pyro, gene)
+
+
+ # Need species to set up beta_prime
+
+        if pyro.local_geometry_type == 'Miller':
+            if pyro.local_geometry['Bunit'] is not None:
+                pyro.local_geometry['beta_prime'] = - pyro.local_species['a_lp'] / pyro.local_geometry['Bunit'] ** 2
```
bpatel2107, line 84: @dake0795 I this needs to be 

```suggestion
            if pyro.local_geometry['B0'] is not None:
                pyro.local_geometry['beta_prime'] = - pyro.local_species['a_lp'] / pyro.local_geometry['B0'] ** 2
```

pyrokinetics/templates/input.gene
```diff
@@ -0,0 +1,161 @@
...
+s_kappa  =   0.24000000
+s_delta  =   0.40000000
+s_zeta   =  -0.20000000E-01
+drR      =  -0.10000000
+sign_Ip_CW =      -1
+sign_Bt_CW =      -1
+/
+
+&species_1
+name   = 'ions'
```
bpatel2107, line 130: @dake0795 I would leave the template as just ```&species``` so we can test out the file  modifier

pyrokinetics/gene.py
```diff
@@ -0,0 +1,370 @@
...
+            # Reference B field
+            bref = miller['B0']
+
+            shat = miller['shat']
+            # Assign Miller values to input file
+            pyro_gene_miller = self.pyro_to_gene_miller()
+
+            # GENE uses definitions consistent with Miller. 
+            for key, val in pyro_gene_miller.items():
+                gene_input[val[0]][val[1]] = miller[key]
```
bpatel2107, line 125: Is it worth sticking to one kind of variable setting where possible. i.e. instead of ```miller['shat']```, we use ```miller.shat```?

Though there are cases like this loop when you iterate over keys where this is not really possible

pyrokinetics/templates/input.gene
dake0795: I then ran into problems with reading the 2+ namelists into a pyro object?

pyrokinetics/gene.py
dake0795: I think where possible it's probably a good idea.

pyrokinetics/templates/input.gene
bpatel2107: So I think the template should be an accurate GENE input file that you could actually run, as at the moment it isn't.

A user should be able to take a working GENE input file and use it. So we should modify the template so it is a "working" input file and then add methods like

```python
def modify_gene_input_for_reading():
    """ 
    Code that changes &species to &species_1
    """

def modify_gene_input_for_writing():
    """
    Code that changes &species_1 back to &species
    """

```
    
which would get called when calling ```read_gk_file``` and ```write_gk_file``` respectively

dake0795: I see. I'll take a look at doing that this afternoon if I can make some time.

pyrokinetics/gene.py
```diff
@@ -0,0 +1,368 @@
...
+            if species_data.z == -1:
+                name = 'electron'
+                te = species_data.temp
+                ne = species_data.dens
+            else:
+                ion_count += 1
+                name = f'ion{ion_count}'
+
+            pressure += species_data.temp * species_data.dens
+            a_lp += species_data.temp * species_data.dens * (species_data.a_lt + species_data.a_ln)
```
bpatel2107, line 273: @dake0795 Look at commit ab1362b

Need to force the normalisation to electron temp/dens

pyrokinetics/gene.py
```diff
@@ -269,6 +269,17 @@ def load_local_species(self, pyro, gene):
...
+        a_lp = 0.0
+
+        # Normalise to pyrokinetics normalisations and calculate total pressure gradient
+        for name in local_species.names:
+
+            species_data = local_species[name]
+
+            species_data.temp = species_data.temp / te
+            species_data.dens = species_data.dens / ne
+            print(species_data.temp, species_data.dens, species_data.a_ln, species_data.a_lt)
```
bpatel2107, line 282: We shoudl remove unnecessary print statement
```suggestion
            
```

pyrokinetics/gene.py
```diff
@@ -0,0 +1,379 @@
...
+
+        miller.kappri = miller.s_kappa*miller.kappa / miller.rho
+        miller.tri = np.arcsin(miller.delta)
+
+        # Get beta normalised to R_major(in case R_geo != R_major)
+        beta = gene['general']['beta'] * (miller.Rmaj/miller.Rgeo)**2
+
+        # Assume pref*8pi*1e-7 = 1.0
+        if beta != 0.0:
+            miller.B0 = np.sqrt(1.0/beta**0.5)
```
bpatel2107, line 218: This was a mistake on my front, double square root

```suggestion
            miller.B0 = np.sqrt(1.0/beta)
```

pyrokinetics/gene.py
```diff
@@ -0,0 +1,379 @@
...
+        # Need shear for map theta0 to kx
+        # shat = pyro.miller['shat']
+        # Fourier space grid
+        # Linear simulation
+        
+        numerics = Numerics()
+
+        numerics['nky'] = gene['box']['nky0']
+        numerics['nkx'] = gene['box']['nx0']
+        numerics['nky'] = 1
```
bpatel2107, line 362: Any reason this is overwritten?

pyrokinetics/gene.py
```diff
@@ -0,0 +1,379 @@
...
+        numerics['ky'] = gene['box']['kymin'] 
+        numerics['kx'] = 1.0
```
bpatel2107, line 364: Why is it set to 1 as default?

pyrokinetics/gene.py
```diff
@@ -0,0 +1,379 @@
...
+
+
+        # Velocity grid
```
bpatel2107, line 367: Do we need to set nz0, nw0 and nv0 somehow? 

I would imagine 
```python
nz0 = numerics.ntheta
nv0 = 2 * numerics.nenergy
nw0 = numerics.npitch
```
 would be reasonable? Thoughts?

dake0795: Probably a good idea to have a sensible default. These seem reasonable to me. 

Was hard to think of something here on the basis that I can't personally imagine not just fiddling with these myself, but no harm in having something to start with.

bpatel2107: Agreed, we can always change it as we moved forward

dake0795: Now done.

bpatel2107:
Update the README and I'll merge!

PR #5, opened 2021-06-10
Feature/allow gk code to be set when writing

d7919:
Quality of life feature to make it easier to transiently set the gk_code type for output.

### DIFF
--- a/pyrokinetics/pyro.py
+++ b/pyrokinetics/pyro.py
@@ -199,7 +199,8 @@ class Pyro:
     def write_gk_file(self,
                       file_name,
                       template_file=None,
-                      directory='.'
+                      directory='.',
+                      gk_code=None
                       ):
         """ 
         Writes single GK input file to file_name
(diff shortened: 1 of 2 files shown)


pyrokinetics/pyro.py
```diff
@@ -222,6 +229,9 @@ def write_gk_file(self,
 
         self.gk_code.write(self, file_name, directory=self.run_directory)
 
+        # Ensure that gk_code is unchanged on exit
+        self.gk_code = original_gk_code
```
bpatel2107, line 233: @d7919  Not sure if it is necessary to save the original gk_code. No data is lost by changing the code as the namelist data for each code is saved under a separate object.

For example the if we load in a GS2 input file and the namelist data with all the flags is saved under pyro.gs2_input. When changing code the pyro.gs2_input remains. Thoughts?

d7919: I prefer function calls to not have side effects if possible. The original approach makes it clear that the user is changing the gk_code value of the pyro instance whilst this function call doesn't make it obvious so I thought we should revert to the original type.

bpatel2107: Fair enough, I think I agree with that

bpatel2107: This currently won't quite work as you can only set gk_code with a string. We need modify the gk_code.setter to handle gk_code objects as inputs

bpatel2107: That would work but we definitely need to re-vist how setting gk_code/gk_type works

bpatel2107 (approved):

PR #7, opened 2021-06-11
Fixed CGYRO defn of s_delta

bpatel2107:
Removed ```sqrt(1 - \delta**2)``` from miller.s_delta for CGYRO defn of s_delta


dake0795 (approved):
Looks great

PR #18, opened 2021-09-17
Gradient change during scan isn't reflected in local_species.a_lp

bpatel2107:
When doing a scan whilst changing the gradient a_lp is changing but that isn't updated in the object. Now a_lp is re-calculated whenever the file is written for CGYRO.

Should we implement a setter/getter that will automatically update a_lp whenever a dens/temp/gradient is changed?

### DIFF
--- a/pyrokinetics/cgyro.py
+++ b/pyrokinetics/cgyro.py
@@ -158,6 +158,7 @@ class CGYRO(GKCode):
             if pyro.local_geometry_type == 'Miller':
                 if miller.Bunit is not None:
                     beta = 1.0 / miller.Bunit ** 2
+                    local_species.calculate_a_lp()
                     beta_prime_scale = - miller.beta_prime / (
                             local_species.a_lp * beta * (miller.Bunit / miller.B0) ** 2)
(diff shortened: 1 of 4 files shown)


d7919:
Yes I think a setter/getter for things we want to keep consistent.

ZedThree:
Python has a built-in thing for this: [`property`](https://docs.python.org/3/library/functions.html#property)

Very useful for invariants like this. You can then use the property like a normal variable.

This also seems like a good thing to write a test for.

bpatel2107:
I should be able to implement this with the property decorator.

Not sure about the testing as I haven't really done much of that. Maybe we can use pytest?

ZedThree:
The best way forward is probably for me to implement some really basic tests and get them running in Github Actions, and that will make it much easier for you to add more tests. We can have a call and work on getting some more complicated ones implemented together.

bpatel2107:
Updated code to automatically update a_lp and pressure when changing density/temperature/gradients. Also added a `add_species` method to localspecies and updated the different code + from kinetics

PR #19, opened 2021-09-22
Pyroscan now loads output data

bpatel2107:
Pyroscan object now make a copy of the base pyro object for each permutation of the values.

Loads in linear output of growth rate, frequency and eigenfunctions into gk_output attribute

Saves the base_directory for the runs and set it recursively for each copy

Temporary fix for a_lp changes but will be handled better in #18

### DIFF
--- a/pyrokinetics/local_species.py
+++ b/pyrokinetics/local_species.py
@@ -167,18 +167,3 @@ class LocalSpecies(CleverDict):
 
         self['pressure'] = pressure
         self['a_lp'] = a_lp
-
-    def calculate_a_lp(self):
-        """
-        Calculate a_lp for species
-        Returns
-        -------
-        self['a_lp']
-        """
-
-        a_lp = 0.0
-        for name in self.names:
-            species = self[name]
-            a_lp += species['temp'] * species['dens'] * (species['a_lt'] + species['a_ln'])
-
-        self['a_lp'] = a_lp
--- a/pyrokinetics/local_species.py
+++ b/pyrokinetics/local_species.py
@@ -167,3 +167,18 @@ class LocalSpecies(CleverDict):
 
         self['pressure'] = pressure
         self['a_lp'] = a_lp
+
+    def calculate_a_lp(self):
+        """
+        Calculate a_lp for species
+        Returns
+        -------
+        self['a_lp']
+        """
+
+        a_lp = 0.0
+        for name in self.names:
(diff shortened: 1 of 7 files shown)


pyrokinetics/local_species.py
```diff
@@ -167,3 +167,18 @@ def from_kinetics(self,
 
         self['pressure'] = pressure
         self['a_lp'] = a_lp
+
+    def calculate_a_lp(self):
```
d7919: This looks like a separate change?

pyrokinetics/pyroscan.py
```diff
@@ -178,17 +205,91 @@ def load_default_parameter_keys(self):
...
+        import xarray as xr
+        from cleverdict import CleverDict
+
+        # Set up output as CleverDict
+        self.gk_output = CleverDict()
+
+        # xarray DataSet to store data
+        ds = xr.Dataset(self.parameter_dict)
+
+        if not self.base_pyro.numerics.nonlinear:
```
d7919, line 227: It might be worth allowing this method to take a function as an argument -- this can then detail what is loaded, what analysis is done etc. It could default to a method which implements this linear analysis, for example.

pyrokinetics/local_species.py
bpatel2107: You think it's worth taking it out and leave the changes for #18 ? I left it in so changes in the gradients in CGYRO are done accurately. This doesn't impact GENE/GS2

d7919: I don't mind really just flagging that this change looked like it wasn't directly a part of the main feature being introduced here.

pyrokinetics/pyroscan.py
bpatel2107: So something along the lines of?
```python
def load_gk_output(self, output_func, data_name):
     for pyro in all_pyros:
           output_data = output_func(pyro)

     ds['data_name'] = output_data
```
where

```python
def output_func(pyro):
     output_data = get_output_from_pyro
  
     return output_data
```

d7919: Something like that, although potentially returning more data.

bpatel2107: I can leave it like this for now and the make a separate issue for this as it would be nice to have this for a general Pyro object rather than just a Pyroscan for now

PR #24, opened 2021-10-01
Remove Bunit from miller object and define everything from B0 instead

bpatel2107:
Currently there are two fields being stored in the Miller object `B0` and `Bunit`. They are both used when defining `beta` in CGYRO but if I do a scan in `B0`, that won't change `Bunit` which will also mess up the setting of the equilibrium pressure gradient `BETA_STAR_SCALE` in CGYRO.

This defines everything in terms of `B0` and stores the ratio of `Bunit/B0`, from which everything is calculated. This allows for more consistent scan in `B0` (`beta`) without having to change two things at once

### DIFF
--- a/pyrokinetics/cgyro.py
+++ b/pyrokinetics/cgyro.py
@@ -84,7 +84,7 @@ class CGYRO(GKCode):
         beta_prime_scale = cgyro.get('BETA_STAR_SCALE', 1.0)
 
         if pyro.local_geometry_type == 'Miller':
-            if pyro.local_geometry.Bunit is not None:
+            if pyro.local_geometry.B0 is not None:
                 pyro.local_geometry.beta_prime = - pyro.local_species.a_lp / pyro.local_geometry.B0 ** 2 * \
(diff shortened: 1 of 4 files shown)


pyrokinetics/cgyro.py
```diff
@@ -151,16 +151,16 @@ def write(self, pyro, file_name, directory='.'):
 
             # Find BETA_STAR_SCALE from beta and p_prime
             if pyro.local_geometry_type == 'Miller':
-                beta_prime_scale = - miller.beta_prime / (local_species.a_lp * beta * (miller.Bunit / miller.B0) ** 2)
+                beta_prime_scale = - miller.beta_prime / (local_species.a_lp * beta * miller.bunit_over_b0 ** 2)
```
bpatel2107, line 154: Here is an example where changing only `B0` won't give you the behavior desired to only change `beta`  as `BETA_STAR_SCALE` will also change

dake0795 (approved):
Looks good—an excellent way around the problem.

PR #25, opened 2021-10-18
CI: Auto-apply black

ZedThree:
Applies [black](https://black.readthedocs.io/en/stable/) to whole project, and adds a Github workflow to auto-apply black on push


bpatel2107:
Does it automatically apply the formatting to the local files too, or just the remote files?

ZedThree:
When you push to Github, it will run an Action which runs `black`. If that results in any changes, it will automatically commit them. If that happens, you'll need to pull the new commit. It won't change anything locally.

So you can either completely rely on the Github Action to sort out the formatting for you, or run `black` locally yourself. I have a plugin for my editor that runs `black` whenever I save the file.

PR #26, opened 2021-10-19
Fix flake8 warnings and add linting to CI

ZedThree:
- Fix lots of flake8 warnings, mostly about `from foo import *`
- Add Github Action to run flake8 automatically
- Add flake8 config to allow longer lines (160)
- Add file to ignore certain commits when running `git blame` (locally, not sure it works on github yet)
  - Run `git config blame.ignoreRevsFile .git-blame-ignore-revs` to do this automatically
  - Useful to ignore the changes from applying `black` to everything!

@bpatel2107 You probably want to have a look at 8db3f6e in particular

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,5 @@
+# Apply black
+dfa5cac7268ada9249d96262dd3bbd8a04866f3e
+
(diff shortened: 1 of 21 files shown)


bpatel2107 (approved):
Looks good

PR #29, opened 2021-10-19
Add some unit and integrated tests

ZedThree:
- Add `pyrokinetics.template_dir` which points to the bundled `templates/` directory
- Add unit tests for a chunk of the code (about ~20%)
- Add integrated test: run `example_SCENE` and read in the generated input files and check they're the same as the original object
  - Takes coverage up to 55%, but doesn't give much specificity

@bpatel2107 Can you look at `test_roundtrip.py` especially? There are some fields that aren't the same between e.g. gs2 and cgyro.  For example, `beta_prime` and `B0` in the geometry, and `nu` in the species. Perhaps you expect these to be different for some reason, but I was expecting them to be identical

### DIFF
--- a/pyrokinetics/equilibrium.py
+++ b/pyrokinetics/equilibrium.py
@@ -156,18 +156,12 @@ class Equilibrium:
 
         self.R_major = InterpolatedUnivariateSpline(psi_n, R_major)
 
-    def get_flux_surface(
-        self,
-        psi_n=None,
-    ):
+    def get_flux_surface(self, psi_n):
 
         import matplotlib as mpl
 
         mpl.use("Agg")
         import matplotlib.pyplot as plt
-
-        if psi_n is None:
-            raise ValueError("get_flux_surface needs a psi_n")
 
         # Generate 2D mesh of normalised psi
         psi_2d = np.transpose(self.psi_RZ(self.R, self.Z))
--- a/pyrokinetics/equilibrium.py
+++ b/pyrokinetics/equilibrium.py
@@ -93,9 +93,8 @@ class Equilibrium:
         from scipy.interpolate import InterpolatedUnivariateSpline, RectBivariateSpline
         from numpy import linspace
 
-        f = open(self.eq_file)
-        gdata = _geqdsk.read(f)
-        f.close()
+        with open(self.eq_file) as f:
+            gdata = _geqdsk.read(f)
 
         # Assign gdata to Equilibriun object
 
--- a/pyrokinetics/examples/example.py
+++ b/pyrokinetics/examples/example.py
@@ -1,13 +1,8 @@
-from pyrokinetics import Pyro
-import os
+from pyrokinetics import Pyro, template_dir
 
-home = os.environ["HOME"]
-base = home + "/pyrokinetics/pyrokinetics/templates/"
-
-gs2input = base + "input.gs2"
-eqfile = base + "test.geqdsk"
-pyro = Pyro(eq_file=eqfile, eq_type="GEQDSK")
-
+gs2input = template_dir / "input.gs2"
+eqfile = template_dir / "test.geqdsk"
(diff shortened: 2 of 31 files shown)


pyrokinetics/equilibrium.py
```diff
@@ -157,19 +156,13 @@ def read_geqdsk(self, psi_n_lcfs=1.0):
...
+    def get_flux_surface(self, psi_n):
 
         import matplotlib as mpl
 
         mpl.use("Agg")
         import matplotlib.pyplot as plt
 
-        if psi_n is None:
-            raise ValueError("get_flux_surface needs a psi_n")
-
```
bpatel2107, line 172: Is this being checked somewhere else?

pyrokinetics/examples/example.py
```diff
@@ -1,13 +1,8 @@
...
-
-home = os.environ["HOME"]
-base = home + "/pyrokinetics/pyrokinetics/templates/"
-
-gs2input = base + "input.gs2"
-eqfile = base + "test.geqdsk"
-pyro = Pyro(eq_file=eqfile, eq_type="GEQDSK")
+from pyrokinetics import Pyro, template_dir
 
+gs2input = template_dir / "input.gs2"
```
bpatel2107, line 3: Is this `/` OS independant? I'm guessing its the same as path.join()?

pyrokinetics/equilibrium.py
ZedThree: Yes, by removing the default argument, python will error if `psi_n` isn't passed, something like this:

```
In [2]: get_flux_surface()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-a054aaf2cb71> in <module>
----> 1 get_flux_surface()

TypeError: get_flux_surface() missing 1 required positional argument: 'psi_n'
```

More generally, `pyrokinetics` uses a lot of default arguments in places when they are actually required.

pyrokinetics/examples/example.py
ZedThree: Yup, `template_dir` is a `pathlib.Path` which is in the standard library. I think elsewhere you use a third-party `path` module, which does the same thing

bpatel2107:
All of the issues with `test_roundtrip` are due to the fact that the local input files don't have reference values of n, T and B, but the full SCENE equilibrium/kinetic profile does.

`B0` can't be identical here. When we have the full equilibrium and kinetics profiles we can know exactly what `B0` is as it is defined as `B0 = f/ Rmaj`. So `beta` is then defined as `beta = 8pi * 1e-7  * ne * Te/ B0^2`

However, with a local  input file we don't have the reference values so I just define it as `B0 = sqrt(1/beta)`. It doesn't really matter but to convert from CGYRO to GS2 we need `bunit_over_b0` to be the same. At the moment these are close but not quite the same.

`bunit_over_b0` isn't exactly the same here as it requires an integration along `theta`, which is defined in the full equilibrium file but not in the local fit so I just set it to a `theta = np.linspace(0, 2pi, 128)`. When I force it to use the same `theta` values, regardless of the equilibrium values, both `bunit_over_b0` and `beta_prime` pass the test. I'll push this change now.

With `nu`, there's a bug with the definition. Here we should have

`nu` is defined as 
image

where 
image


At the moment, the code is using the individual species density and temperature rather than the electron density and temperature. Fixing this makes the test pass

bpatel2107:
You could add GENE to the `example_SCENE` and `test_roundtrip` which would increase the coverage

ZedThree:
I wonder if it might be worth adding some metadata in comments to the input files, that way we could store things like reference values, or even the path to the equilibrium file. Not something to do here, but worth thinking about for the future!

I've added GENE to the example and test, and in doing so uncovered and fixed a bug!

bpatel2107:
Looks good!

PR #30, opened 2021-10-20
Modernise and automate packaging

ZedThree:
- Prefer `setup.cfg` over `setup.py`
  - Main advantage is that it's declarative data, and not code that needs to be run
- Use `setuptools_scm` to get the version relative to the latest git tag
- Add Github Action to automate publishing to PyPI

@bpatel2107 You'll need to get a [PyPI API token](https://pypi.org/help/#apitoken) and add it to this [repo's secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets) with the name `PYPI_DEPLOYMENT_TOKEN`.

Once you've done that, whenever you create a new Github Release, it will automatically build the package and upload it to PyPI as a new version.

When you make a release, make sure the tag is of the form `v[0-9]+.[0-9]+.[0-9]+`: this will auto-bump the version number. If there are changes since the last release, we'll get a version number like: [`0.1.dev232+gacbe225.d20211020`](https://github.com/pypa/setuptools_scm#default-versioning-scheme).


ZedThree:
There's a couple of small fixes for the docs here too, because I was checking they build with the packaging changes

bpatel2107:
I've added the PyPi token so is it worth testing if this works now?

ZedThree:
Good idea, maybe name it something like `v0.2.0-alpha`?

ZedThree:
Ta-da!: https://pypi.org/project/pyrokinetics/0.2.0a0

bpatel2107:
This is amazing thanks! Might need you to walk me through how releasing it works

ZedThree:
I'm free the rest of the day if you want a 5 minute zoom call

PR #36, opened 2021-10-21
Added flag for linear vs nonlinear fluxes and fixed error in non-zona…

bpatel2107:
…l flux component

Add check for fluxes in output file and set to 0 if not there.

Divide non-zonal flux components by 2 due to reality condition

### DIFF
--- a/pyrokinetics/gs2.py
+++ b/pyrokinetics/gs2.py
@@ -671,6 +671,7 @@ class GS2(GKCode):
         """
 
         import netCDF4 as nc
+        import logging
 
         gk_output = pyro.gk_output
         data = gk_output.data
@@ -693,8 +694,7 @@ class GS2(GKCode):
         )
 
         if f'{field_keys[0]}_{moment_keys[0]}' not in netcdf_data.variables.keys():
-            print('Flux data not written, setting fluxes to NaN')
-            fluxes[:] = np.nan
+            logging.warning('Flux data not written to netCDF file, setting fluxes to 0')
 
         else:
             for ifield, field in enumerate(field_keys):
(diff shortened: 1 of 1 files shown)


pyrokinetics/gs2.py
```diff
@@ -683,20 +683,36 @@ def load_fluxes(self, pyro):
...
+                for imoment, moment in enumerate(moment_keys):
+                    key = f"{field}_{moment}"
+
+                    if pyro.numerics.nonlinear:
+                        # Sum over kx
+                        flux = np.sum(netcdf_data[key], axis=-1)
+                        flux = np.moveaxis(flux, [1, 2, 0], [0, 1, 2])
+
+                        # Divide non-zonal components by 2 due to reality condition
+                        flux[:, 1:, :] *= 0.5
```
d7919, line 710: This factor looks correct for consistency with GS2.

pyrokinetics/gs2.py
```diff
@@ -683,20 +683,36 @@ def load_fluxes(self, pyro):
...
 
         fluxes = np.empty(
             (gk_output.nspecies, 3, gk_output.nfield, gk_output.nky, gk_output.ntime)
         )
 
-        for ifield, field in enumerate(field_keys):
-            for imoment, moment in enumerate(moment_keys):
-                key = f"{field}_{moment}"
+        if f"{field_keys[0]}_{moment_keys[0]}" not in netcdf_data.variables.keys():
+            print("Flux data not written, setting fluxes to NaN")
```
bpatel2107: Actually might not be a good idea to print this out if we're doing a large scan... thoughts? @d7919 @ZedThree

d7919: Might be worth looking at using a logging library so that warnings, debug messages can be controlled across the code easily etc.

ZedThree: There's one in the standard library: https://docs.python.org/3/library/logging.html

bpatel2107: Added in a warning

PR #37, opened 2021-10-21
Write json file describing pyro object and read in tolerance and flux…

bpatel2107:
…es from scan

Writes a json file to the PyroScan directory so it is possible to read in as 

```python
pyroscan = PyroScan(pyro, /path/to/pyroscan.json)
```

Also reads in fluxes and growth rate tolerance in a scan

### DIFF
--- a/pyrokinetics/pyroscan.py
+++ b/pyrokinetics/pyroscan.py
@@ -7,6 +7,7 @@ import operator
 import copy
 import json
 
+
 class PyroScan:
     """
     Creates a dictionary of pyro objects in pyro_dict
@@ -28,7 +29,7 @@ class PyroScan:
         file_name=None,
         base_directory=".",
         load_default_parameter_keys=True,
-        pyroscan_json=None
+        pyroscan_json=None,
     ):
 
         # Dictionary of parameters and values
@@ -87,12 +88,14 @@ class PyroScan:
                 self.base_directory = self.pyroscan_json["base_directory"]
 
         else:
-            self.pyroscan_json = {"value_fmt": self.value_fmt,
-                                  "value_separator": self.value_separator,
-                                  "parameter_separator": self.parameter_separator,
-                                  "parameter_dict": self.parameter_dict,
-                                  "file_name": self.file_name,
-                                  "base_directory": self.base_directory}
+            self.pyroscan_json = {
+                "value_fmt": self.value_fmt,
+                "value_separator": self.value_separator,
+                "parameter_separator": self.parameter_separator,
+                "parameter_dict": self.parameter_dict,
+                "file_name": self.file_name,
+                "base_directory": self.base_directory,
+            }
 
         pyro_dict = {}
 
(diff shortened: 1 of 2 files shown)


pyrokinetics/pyroscan.py
```diff
@@ -20,14 +21,15 @@ class PyroScan:
     def __init__(
         self,
         pyro,
-        parameter_dict=None,
+        parameter_dict={},
```
ZedThree: Mutable default arguments are [not a great idea](https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument). Better to leave the default as `None` and do

```python
if parameter_dict is None:
  self.parameter_dict = {}
else:
  self.parameter_dict = parameter_dict

# Alternative, but perhaps less explicit:
self.parameter_dict = parameter_dict or {}
```

Looks like there's also currently a bug, because the argument is ignored on L36: `self.parameter_dict = {}`

pyrokinetics/pyroscan.py
```diff
@@ -68,47 +73,66 @@ def __init__(
...
+        pyro_dict = {}
+
+        # Get len of values for each parameter
+        self.value_size = [len(value) for value in self.parameter_dict.values()]
+
+        # Outer product of input dictionaries - could get very large
+        self.outer_product = list(
+            dict(zip(self.parameter_dict, x))
+            for x in product(*self.parameter_dict.values())
+        )
```
ZedThree: An alternative is to leave this as a generator:

```suggestion
        # Outer product of input dictionaries - could get very large
        self.outer_product = (
            dict(zip(self.parameter_dict, x))
            for x in product(*self.parameter_dict.values())
        )
```

This will then generate the outer product on the fly, rather than storing the whole thing.

pyrokinetics/pyroscan.py
```diff
@@ -129,6 +153,10 @@ def write(self, file_name=None, base_directory=None, template_file=None):
                 for run_dir in self.pyro_dict.keys()
             ]
 
+        # Dump json file with pyroscan data
+        json_file = os.path.join(self.base_directory, "pyroscan.json")
+        json.dump(self.pyroscan_json, open(json_file, "w+"))
```
ZedThree: This won't `close` the file which might result in the contents not getting written to disk. It's better to always use `with open` to ensure the file gets closed

```suggestion
        with open(json_file, "w+") as f:
            json.dump(self.pyroscan_json, f)
```

bpatel2107: Also for the loading of the json?

ZedThree:
You should try writing a test for this! :) Something like:

1. create a `Pyro` object
2. write to JSON
3. read from JSON into new `Pyro` object
4. check new object is identical to old one (or at least equivalent for the bits we write out)

pyrokinetics/pyroscan.py
ZedThree: Yep! I didn't spot that at first

pyrokinetics/pyroscan.py
```diff
@@ -68,47 +73,66 @@ def __init__(
...
-                single_run_name = ""
-                # Param value for each run written accordingly
-                for param, value in run.items():
-                    single_run_name += (
-                        f"{param}{self.value_separator}{value:{self.value_fmt}}"
-                    )
+        # Load in pyroscan json if there
+        if pyroscan_json is not None:
+            if isinstance(pyroscan_json, str):
+                self.pyroscan_json = json.load(open(pyroscan_json))
```
bpatel2107:
```suggestion
                with open(pyroscan_json) as f:
                    self.pyroscan_json = json.load(f)

                
```

bpatel2107:
@ZedThree  Added a unit test, thoughts?

ZedThree (approved):

PR #41, opened 2021-11-10
Data in each Pyro in PyroScan now a deepcopy

bpatel2107:
Before all the data in each pyro object in pyroscan was actually pointing to the same value so now Numerics, LocalSpecies and LocalGeometry now have `__deepcopy__` ensuring when data is set, it is set independently

### DIFF
--- a/pyrokinetics/local_geometry.py
+++ b/pyrokinetics/local_geometry.py
@@ -39,7 +39,7 @@ class LocalGeometry(CleverDict):
 
         pass
 
-    def __deepcopy__(self, memodict={}):
+    def __deepcopy__(self, memodict):
         """
         Allows for deepcopy of a LocalGeometry object
 
--- a/pyrokinetics/local_geometry.py
+++ b/pyrokinetics/local_geometry.py
@@ -38,3 +38,19 @@ class LocalGeometry(CleverDict):
         """
 
         pass
+
+    def __deepcopy__(self, memodict={}):
+        """
+        Allows for deepcopy of a LocalGeometry object
+
+        Returns
+        -------
+        Copy of LocalGeometry object
+        """
+
+        new_localgeometry = LocalGeometry()
+
+        for key, value in self.items():
+            setattr(new_localgeometry, key, value)
+
+        return new_localgeometry
--- a/pyrokinetics/local_species.py
+++ b/pyrokinetics/local_species.py
@@ -233,10 +233,22 @@ class LocalSpecies(CleverDict):
 
         # Add in each species
         for name in self["names"]:
-            dict_keys = {"name": "name", "mass": "mass", "z": "z", "nu": "nu", "vel": "vel",
-                         "a_lv": "a_lv", "_dens": "dens", "_temp":"temp",
-                         "_a_ln": "a_ln", "_a_lt": "a_lt"}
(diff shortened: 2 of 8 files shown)


pyrokinetics/local_geometry.py
```diff
@@ -38,3 +38,19 @@ def load_from_gk_file(self, pyro, gk_code=None):
         """
 
         pass
+
+    def __deepcopy__(self, memodict={}):
```
ZedThree: Avoid mutable default arguments

```suggestion
    def __deepcopy__(self, memodict):
```

pyrokinetics/local_species.py
```diff
@@ -214,6 +214,46 @@ def add_species(self, name, species_data):
         self.names.append(name)
         self.update_pressure()
 
+    def __deepcopy__(self, memodict={}):
```
ZedThree:
```suggestion
    def __deepcopy__(self, memodict):
```

pyrokinetics/numerics.py
```diff
@@ -40,3 +40,19 @@ def default(self):
         }
 
         super(Numerics, self).__init__(_data_dict)
+
+    def __deepcopy__(self, memodict={}):
```
ZedThree:
```suggestion
    def __deepcopy__(self, memodict):
```

pyrokinetics/pyroscan.py
```diff
@@ -267,28 +267,35 @@ def load_gk_output(self):
...
+                        .isel(time=-1)
+                        .sum(dim="ky")
+                        .drop_vars(["time"])
+                    )
+
+                    pyro.gk_code.get_growth_rate_tolerance(pyro, time_range=0.95)
+                    growth_rate_tolerance.append(
+                        pyro.gk_output.data["growth_rate_tolerance"]
+                    )
+                except (FileNotFoundError, OSError) as e:
```
ZedThree: Flake8 is complaining that `e` isn't used. Looks like there might a bug in the action as it cuts off the message.

```suggestion
                except (FileNotFoundError, OSError):
```

pyrokinetics/pyroscan.py
```diff
@@ -395,7 +402,9 @@ def set_in_dict(data_dict, map_list, value):
     """
     Sets item in dict given location as a list of string
     """
-    get_from_dict(data_dict, map_list[:-1])[map_list[-1]] = value
+    from copy import deepcopy, copy
```
ZedThree: This one is complaining about `copy.copy` being imported and not used

```suggestion
    from copy import deepcopy
```

ZedThree (changes requested):
It would be good to write a test for the `__deepcopy__` methods, and the `PyroScan.set_in_dict`.

A test for `deepcopy` would be to create an instance of the object, make a deepcopy and `assert` that the `id()` of each component was different between the original and the copy.

bpatel2107:
Do you think it is worth making a deepcopy of all the different parts of a Pyro object? At the moment I don't make a copy of the Equilibrium of Kinetics object

ZedThree:
Do you think they'll change between scans? If so, then it's probably a good idea

bpatel2107:
Probably not as I expect only local changes during scans. 

Its not really built to handle different equilibrium in scans yet, though I guess that could be useful in the future  though it should be relatively simple to add. I don't think we're near any sort of memory issue either.

PR #44, opened 2022-01-05
Switched CGYRO frequency sign to match pyrokinetics convention

bpatel2107:
CGYRO uses positive to denote the electron diagmagnetic direction whereas pyrokinetics (along with GS2 and GENE) use negative to denote electron direction. Bugfix switches CGYRO to this convention.

### DIFF
--- a/pyrokinetics/cgyro.py
+++ b/pyrokinetics/cgyro.py
@@ -702,7 +702,7 @@ class CGYRO(GKCode):
                     )
 
                     complex_field = (
-                        field_data[0, :, :, :, :] + 1j * field_data[1, :, :, :, :]
+                        field_data[0, :, :, :, :] - 1j * field_data[1, :, :, :, :]
                     )
 
                     fields[ifield, :, :, :, :] = np.swapaxes(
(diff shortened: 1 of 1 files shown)


pyrokinetics/cgyro.py
```diff
@@ -702,7 +702,7 @@ def load_fields(self, pyro):
                     )
 
                     complex_field = (
-                        field_data[0, :, :, :, :] + 1j * field_data[1, :, :, :, :]
```
d7919, line 705: Just to check I understand, is this sign change coming from essentially undoing `exp(-i*omega*t)` -> `exp(i*omega*t)` ? Do we need to worry about relative sign of exb shear as well?

d7919: Is it worth a comment to explain the sign, which otherwise looks a little surprising.

bpatel2107: Yep that's basically what I've done. Can add a comment clarifying this

bpatel2107: Not sure about ExB, but that's not implemented in pyrokinetics yet so will keep in mind?

d7919 (approved):

PR #45, opened 2022-01-07
Re-scale beta_prime if using different Bref

bpatel2107:
Scale's beta_prime if using a different reference field in GS2 input file.


d7919 (approved):
Is it worth thinking about adding some test cases where pyro generates linear runs, we run with the codes and then compare outputs. Quite a lot of work perhaps, but pretty nice for testing things like this. Could presumably cache the builds of the GK codes so don't need to build those in CI each time.

ZedThree:
I think that's definitely a good idea, but potentially something to do in a superproject? Some codes have private/gatekeeper-ed repos, so the superproject would either need to use a subset or be private.

PR #47, opened 2022-01-21
Do line integral of path in psi contour fit to find closed loop for s…

bpatel2107:
…urface.

Reset MPL backend rather than overwriting

For certain files, the wrong contour was being chosen when reading in GEQDSK files like below

The yellow shows the lines of constant psi in the GEQDSK file and the dash line shows the contour it chose for the surface. Clearly it has chosen the wrong contour.

 
image

I’ve added a fix that does a line integral along the contours and looks for the smallest absolute value (which a closed loop should be). Doing so finds the right contour
 
image

Also ensured matplotlib backend doesn't get changed during contour fitting

### DIFF
--- a/pyrokinetics/equilibrium.py
+++ b/pyrokinetics/equilibrium.py
@@ -161,6 +161,9 @@ class Equilibrium:
         import matplotlib as mpl
         import matplotlib.pyplot as plt
 
+        if psi_n > 1.0 or psi_n < 0.0:
+            raise ValueError("You must have 0.0 <= psi_n <= 1.0")
+
         # Generate 2D mesh of normalised psi
         psi_2d = np.transpose(self.psi_RZ(self.R, self.Z))
 
--- a/pyrokinetics/equilibrium.py
+++ b/pyrokinetics/equilibrium.py
@@ -107,8 +107,8 @@ class Equilibrium:
         self.psi_axis = gdata["simagx"]
         self.psi_bdry = gdata["sibdry"]
 
-        self.R_axis = gdata['rmagx']
-        self.Z_axis = gdata['zmagx']
+        self.R_axis = gdata["rmagx"]
+        self.Z_axis = gdata["zmagx"]
 
         psi_n = linspace(0.0, psi_n_lcfs, self.nr)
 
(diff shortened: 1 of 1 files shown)


bpatel2107:
@jfparisi Relevant to you

pyrokinetics/equilibrium.py
```diff
@@ -159,26 +159,40 @@ def read_geqdsk(self, psi_n_lcfs=1.0):
...
+        original_backend = mpl.get_backend()
+        mpl.use("Agg")
+
         con = plt.contour(self.R, self.Z, psin_2d, levels=[0, psi_n])
 
+        mpl.use(original_backend)
+
         paths = con.collections[1].get_paths()
 
         if psi_n == 1.0 and len(paths) == 0:
```
d7919, line 185: Should you also guard for psi_n > 1?

bpatel2107: Good idea, will add that in!

pyrokinetics/equilibrium.py
```diff
@@ -159,26 +159,40 @@ def read_geqdsk(self, psi_n_lcfs=1.0):
...
             raise ValueError(
                 "PsiN=1.0 for LCFS isn't well defined. Try lowering psi_n_lcfs"
             )
 
-        path = paths[np.argmax(len(paths))]
+        # Find smallest path integral to find closed loop
+        loop_integrals = []
+        for path in paths:
```
d7919: Looks good, I do something similar in my step_utils repo but instead look for the path with the lowest distance from rmaxis, zmaxis. If the equilibrium output contains coils on the grid (e.g. a free boundary code) then there can be closed loops around them.

d7919 (approved):

pyrokinetics/equilibrium.py
bpatel2107: I did think about that but I wasn't sure if there could be cases where this might not be true. I guess if we compare the the closest point of each contour to raxis and zaxis, then the smallest one must be the closed loop?
 
Worth changing?

d7919: My code for this is

~~~~
        if len(results) == 1:
            result = results[0]
        else:
            # Otherwise find the path with the least average distance from rmaxis, zmaxis
            distances = [ np.mean( np.sqrt((x.vertices[:,0] - self['rmaxis'])**2 +
                                           (x.vertices[:,1] - self['zmaxis'])**2)) for x in results]
            result = results[np.array(distances).argmin()]
~~~~

where results if your paths.

Probably not very likely to hit currently but it could be an issue with free boundary sources.

bpatel2107: @d7919  Just updated to this method

PR #48, opened 2022-01-21
WIP: Read transp eq

bpatel2107:
Load in equilibrium data from fourier moments in to Miller objects
Select time slice for analysis when loading global equilibrium and kinetic profiles.


pyrokinetics/miller.py
```diff
@@ -154,6 +161,8 @@ class Miller(LocalGeometry):
         Normalised Major radius (Rmajor/a_minor)
     Rgeo : Float
         Normalisd major radius of normalising field (Rreference/a)
+    Z0 : Float
```
d7919, line 164: You've renamed to zmid below, or is z0 distinct from zmid as the next line suggests?

pyrokinetics/equilibrium.py
```diff
@@ -182,10 +352,13 @@ def get_flux_surface(self, psi_n):
...
-            raise ValueError(
-                "PsiN=1.0 for LCFS isn't well defined. Try lowering psi_n_lcfs"
-            )
+        if psi_n == 1.0:
+            if len(paths) == 0:
+                raise ValueError(
+                    "PsiN=1.0 for LCFS isn't well defined. Try lowering psi_n_lcfs"
+                )
+            elif len(paths) > 1:
+                paths = [np.concatenate([path for path in paths])]
```
bpatel2107: This was designed to merge two contours if they didn't form a closed loop. 

This will break the result in #47, need to examine in detail

pyrokinetics/miller.py
bpatel2107: Zmid is different. Z0 = Zmid/aminor

pyrokinetics/miller.py
```diff
@@ -279,6 +288,7 @@ def load_from_eq(self, eq: Equilibrium, psi_n, verbose=False):
 
         self.kappa = kappa
         self.delta = delta
+        self.Z0 = float(Zmid / eq.a_minor)
```
bpatel2107, line 288: @d7919

README.md
```diff
@@ -64,6 +64,7 @@ There also exists the PyroScan object which allows you to make a N-D parameter s
 ## Supports sources of Equilibrium data
 pyrokinetics currently supports
 * GEQDSK
+* TRANSP
```
d7919: If you have documentation that defines the data format is it possible to link to it/give a reference? Helpful to see where later implementation comes from etc.

pyrokinetics/miller.py
bpatel2107: I needed both as getting the regular flux surface requires Zmid. But when I do loop integrals around a flux surface to find q or Bunit I need Zmin/aminor = Z0. 

The distinction is especially important as a local input file may have Zmid/aminor but not Zmid.

Happy to rename if confusing?

pyrokinetics/equilibrium.py
```diff
@@ -159,6 +162,173 @@ def read_geqdsk(self, psi_n_lcfs=1.0):
...
+            try:
+                R_mom_cos[i, :, :] = (
+                    np.cos(i * theta) * data[f"RMC{i:02d}"][itime, :] * 1e-2
+                )
+            except IndexError:
+                break
+            Z_mom_cos[i, :, :] = (
+                np.cos(i * theta) * data[f"YMC{i:02d}"][itime, :] * 1e-2
+            )
+            if i == 0:
```
d7919, line 230: I'm guessing you need this branch as TRANSP decides not to store the RMS0 data as it is automatically zero (i.e. the other branch would calculate the same data but it will error as the key to data is invalid. Worth a comment?

pyrokinetics/equilibrium.py
```diff
@@ -159,6 +162,173 @@ def read_geqdsk(self, psi_n_lcfs=1.0):
...
+            itime = np.argmin(abs(time_cdf - time))
+        else:
+            raise ValueError("time input needs to be float or int")
+
+        ntheta = 256
+        theta = np.linspace(0, 2 * np.pi, ntheta)
+        theta = theta[:, np.newaxis]
+
+        # Calculate flux surfaces from moments up to 17
+        R_mom_cos = np.empty((17, ntheta, nr))
```
d7919: `nmoments = 17`?

README.md
bpatel2107: No clear documentation, this was an educated guess that turned out to be correct. This is the closest thing I could find 
https://w3.pppl.gov/~pshare/help/body_transp_hlp.html#outfile56.html

pyrokinetics/equilibrium.py
```diff
@@ -159,6 +162,173 @@ def read_geqdsk(self, psi_n_lcfs=1.0):
...
+        rbdry = Rsur[:, -1]
+        zbdry = Zsur[:, -1]
+
+        Rmin = min(rbdry)
+        Rmax = max(rbdry)
+        Zmin = min(zbdry)
+        Zmax = max(zbdry)
+
+        Rgrid = np.linspace(Rmin, Rmax, nr)
+        Zgrid = np.linspace(Zmin, Zmax, nr)
```
d7919: Is the Z grid required to have the same size as the R grid?

pyrokinetics/miller.py
d7919: Think it's fine. I just saw you renaming from Z0 to Zmid at the same time you adding docs referring to Z0 so thought I should double check.

README.md
d7919: Thanks, in that case it might be worth including something in the docs about what you're assuming/guessing so that there's a record etc. I did have the GEQDSK documentation somewhere and I'd guess you do as well. Handy to link to as these things can be hard to find.

pyrokinetics/equilibrium.py
bpatel2107: I don't think needs to be the same, only requirements is that `RectBivariateSpline` works

d7919: OK great, in that case you might want to optionally allow user to set nr and nz?

pyrokinetics/equilibrium.py
```diff
@@ -159,6 +162,173 @@ def read_geqdsk(self, psi_n_lcfs=1.0):
...
+        psiflat = np.repeat(psi, ntheta)
+
+        RZflat = np.stack([Rflat, Zflat], -1)
+        RZmesh_flat = np.stack([Rmesh_flat, Zmesh_flat], -1)
+
+        # Interpolate using flat data
+        psiRZ_interp = RBFInterpolator(RZflat, psiflat, kernel="cubic")
+
+        # Map data to new grid and reshape
+        psiRZ_data = np.reshape(psiRZ_interp(RZmesh_flat), np.shape(Rmesh)).T
```
d7919, line 320: What does this do for the psi outside the LCFS? Just wondering if it's possible that this could create "phantom" contours outside the LCFS when looking for flux surfaces.

bpatel2107: It does allow for extrapolation so you could do something like this

```python
    # Interpolate using flat data
    psiRZ_interp = RBFInterpolator(RZflat, psiflat, kernel='cubic')

    # Set up new grid with slightly larger domain
    Rgrid = np.linspace(Rmin*0.9, Rmax*1.1, gq_nr)
    Zgrid = np.linspace(Zmin*1.1, Zmax*1.1, gq_nz)

    rdim = max(Rgrid) - min(Rgrid)
    zdim = max(Zgrid) - min(Zgrid)
    rleft = min(Rgrid)
    zmid = (max(Zgrid) + min(Zgrid)) / 2
    nx = gq_nr
    ny = gq_nz

    Rmesh, Zmesh = np.meshgrid(Rgrid, Zgrid)
    Rmesh_flat = np.ravel(Rmesh)
    Zmesh_flat = np.ravel(Zmesh)
    RZmesh_flat = np.stack([Rmesh_flat, Zmesh_flat], -1)
    
    # Map data to new grid and reshape
    psiRZ_data = np.reshape(psiRZ_interp(RZmesh_flat), np.shape(Rmesh)).T
```

Works quite nicely

pyrokinetics/equilibrium.py
bpatel2107: Removed in https://github.com/pyro-kinetics/pyrokinetics/pull/48/commits/f660b0c8adfc49b4ed6a34a999e80b2070a56e00

pyrokinetics/equilibrium.py
bpatel2107: Added options in https://github.com/pyro-kinetics/pyrokinetics/pull/48/commits/39f1d67f70d0d8b4190c5b2e90b9de012c76c1bf

pyrokinetics/equilibrium.py
bpatel2107: Added options in https://github.com/pyro-kinetics/pyrokinetics/pull/48/commits/39f1d67f70d0d8b4190c5b2e90b9de012c76c1bf

bpatel2107:
Removed redundant parameters in Miller `tri` and `kappri` and set in GS2 using `delta` and `s_kappa`

bpatel2107:
Add a test for comparing a Miller object loaded from a TRANSP GEQDSK and NetCDF in 
https://github.com/pyro-kinetics/pyrokinetics/pull/48/commits/e6528b6a6b193dc03fb5f36658f5b0c4dd171a17

Most parameters are within 5% EXCEPT for `delta`

Output from current test (allowable difference temporarily set to 50%)

```
Miller parameter:Rmaj
NetCDF value: 1.5538211703291978
GEQDSK value: 1.552735107553464
Percentage difference 0.07%

Miller parameter:beta_prime
NetCDF value: -0.4821034386050363
GEQDSK value: -0.4662716110902871
Percentage difference 3.40%

Miller parameter:btccw
NetCDF value: -1
GEQDSK value: -1
Percentage difference 0.00%

Miller parameter:delta
NetCDF value: 0.17894987206898305
GEQDSK value: 0.15729772352884308
Percentage difference 13.77%

Miller parameter:ipccw
NetCDF value: -1
GEQDSK value: -1
Percentage difference 0.00%

Miller parameter:kappa
NetCDF value: 1.4227144285404305
GEQDSK value: 1.4248945620864262
Percentage difference 0.15%

Miller parameter:q
NetCDF value: 1.209671224245032
GEQDSK value: 1.2130183300000001
Percentage difference 0.28%

Miller parameter:rho
NetCDF value: 0.6118436424922641
GEQDSK value: 0.6124072558652733
Percentage difference 0.09%

Miller parameter:s_delta
NetCDF value: 0.17766377919748383
GEQDSK value: 0.18769928144024897
Percentage difference 5.65%

Miller parameter:s_kappa
NetCDF value: 0.04387287614871778
GEQDSK value: 0.043562172657136924
Percentage difference 0.71%

Miller parameter:shat
NetCDF value: 1.1601626541305765
GEQDSK value: 1.156370397320653
Percentage difference 0.33%

Miller parameter:shift
NetCDF value: -0.18781220085307973
GEQDSK value: -0.1877255930917316
Percentage difference 0.05%

Miller parameter:Rgeo
NetCDF value: 1.5538211703291978
GEQDSK value: 1.552735107553464
Percentage difference 0.07%

Miller parameter:bunit_over_b0
NetCDF value: 1.68858022407163
GEQDSK value: 1.6884504529956688
Percentage difference 0.01%
```

The actual surfaces are very close. But when it finds Rupper, it leads to a different value due to the number of points. I tried interpolating but that did make it worse...
image

image

pyrokinetics/equilibrium.py
```diff
@@ -159,6 +162,203 @@ def read_geqdsk(self, psi_n_lcfs=1.0):
...
+
+        import netCDF4 as nc
+        import numpy as np
+        from scipy.interpolate import (
+            RBFInterpolator,
+            InterpolatedUnivariateSpline,
+            RectBivariateSpline,
+        )
+
+        data = nc.Dataset(self.eq_file)
```
ZedThree, line 190: It's best to always open files using a `with` context manager, to ensure they get closed.

pyrokinetics/equilibrium.py
```diff
@@ -159,6 +162,203 @@ def read_geqdsk(self, psi_n_lcfs=1.0):
...
+
+        nradial = len(data["XB"][-1, :])
+
+        time_cdf = data["TIME3"][:]
+
+        if isinstance(time, int):
+            itime = time
+        elif isinstance(time, float):
+            itime = np.argmin(abs(time_cdf - time))
```
ZedThree: This uses `time` in two different ways depending on its type, which feels a little error prone. Instead, perhaps add a `time_index: int = -1` input alongside `time: float = None`, and then:

```python
if time_index != -1 and time is not None:
    raise ValueError("Cannot set both `time` and `time_index`")

if time is not None:
    time_index = np.argmin(np.abs(time_cdf - time))
```

The alternative is make sure the behaviour is very well documented.

bpatel2107: I prefer the explicit variable I think

bpatel2107:
@ZedThree Any thoughts on why pytest fails. Sometimes python 3.7 fails and sometimes its 3.8

ZedThree:
The error is from residuals not being finite when calculating a least squares fit -- it suggests sometimes the data is NaN, or there's a divide by zero somewhere. I'm not sure how that would happen only sometimes. Do you have some random data?

bpatel2107:
No random data so I'm puzzled how this is happening

ZedThree:
Running the tests locally, I noticed that the new `test_compare_transp_cdf_geqdsk` test uses all my cores, so one of the SciPy functions must be using OpenMP. And running like `OMP_NUM_THREADS=2 pytest -vvk transp` I was able to reproduce the failure. I'm guessing that this is the source of the non-determinism. Now to find out why and how to fix it...

ZedThree:
Running the test multiple times within the same session, like:

```python
@pytest.mark.parametrize("run", np.arange(10))
def test_compare_transp_cdf_geqdsk(transp_cdf_equilibrium, transp_gq_equilibrium, run):
```

shows that the tests either all pass or all fail, which I think points at the interpolation.

Setting `OMP_NUM_THREADS` to a square number (4, 9, 16) seems to result in the tests passing, and a non-square (2, 10) number they fail. I'm not sure how rigorous that is -- creating the equilibrium takes between 20s and 1min on my machine, depending on the number of OpenMP threads used.

pyrokinetics/equilibrium.py
```diff
@@ -159,6 +162,210 @@ def read_geqdsk(self, psi_n_lcfs=1.0):
...
+
+        Rflat = np.ravel(Rsur.T)
+        Zflat = np.ravel(Zsur.T)
+        psiflat = np.repeat(psi, ntheta)
+
+        RZflat = np.stack([Rflat, Zflat], -1)
+        RZmesh_flat = np.stack([Rmesh_flat, Zmesh_flat], -1)
+
+        # Interpolate using flat data
+        psiRZ_interp = RBFInterpolator(RZflat, psiflat, kernel="cubic")
```
ZedThree:
```suggestion
        psiRZ_interp = RBFInterpolator(RZflat, psiflat, kernel="cubic", neighbors=5)
```

This seems to fix the intermittent test failures on non-square number of OpenMP threads, with the added bonus of being _significantly_ faster: 40s vs 4s.

By default, the `RBFInterpolator` uses all of the points to construct the radial basis functions. From the [docs](https://docs.scipy.org/doc/scipy//reference/generated/scipy.interpolate.RBFInterpolator.html#scipy.interpolate.RBFInterpolator):

> The memory required to solve for the RBF interpolation coefficients increases quadratically with the number of data points, which can become impractical when interpolating more than about a thousand data points. To overcome memory limitations for large interpolation problems, the neighbors argument can be specified to compute an RBF interpolant for each evaluation point using only the nearest data points.

bpatel2107: Very nice spot there!

PR #52, opened 2022-01-26
Check len con.collections and set path correctly

### LINKED ISSUE #49: Error in geqdsk equilibrium
In trying to generate a local miller equilibrium from the highq0tdotp case on surface psi_N=0.5 I hit the following

>   File "pyro_generate_inputs.py", line 10, in <module>
>     kinetics_type="SCENE",
>   File "/users/dd502/.local/lib/python3.7/site-packages/pyrokinetics/pyro.py", line 53, in __init__
>     self.load_global_eq()
>   File "/users/dd502/.local/lib/python3.7/site-packages/pyrokinetics/pyro.py", line 139, in load_global_eq
>     self.eq = Equilibrium(self.eq_file, self.eq_type, **kwargs)
>   File "/users/dd502/.local/lib/python3.7/site-packages/pyrokinetics/equilibrium.py", line 44, in __init__
>     self.read_geqdsk(**kwargs)
>   File "/users/dd502/.local/lib/python3.7/site-packages/pyrokinetics/equilibrium.py", line 142, in read_geqdsk
>     surface_R, surface_Z = self.get_flux_surface(psi_n=i_psiN)
>   File "/users/dd502/.local/lib/python3.7/site-packages/pyrokinetics/equilibrium.py", line 184, in get_flux_surface
>     paths = con.collections[1].get_paths()
> IndexError: list index out of range

Changing `paths = con.collections[1].get_paths()` to `paths = con.collections[0].get_paths()` appears to allow me to proceed as expected.

bpatel2107:
Fixes #49

### DIFF
--- a/pyrokinetics/equilibrium.py
+++ b/pyrokinetics/equilibrium.py
@@ -182,7 +182,9 @@ class Equilibrium:
 
         # Check if more than one contour has been found
         if len(con.collections) > 1:
-            raise ValueError("More than one contour level found in equilibrium.get_flux_surface")
+            raise ValueError(
+                "More than one contour level found in equilibrium.get_flux_surface"
+            )
 
         paths = con.collections[0].get_paths()
 
--- a/pyrokinetics/equilibrium.py
+++ b/pyrokinetics/equilibrium.py
@@ -180,7 +180,13 @@ class Equilibrium:
 
         mpl.use(original_backend)
 
-        paths = con.collections[1].get_paths()
+        # Check if more than one contour has been found
+        if len(con.collections) > 1:
+            ipath = 1
+        else:
+            ipath = 0
+
+        paths = con.collections[ipath].get_paths()
 
         if psi_n == 1.0 and len(paths) == 0:
             raise ValueError(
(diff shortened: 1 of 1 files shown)


bpatel2107:
@d7919 Can you check this fixes your problem

d7919 (approved):
This would fix my issue although an alternative may be to remove the 0 level from the contour call and then finding more than one would be a red flag?

ZedThree:
It would be good to get a test on this! Adding a test along with a bugfix is a good habit, it makes sure the bug doesn't slip back in some how :)

Also, is there a reason to not just always use the zero-th index? Is that usually an uninteresting level if there are multiple contours?

bpatel2107:
I'm sure I did that for a reason but can't remember for the life of me. I've changed it to just find psi_n and it passes all the tests

bpatel2107:
Not sure what test to have beyond the error raised

ZedThree:
Perhaps add a test case that triggered the original bug and should now work?

bpatel2107:
We would need the equilibrium that @d7919 Used. Though I suspect that given that it is no longer trying to find that psi_n=0 surface it might not have the same issue

PR #53, opened 2022-01-26
Removed Rgeo from miller object and corrected normalisations

### LINKED ISSUE #50: Inconsistency in gs2 normalisation handling?
In testing a round trip converting from a custom gs2 input to cgyro and then back to gs2 I find three parameters change, two in the least significant digit, so basically identical, but one by 20%.

The original input file is setup with R_geo /= rmaj whilst pyrokinetics likes R_geo=rmaj. This changes the location at which the normalising magnetic field is evaluated. Hence I would expect the round trip to 

1. change R_geo and rmaj to be equal
2. scale anything using the normalising magnetic by a factor like R_geo/rmaj

In particular, I'd expect the rescaling to change beta and beta_prime. It seems that beta_prime is changing (this is the parameter which varies a lot) but beta is being held more or less fixed (this is one of the parameters which is changing in the least significant digit). Further, neither of R_geo and rmaj are changed.

As it stands the round trip input either has an incorrect beta_prime (as it has been scaled for a different normalising field)  or beta and R_geo/rmaj are incorrect as they have not been modified to reflect the change in normalising field. 

On a more minor note gs2 can store the normalising parameters in an input namelist. It would be nice if pyrokinetics would write these where possible and could see if these already exist in a source input file and scale as appropriate.

bpatel2107:
Fixes #50


bpatel2107:
@d7919 Can you check to see if this fixes the error with Rgeo normalisations

pyrokinetics/gene.py
```diff
@@ -252,7 +252,7 @@ def load_miller(self, pyro, gene):
         miller.tri = np.arcsin(miller.delta)
 
         # Get beta normalised to R_major(in case R_geo != R_major)
```
d7919: Is this comment still relevant?

bpatel2107: No can remove it

PR #54, opened 2022-01-26
Enhancement/name gk type

### LINKED ISSUE #51: Minor : Pyro init takes gk_type but then we set gk_code to change type
Fairly minor but I always forget that the init call takes gk_type but we change the type of the instance by setting the gk_code property. Could these have the same name?

bpatel2107:
Fixes #51

### DIFF
--- a/pyrokinetics/examples/example.py
+++ b/pyrokinetics/examples/example.py
@@ -2,14 +2,12 @@ from pyrokinetics import Pyro, template_dir
 
 gs2input = template_dir / "input.gs2"
 eqfile = template_dir / "test.geqdsk"
-pyro = Pyro(eq_file=eqfile, eq_type="GEQDSK", local_geometry="Miller", gk_type="GS2")
+pyro = Pyro(gk_code="GS2", gk_file=gs2input)
 
-geo = pyro.geo
+geo = pyro.local_geometry
 
-print(geo["shat"])
 geo["shat"] = 2.3
 
-print(geo["shat"])
 
 flags = {
     "gs2_diagnostics_knobs": {
--- a/pyrokinetics/pyro.py
+++ b/pyrokinetics/pyro.py
@@ -5,7 +5,7 @@ from .equilibrium import Equilibrium
 from .kinetics import Kinetics
 from .gk_output import GKOutput
 from .miller import Miller
-
+import warnings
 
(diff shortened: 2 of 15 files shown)


pyrokinetics/examples/example.py
```diff
@@ -2,14 +2,12 @@
 
 gs2input = template_dir / "input.gs2"
 eqfile = template_dir / "test.geqdsk"
-pyro = Pyro(eq_file=eqfile, eq_type="GEQDSK", local_geometry="Miller", gk_type="GS2")
+pyro = Pyro(gk_code="GS2", gk_file=gs2input)
```
ZedThree, line 5: `eqfile` is now not used?

pyrokinetics/pyro.py
```diff
@@ -31,8 +32,12 @@ def __init__(
         self.supported_local_geometries = ["Miller", None]
 
         self.gk_file = gk_file
-        self.gk_type = gk_type
-        self.gk_code = self.gk_type
+        if gk_type is not None and gk_code is None:
+            raise DeprecationWarning(
+                "gk_type is no longer used, please use gk_code instead"
+            )
```
ZedThree: Should probably also keep `self.gk_code = gk_type` for now and remove in a future version?

ZedThree (commented):
Might be good to note the change in CHANGELOG.md?

pyrokinetics/examples/example.py
bpatel2107: It wasn't really being used before. A lot of the example needed updating

pyrokinetics/pyro.py
bpatel2107: How do I "raise" the warning without stopping the program. Can't seem to figure that out

ZedThree: The `warnings` module is what you want:

https://docs.python.org/3/library/warnings.html#warnings.warn

This users filter the warning out

bpatel2107:
I don't think we have a CHANGELOG.md. Can add it to the base directory

bpatel2107:
Any extra thoughts on the changelog @ZedThree ?

ZedThree:
LGTM :+1:

ZedThree (approved):

Issue #42, opened 2021-11-23
Plug-in architecture


ZedThree:
The various file readers and writers should use a plug-in architecture that can be easily extended to new codes/versions

bpatel2107:
Pretty sure this is done, closing now

Issue #32, opened 2021-10-20
Auto-detect file types


ZedThree:
We can probably do this to some extent already with the kinetic files by looking for certain metadata, e.g. `TRANSP_version`. For GK input files, we could either look for unique namelists/variables if they're present, or perhaps metadata in comments. For example, GS2 now creates `<filename>.used_inputs.in` files after finishing a simulation, which contains a comment like `! GS2 version: 8.0.6-1858-g4d2e5ecb`.

It might not always be possible, but it would be neat.

bpatel2107:
For the JETTO CDF this would also be possible as under the description is saved as a string `"JETTO profile data"`. 

For SCENE, the netCDF output doesn't have any metadata so we would need to add that into SCENE itself.

At the moment, when loading output data, you need to have the input file so we would need to have the autocheck on the input file which is possible but we'd need to be careful.

For CGYRO the output files are always written as `out.cgyro.*` so that is an easy check. It also write a `out.cgyro.version` file with the git has

GENE also writes out the git hash in the `parameter_000*` file

ZedThree:
We could also write metadata to the files we write, which would allow us to identify them later if necessary

bpatel2107:
Fixed in #58

PR #58, opened 2022-02-18
Kinetics plugin architecture

LiamPattinson:
I've refactored `Kinetics` to make use of a plugin architecture, and added automatic file type inference. This PR goes some way towards resolving Issues https://github.com/pyro-kinetics/pyrokinetics/issues/32 and https://github.com/pyro-kinetics/pyrokinetics/issues/42, though similar changes will need to be made to `GKCode` and `Equilibrium` (and more?) to fully resolve them. There are quite a few changes here, and I've tried to list them all below. Feel free to ask if you'd like anything explained in more detail, or if you want any further changes.

- The `Kinetics` class no longer deals with reading files; this responsibility has been handed over to `KineticsReader` classes (one for each supported file type, plus an abstract base class), and a dict-like factory object.
    - `KineticsReader` itself is an abstract base class which simply defines the expected function signatures of its subclasses.
    - Each `KineticsReader` subclass provides a read function and a verify function. The former creates a dict of Species, while the latter performs quick checks to ensure the provided file is valid.
    - `KineticsReaderFactory` behaves like a dict that generates new `KineticsReader` subclasses when given either a kinetics type ("SCENE", "JETTO", etc) or a filename. It contains the function `_infer_kinetics_type` that uses the `verify` function of each reader in turn until it finds a match. A global instance of this factory -- `kinetics_readers` -- is defined, and each reader is 'registered' with it in the `__init__.py` file within the new kinetics directory. 
- Attributes `species_names` and `nspecs` are now properties, and their values are generated lazily.
- The attribute `kinetics_type` is also a property, and may now only be set to a supported kinetics type
- The user-facing interface of `Kinetics` is mostly unchanged. The only change made is that it can longer be initialised without a filename. The only place I could see it being useful is in its `__deepcopy__` function, and when used outside of this function it had the potential to create an instance in a potentially-dangerous undefined state.
- The SCENE reader now uses xarray. It generally seems like a nicer package to work with than netCDF4
- The JETTO and TRANSP readers still use netCDF4, as both files contain variables that have the same names as one of their dimensions (which apparently isn't allowed). It may be possible to use xarray by dropping the offending variables.
- File readers now make use of 'with' statements to ensure everything is cleaned up properly.
- I've added unit tests for each of the new classes, and made some minor edits to `test_kinetics.py` to check if file inference is working.


pyrokinetics/kinetics/KineticsReaderSCENE.py
```diff
@@ -0,0 +1,107 @@
...
+            raise FileNotFoundError(
+                f"KineticsReaderSCENE could not find {filename}"
+            ) from e
+        except ValueError as e:
+            raise ValueError(
+                f"KineticsReaderSCENE must be provided a NetCDF, was given {filename}"
+            ) from e
+        # Given it is a netcdf, check it has the expected data_vars
+        try:
+            if not np.all(np.isin(["Psi", "TGLF_RMIN", "Te", "Ne"], data.data_vars)):
```
bpatel2107, line 104: @ZedThree It might be worth adding in a description to the SCENE netCDF that identifies the SCENE version in output. If that gets added in then we can have a more rigorous check here. Not a pyrokinetics issue though

bpatel2107:
This looks amazing! 

- To fully address #32 we would need to add in ones for `GKCode` and the `Equilibrium` objects. For #42 I think we potentially would need to do this for the  `LocalGeometry` class. At the moment there is only a `Miller` subclass but I need to add in another representation (Fourier). Each `GKCode` has its own bespoke reader/writer for that so that could be another addition.
- For using xarray for JETTO/TRANSP, I found that I would have to load in the data with netCDF first to delete the variable. Not sure if we save any time if we have to load the data normally anyways

When we create a pyro object, we're not currently implementing the auto-file checker. 

 https://github.com/pyro-kinetics/pyrokinetics/blob/86b6e210f468cbccc02165bbcc497d9aea6d1e22/pyrokinetics/pyro.py#L149-L164

We would need a `try` `except`  here if `kinetics_type is None`

pyrokinetics/kinetics/KineticsReaderSCENE.py
ZedThree: Already done! https://github.com/PlasmaFAIR/SCENE/blob/d78c3d6861c7eed7854590ba52c128f1d393cb96/src/outcdf.f90#L224-L227

There are now `software_name` and `software_version` file-level attributes that could be checked.

It would be good if we could also try to get similar metadata into TRANSP and JETTO, but that might be a little bit harder

ZedThree:
> For using xarray for JETTO/TRANSP, I found that I would have to load in the data with netCDF first to delete the variable. Not sure if we save any time if we have to load the data normally anyways

`xarray.open_dataset` has a `drop_variables` argument, but the issue is that `Equilibrium` does need some of those variables. So we could use it here for `Kinetics`, but not in general.

This is actually an open issue on xarray, and should hopefully be fixed in the near future, at which point we should be able to just switch over to xarray entirely.

pyrokinetics/kinetics/KineticsReaderSCENE.py
bpatel2107: TRANSP already has a version number which pyrokinetics now check if it exists, though it doesn't do anything with it.

JETTO has a build date but not any version info. I can probably add in a version output to the JETTO CDF at some point (sometimes it's good to fingers in a lot of pies!)

bpatel2107:
> > For using xarray for JETTO/TRANSP, I found that I would have to load in the data with netCDF first to delete the variable. Not sure if we save any time if we have to load the data normally anyways
> 
> `xarray.open_dataset` has a `drop_variables` argument, but the issue is that `Equilibrium` does need some of those variables. So we could use it here for `Kinetics`, but not in general.
> 
> This is actually an open issue on xarray, and should hopefully be fixed in the near future, at which point we should be able to just switch over to xarray entirely.

The issue is the netCDF has the same key for the variable and coordinate so xarray gets confused. Maybe `drop_variables`  could still work, so the data isn't lost, it's just stored as a coordinate rather than a variable.

ZedThree:
> The issue is the netCDF has the same key for the variable and coordinate so xarray gets confused.

Sort of -- having the same name for a variable and a dimension is fine, in fact, this is a netCDF convention for a coordinate array. The problem xarray has is when the variable has more than one dimension, for example in JETTO, `X` is 2D:  `X(TIME3, X)`. If it were just `X(X)`, then xarray wouldn't have a problem.

This is an open issue that they are actively working on though, but I'm not sure when it will be fixed.

pyrokinetics/kinetics/KineticsReaderSCENE.py
ZedThree: Ah, great! Maybe you could add the full set of metadata: https://github.com/PlasmaFAIR/neasy-f/blob/905656649f2139981e5b25582ad51db8e6785269/src/neasyf.f90#L180

- title
- software_name
- software_version
- netcdf_version
- date_created
- id

LiamPattinson: I don't have access to a SCENE file with that metadata. What would be the best way to access `software_name` from the xarray dataset? Would it be as simple as checking `"SCENE" in data.software_name` and catching an AttributeError?

ZedThree: Yes, I think that's right. I'll send you an example file

LiamPattinson:
Thanks for the feedback, I'll try to get a better check for SCENE files working, and will take a look at Pyro. I just made a further commit that generalises KineticsReaderFactory, as I was setting up a similar system for Equilibrium and found I was essentially copy-pasting and making slight edits to error messages and attribute names. There's now a generic `Reader` object and a `create_reader_factory` function which defines and instantiates a factory class on-the-fly. It works essentially the same, though I'd advise not dwelling too long on the fact that it's technically a factory-of-factories.

ZedThree (approved):
LGTM :+1:

@bpatel2107 Is there anything you think is missing? Liam's working on `Equilibrium`, that'll be in a new PR so this one doesn't get massive

bpatel2107:
Also happy, great work @LiamPattinson!

Issue #60, opened 2022-02-22
Equilibrium.get_flux_surface interferes with user's matplotlib code


LiamPattinson:
```python
from pyrokinetics.equilibrium import Equilibrium
import matplotlib.pyplot as plt
import numpy as np

# Do some plotting
x = np.linspace( 0, 2*np.pi, 1000)
y = np.sin(x)
plt.plot(x,y)

# get flux surfaces
eq = Equilibrium("pyrokinetics/templates/transp_eq.cdf", "TRANSP")
R_con, Z_con = eq.get_flux_surface(1.0)

plt.show() # nothing happens!
```
When using `get_flux_surface`, it deletes any work the user may have been doing with matplotlib. The cause lies with the calls to `mpl.use`:

https://github.com/pyro-kinetics/pyrokinetics/blob/766ec1afa1875a3f1923b6404180ddf8eddcea78/pyrokinetics/equilibrium.py#L383-L388
https://github.com/pyro-kinetics/pyrokinetics/blob/766ec1afa1875a3f1923b6404180ddf8eddcea78/pyrokinetics/equilibrium.py#L425

If you comment out the backend changing lines, and replace the contour line with this:

```python
fig, ax = plt.subplots()
con = ax.contour(self.R, self.Z, psin_2d, levels=[psi_n])
plt.close(fig)
```
everything runs just fine, but the tests now take 3 times as long to run. Ideally, there should be a way to retrieve the contour without matplotlib. Perhaps [skimage](https://scikit-image.org/docs/dev/auto_examples/edges/plot_contours.html) would help? It uses the same underlying algorithm, but I'd need to do some side-by-side testing to ensure it can deliver the same results.

bpatel2107:
I had a go with skimage but it often struggled to find the flux surfaces as one continuous contour. Rather it found multiple contours that look linked by eye but aren't in the code. I believe I used the following method

https://scikit-image.org/docs/dev/api/skimage.measure.html#skimage.measure.find_contours

Maybe there's a better way or I didn't do a good job the first time.

The reason I did the backend switch was for speed as you see, but clearing the figure like this is not great. The only other option I can think of is to save any existing figures before changing the backend though I don't know if that's fast/better

bpatel2107:
Fixed in #123

PR #61, opened 2022-02-22
Equilibrium plugin architecture

LiamPattinson:
Building on PR https://github.com/pyro-kinetics/pyrokinetics/pull/58, this updates Equilibrium to make use of a plugin architecture and automatically infer file types. It also includes a few general improvements to factories/typing/etc.

- I'm not sure what to include in the docstring for Equilibrium. Some attributes have fairly obvious meanings, but as somebody unfamiliar with this domain, some of them are quite mysterious. I'd be grateful if you could let me know the physical meaning of each one. 
- `get_flux_surface` has been moved to its own file. There are some issues with this function, as described in Issue https://github.com/pyro-kinetics/pyrokinetics/issues/60

Please let me know if you'd like to see any further changes.


pyrokinetics/tests/equilibrium/test_equilibrium.py
```diff
@@ -5,29 +5,28 @@
...
-    return Equilibrium(eq_type="GEQDSK", eq_file=template_dir / "test.geqdsk")
+    return Equilibrium(eq_type="GEQDSK", eq_file=template_dir.joinpath("test.geqdsk"))
 
 
 @pytest.fixture(scope="module")
 def transp_cdf_equilibrium():
-    template_dir = pathlib.Path(__file__).parent / "../templates"
     return Equilibrium(
-        eq_type="TRANSP", eq_file=template_dir / "transp_eq.cdf", time=0.2
+        eq_type="TRANSP", eq_file=template_dir.joinpath("transp_eq.cdf"), time=0.2
```
ZedThree, line 21: Just curious, is there a difference between `.joinpath` and `/`?

pyrokinetics/readers.py
```diff
@@ -102,13 +109,13 @@ def __setitem__(self, key: str, value: Type[BaseReader]):
...
-        def __len__(self):
-            return len(self.__dict)
+    # Rename class
+    if name is None:
+        name = f"{BaseReader.__name__}Factory"
+    ReaderFactory.__name__ = name
+    ReaderFactory.__qualname__ = name
 
-    return ReaderFactory()
+    return ReaderFactory, ReaderFactory()
```
ZedThree: Possibly I've missed something, but do we need the class itself?

pyrokinetics/tests/equilibrium/test_equilibrium.py
LiamPattinson: Not as far as I can tell. I used `joinpath` in the kinetics tests, so switched it over here for consistency. I personally find `joinpath` to be clearer, but I'm happy to revert the change if I'm in the minority.

pyrokinetics/readers.py
LiamPattinson: Perhaps not. I think I made this change because I wanted to ensure it was creating types with the correct names in my unit tests, but ended up not needing it. It makes an already confusing bit of code even harder to understand, so it's probably best to revert it.

pyrokinetics/tests/equilibrium/test_equilibrium.py
bpatel2107: Happy with `joinpath`

bpatel2107:
Descriptions for each variable in Equilibrium

Not sure how to correctly format the info

```python
   """
    Defines information about Tokamak plasma equilibrium.
    Partially adapted from equilibrium option developed by B. Dudson in FreeGS
    Args
    ----
    eq_file (str/Path): Filename of an equilibrium file to read from.
    eq_type (str): Name of the equilibrium input type, such as "TRANSP" or "GEQDSK".
        If left as None, this is inferred from the input file.
    **kwargs: Extra arguments to be passed to the reader function. Not used by
        all readers, so only include this if necessary.
    Attrs
    -----
    supported_equilibrium_types: A list of all supported equilibrium input types.
    eq_files: Stored reference of the last file read
    eq_type: Stored reference of the last equilibrium type. May be inferred.
    nr: No. of radial points in the equilibrium file
    nz: No. of vertical point in the equilibrium file
    psi_axis: Poloidal flux on axis (Wb/m2)
    psi_bdry: Poloidal flux at the plasma boundary (Wb/m2)
    R_axis: Major radius of magnetic axis (m)
    Z_axis: Vertical position of magnetic axis (m)
    lcfs_R: (array) Radial positions of last closed flux surface (m)
    lcfs_Z: (array) Vertical positions of last closed flus surface (m)
    a_minor: Minor radius (m)
    bcentr: Magnetic field at magnetic axis (T)
    current: Total plasma current (A)
    Functions
    --------
    get_b_radial: Returns radial B field for a given (R,Z)
    get_b_vertical: Returns vertical B field for a given (R,Z)
    get_b_toroidal: Returns toroidal B field for a given (R,Z)
    get_b_poloidal: Returns poloidal B field for a given (R,Z)
    psi_RZ: Returns poloidal flux for a given (R, Z)
    get_flux_surface: Returns a flux surface for a given (psi)
    f_psi: Returns f = R * Bphi (m*T) for given normalised (psiN)
    ff_prime: Returns ff' for given a normalised (psiN)
    q: Returns safety factor for given a normalised (psiN)
    pressure: Returns total pressure (Pa) for a given normalised (psiN)
    p_prime: Returns pressure gradient w.r.t psi (Pa /(Wb /m2)) for a given normalised (psiN)
    rho: Returns Normalised minor radius r/a for a given normalised (psiN)
    R_major: Returns f = R * Bphi (m*T)  for given normalised (psiN)
    """
```

A lot of the attributes like `pressure` are `InterpolatedUnivariateSplines` so they're more like functions in the equilibrium. Do you think it is worth relabeling them like `get_pressure` to make that clearer?

I think we can remove certain attributes like `R`, `Z`, `psi` as there are just gridpoints. Given that everything is an interpolated function we don't necessarily need them. We don't really need `nr` and `nz` either. It might be useful if we ever wanted to recreate the original eq file but I can't imagine why we would do that or even if we should support that. Thoughts @d7919 ?

`bcentr` and `current` can be calculated in both GEQDSK and TRANSP so we should make sure both are there in each case

LiamPattinson:
Thanks for this! I've updated the docstrings, and removed a few variables. I had to keep `R` and `Z`, as they're used by `get_flux_surface()`, and everything would break otherwise. I think it's fine to leave the names of function-like objects as they are, provided they're properly documented as functions.

I'm not sure how to calculate `bcentr` in TRANSP or `current` in GEQDSK, and I think that should be raised as a separate issue.

pyrokinetics/equilibrium/Equilibrium.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+    Functions
+    --------
+    get_b_radial: Returns radial B field for a given (R,Z)
+    get_b_vertical: Returns vertical B field for a given (R,Z)
+    get_b_toroidal: Returns toroidal B field for a given (R,Z)
+    get_b_poloidal: Returns poloidal B field for a given (R,Z)
+    psi_RZ: Returns poloidal flux for a given (R, Z)
+    get_flux_surface: Returns a flux surface for a given (psi)
```
bpatel2107: Sorry this should be `psiN`

```suggestion
    get_flux_surface: Returns a flux surface for a given (psiN)
```

bpatel2107 (approved):
Looks good overall! Small correction needed for the docstring

PR #62, opened 2022-02-22
Flake8 fixes from #58

bpatel2107:
Addresses a few `flake8` issues from #58 

This still currently fails a few `flake8` tests.

```
flake8 --exit-zero
./pyrokinetics/kinetics/__init__.py:2:1: F401 '.KineticsReader.KineticsReader' imported but unused
./pyrokinetics/kinetics/__init__.py:13:1: F401 '.Kinetics.Kinetics' imported but unused
./pyrokinetics/kinetics/__init__.py:13:1: E402 module level import not at top of file
```

 @LiamPattinson is there any reason I can't remove those imports/shuffle them around

### DIFF
--- a/pyrokinetics/kinetics/Kinetics.py
+++ b/pyrokinetics/kinetics/Kinetics.py
@@ -1,8 +1,7 @@
 from pathlib import Path
-from typing import Union, Dict, Optional
+from typing import Union, Optional
 from cleverdict import CleverDict
 
-from ..species import Species
 from .KineticsReader import kinetics_readers
 
 
@@ -63,13 +62,13 @@ class Kinetics:
 
     @property
     def kinetics_type(self):
-        return _kinetics_type
+        return self._kinetics_type
 
     @kinetics_type.setter
     def kinetics_type(self, value):
         if value not in self.supported_kinetics_types:
             raise ValueError(f"Kinetics type {value} is not currently supported.")
-        _kinetics_type = value
+        self._kinetics_type = value
 
     @property
     def nspec(self):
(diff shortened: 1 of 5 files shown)


ZedThree:
For the unused imports, do:

```python
__all__ = ["Kinetics"]
```

I suspect the `KineticsReader` import can be deleted.

The other lines can be rearranged.

bpatel2107:
Seems like I can't move the module level import to the beginning, otherwise it won't recognise the different readers and the tests will fail. This set up will always fail with flake8.

bpatel2107:
Added a line ignoring that specific PEP8 rule for that import. Unless there's a reason not to do that @ZedThree?

ZedThree (approved):

ZedThree:
"Beginning of the file" is a bit misleading, it really means before any of the executable statements. But the `#noqa` is harmless

pyrokinetics/kinetics/Kinetics.py
```diff
@@ -63,13 +62,13 @@ def __init__(
...
     def kinetics_type(self):
-        return _kinetics_type
+        return self._kinetics_type
 
     @kinetics_type.setter
     def kinetics_type(self, value):
         if value not in self.supported_kinetics_types:
             raise ValueError(f"Kinetics type {value} is not currently supported.")
-        _kinetics_type = value
+        self._kinetics_type = value
```
LiamPattinson, line 71: Apologies for not catching that bug! I'll include unit tests for it in the next pull request.

LiamPattinson:
Sorry for not contributing to this yesterday, I was taking some time to work on my thesis. In a previous project we added a rule to `setup.cfg` that flake8 should ignore 'unused import' errors in `__init__.py` files, as the purpose of these imports is to expose features of the library to the end user. Would it be a good idea to do the same here, or would it be better to add noqa to individual lines?

bpatel2107:
That would be better than adding these noqa lines each time we add an `__init__.py` file.

PR #63, opened 2022-02-23
Bugfix/miller float point error

bpatel2107:
Handle floating point error preventing `arcsin(>1.0)` when setting up `theta` grid in `miller.py`


ZedThree (approved):

ZedThree:
Might be good to get a test case for this

PR #65, opened 2022-02-28
GKCode plugin architecture

LiamPattinson:
Building on PR https://github.com/pyro-kinetics/pyrokinetics/pull/58 and https://github.com/pyro-kinetics/pyrokinetics/pull/61, this updates GKCode to make use of a plugin system. This has been achieved using existing tools that were previously applied to Kinetics and Equilibrium.

### DIFF
--- a/pyrokinetics/gk_code/GKCodeCGYRO.py
+++ b/pyrokinetics/gk_code/GKCodeCGYRO.py
@@ -13,7 +13,7 @@ from path import Path
 from cleverdict import CleverDict
 
 
-class CGYRO(GKCode):
+class GKCodeCGYRO(GKCode):
     """
     Basic CGYRO object inheriting method from GKCode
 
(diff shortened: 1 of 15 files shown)


LiamPattinson:
I'm confused as to why the commits from my last pull request are still included here as 'files changed'. I expected them to be removed from here if they were merged into 'unstable'. Am I missing something?

bpatel2107:
Maybe because you pushed before the merge and haven't committed since? Can you do a blank/empty commit, maybe that will fix it?

ZedThree:
That's a bit strange. Maybe merge in `unstable`?

LiamPattinson:
Ah, that seems to have done it. Cheers!

pyrokinetics/gk_code/GKCodeCGYRO.py
```diff
@@ -59,6 +58,13 @@ def read(self, pyro, data_file=None, template=False):
         if not hasattr(pyro, "numerics"):
             self.load_numerics(pyro, cgyro)
 
+    def verify(self, filename: PathLike):
+        """read cgyro file, check the dict returned holds the expected data"""
+        data = self.cgyro_parser(filename)
+        expected_keys = ["EQUILIBRIUM_MODEL", "N_FIELD", "DELTA_T", "N_SPECIES"]
```
bpatel2107, line 64: Normally all of these keys are in the input file but we may need to re-evaluate this in the future because you can technically run CGYRO with an empty file.

bpatel2107 (approved):

PR #66, opened 2022-03-08
LocalGeometry plugin architecture

LiamPattinson:
Similar to pull requests https://github.com/pyro-kinetics/pyrokinetics/pull/58, https://github.com/pyro-kinetics/pyrokinetics/pull/61, and https://github.com/pyro-kinetics/pyrokinetics/pull/65, this implements a plugin architecture for LocalGeometry classes, anticipating the addition of a Fourier type in some future commit. This completes issue https://github.com/pyro-kinetics/pyrokinetics/issues/42.

I generalised the ReaderFactory from previous commits to a generic Factory class, which ReaderFactory now subclasses. It behaves similarly, but now also works in cases where we don't need file type inference.

The Miller class has been renamed LocalGeometryMiller to make it clearer that it belongs to the 'LocalGeometry' family, but I'm happy to revert that change if it seems too verbose.

### DIFF
--- a/pyrokinetics/factory.py
+++ b/pyrokinetics/factory.py
@@ -1,0 +1,64 @@
+"""
+Defines generic Factory object.
+Factory objects used throughout this project may inherit this class and add additional
+features, such as filetype inference.
+"""
+
+from typing import Type
+from collections import UserDict
+
+
+class Factory(UserDict):
+    """
+    Given a key as a string, returns an object object derived from BaseClass, which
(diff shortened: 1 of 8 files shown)


bpatel2107 (approved):
Overall looks good! The name is getting a tad long but for now it seems okay. Most people won't be directly calling it anyways

bpatel2107:
The `black` and `flake8` workflows don't seem to run when you push. @ZedThree is that due him not having "access"

ZedThree:
Ah, yes, that's because the PR is from a fork

Issue #64, opened 2022-02-26
Fails to load linear data with more than one ky or kx value


d7919:
In the gk_code base object there's a default implementation for calculating the growth rate and frequency from the amplitude of the fields, which is used in the cgyro child if the fields are not Nan. This tries to drop-squeeze the ky and kx dimensions, which will error if there are more than one ky/kx value.

I think we should be able to calculate the growth rate and frequency even when there is more than one wavenumber present, although it may require some change to the code (e.g. looping/using broadcastable logic).

bpatel2107:
So this would be the case when we have a `N_TOROIDAL>1` and the nonlinear term turned off. I think it would be trivial to do this for different `ky` and `kx`. Essentially the datarray containing the eigenvalues would just have another co-ordinate in `ky` and `kx`

My only concern is avoid any confusing with ballooning coordinates as with `N_TOROIDAL=1` we have to convert the `kx` grid to a ballooning space grid. I don't think we would need to do this for multiple `ky` but want to confirm?

d7919:
For the time being I'm just forcing the code to drop back to the 'read from file' path. This seems to work until `BOX_SIZE > 1`. Once `BOX_SIZE>1` it fails as Pyro seems to want to read into something where one of the dimensions is `N_RADIAL`, but the data seems to contain `N_RADIAL/BOX_SIZE` instead.

PR #67, opened 2022-03-11
Pyro/GK refactor

LiamPattinson:
I've drafted some substantial changes to Pyrokinetics which aim to clarify the responsibilities of each class. Currently, `GKCode` objects are mostly stateless, and they serve to set up `Pyro` objects. This leads to confusing control flow, and requires each `GKCode` object to have a detailed knowledge of the inner workings of `Pyro`. It also presents difficulties for future updates, as any change to the workings of `GKCode`, `GKOutput` or `Pyro` must be reflected in each other.

So far I've only worked with GS2 classes, and I've developed a competing architecture instead of overwriting the current system. If these updates are accepted, I hope to replace the current `GKCode`, `GKOutput`, and `Pyro` objects (and would probably need to update `PyroScan` too).

- Each `GKCodeGS2` has been split into three classes: a `GKInputReaderGS2`, a `GKInputWriterGS2`, and a `GKOutputReaderGS2`. All have associated base classes and factories.
    - `GKInputReaderGS2` reads in a Fortran 90 namelist and saves it to its `self.data` attribute. It then provides the functions `get_local_geometry`, `get_local_species` and `get_numerics`.
    - `GKInputWriterGS2` provides a `write` function that takes a `LocalGeometry`, `LocalSpecies`, and `Numerics`. It uses a `GKInputReaderGS2` to read in a template file, and then edits the namelist before writing.
    - `GKOutputReaderGS2` reads in a GS2 netcdf and converts it to the Pyro standard. It currently has no unit tests, and I think there are some issues with it. I've only tried it with the output file generated by running GS2 on `defaults.in`, itself created by the script `gs2/bin/write_default_input_file`. I'm fairly certain that the default input file is broken, and it doesn't produce any useful info from GS2, so it would be helpful for the project to contain a collection of output files for testing.
- A new class `PyroAlt` has been created, which may replace `Pyro` in future. It doesn't share the interface of `Pyro`, so all examples, tests, and docs will need to be updated. It can read in `gk_input_file`s, `eq_file`s, and `kinetics_file`s, and if provided with a `psi_n` it automatically overwrites the `local_geometry` and `local_species` with versions derived from the global equilibrium and kinetics.

I've submitted this as a draft pull request so I can get some feedback before going any further. I've left in a lot of TODO and FIXME notes which will be removed before the full pull request. It would be very helpful if comments could be added to clarify some of these points.


pyrokinetics/gk_code/GKInputReaderGS2.py
```diff
@@ -0,0 +1,244 @@
...
+        return self.data
+
+    def verify(self, filename: PathLike):
+        """
+        Ensure this file is a valid gs2 input file"
+        """
+        data = f90nml.read(filename).todict()
+        expected_keys = ["kt_grids_knobs", "theta_grid_knobs", "theta_grid_eik_knobs"]
+        if not np.all(np.isin(expected_keys, list(data))):
+            raise ValueError(f"Expected GS2 file, received {filename}")
```
ZedThree: @d7919 will be able to weigh in more, but I think these namelists are all technically optional, so perhaps a more accurate error would be something like "unable to verify {filename} as GS2 file"?

pyrokinetics/gk_code/GKInputReaderGS2.py
```diff
@@ -0,0 +1,244 @@
...
+                    return self.get_local_geometry_miller()
+                else:
+                    # return self.get_fourier()
+                    raise NotImplementedError("GS2 Fourier options are not implemented")
+            else:
+                raise RuntimeError("GS2 is not using local equilibrium")
+        else:
+            raise NotImplementedError(
+                f"GS2 equilibrium option {gs2_eq} not implemented"
+            )
```
ZedThree: I would invert most of these to have the preconditions up front:

```suggestion
        if gs2_eq not in ["eik", "default"]:
            raise NotImplementedError(f"GS2 equilibrium option {gs2_eq} not implemented")

        local_eq = self.data["theta_grid_eik_knobs"].get("local_eq", True)
        if not local_eq:
            raise RuntimeError("GS2 is not using local equilibrium")

        iflux = self.data["theta_grid_eik_knobs"].get("iflux", 0)
        if iflux != 0:
            # return self.get_fourier()
            raise NotImplementedError("GS2 Fourier options are not implemented")

        return self.get_local_geometry_miller()
```

pyrokinetics/gk_code/GKInputReaderGS2.py
```diff
@@ -0,0 +1,244 @@
...
+    def get_local_geometry_miller(self) -> LocalGeometryMiller:
+        """
+        Load Miller object from GS2 file
+        """
+        # TODO can we get psi_n from this?
+
+        # Set some defaults here
+        # FIXME Why do we modify self as a side effect?
+        self.data["theta_grid_eik_knobs"]["bishop"] = 4
+        self.data["theta_grid_eik_knobs"]["irho"] = 2
```
ZedThree: One for @bpatel2107 ?

pyrokinetics/gk_code/GKInputWriterGS2.py
```diff
@@ -0,0 +1,212 @@
...
+        gs2_input = self.numerics_to_gs2_input(gs2_input, numerics, local_geometry)
+
+        # Create directories if they don't exist already
+        filename = Path(filename)
+        filename.parent.mkdir(parents=True, exist_ok=True)
+
+        # Create Fortran namelist and write
+        gs2_nml = f90nml.Namelist(gs2_input)
+        gs2_nml.float_format = float_format
+        gs2_nml.write(filename, force=True)
```
ZedThree: Just a note that it might be nice to add a header to input files, something like

```
! Input file generated by pyrokinetics on ${date}
```

Not necessarily in this PR!

pyrokinetics/gk_code/GKOutputReaderGS2.py
```diff
@@ -0,0 +1,203 @@
...
+from ..typing import PathLike
+
+
+class GKOutputReaderGS2(GKOutputReader):
+    @staticmethod
+    def get_input_nml_from_output_dataset(dataset: xr.Dataset):
+        # Get input file as string
+        input_str = str(dataset["input_file"].data)
+        # Remove beginning 'b and ending '
+        input_str = input_str[2:-1]
```
ZedThree: This is a little suspicious! I'm guessing this is because `dataset["input_file"].data` is `bytes` -- if so, instead do:

```
input_str = dataset["input_file"].data[()].decode("utf-8")
```

`data[()]` is the annoying way you have to index a scalar numpy array

Also, unfortunately, GS2 currently has two subtly different output file formats. In the "other" format, `input_file` is stored as an `(N, 200)` array, so you will need to handle that something like:

```python
input_file = dataset["input_file"]
if input_file.shape == ():
    # New diagnostics
    input_str = dataset["input_file"].data[()].decode("utf-8").split("\\n")
    input_str = "\n".join((line.rstrip() for line in input_str))
else:
    # Old diagnostics (and eventually the single merged diagnostics)
    input_str = "\n".join((line.decode("utf-8").rstrip() for line in dataset["input_file"].data))

return f90nml.reads(input_str)
```

pyrokinetics/gk_code/GKOutputReaderGS2.py
```diff
@@ -0,0 +1,203 @@
...
+        # Split into lines by stringified newlines
+        lines = input_str.split("\\\\n")
```
ZedThree: If you decode as above, then I think you'll only need `\\n` here

pyrokinetics/local_geometry/LocalGeometryMiller.py
```diff
@@ -1,10 +1,30 @@
...
+from copy import copy
 from scipy.optimize import least_squares  # type: ignore
 from ..constants import pi
-import numpy as np
 from .LocalGeometry import LocalGeometry
 from ..equilibrium import Equilibrium
 from ..typing import Scalar, ArrayLike
-from typing import Tuple
+
+default_miller_inputs = {
```
ZedThree: This is perhaps more of a style thing, but I think it's good to keep constants in `ALL_CAPS`

pyrokinetics/pyro_refactor.py
```diff
@@ -0,0 +1,318 @@
...
+            code, such as SCENE, JETTO, or TRANSP. When passed, this will set the
+            'kinetics' attribute. If the parameter 'psi_n' is set, or psi_n can be
+            inferred from the gk_input_file, this will be used to create/overwrite
+            local_species.
+        kinetics_type (str, optional): Type of kinetics file. When set, this will skip
+            file type inference. Possible values are SCENE, JETTO, or TRANSP. If set to
+            None, the file type is inferred automatically.
+        psi_n (float, optional): The normalised local flux surface, with 0.0 being the
+            central toroidal field line and 1.0 being the last closed flux surface. If
+            gk_input_file is set, psi_n is inferred, and this parameter is ignored.
```
ZedThree: There's an argument to be made perhaps that the passed `psi_n` should override the value from the GK input file?
@bpatel2107 @d7919 thoughts?

pyrokinetics/tests/test_pyro.py
```diff
@@ -0,0 +1,97 @@
...
+def test_init_with_kinetics_file(kinetics_file, kinetics_type):
+    pyro = Pyro(kinetics_file=template_dir / kinetics_file, kinetics_type=kinetics_type)
+    assert str(pyro.kinetics_file) == str(template_dir / kinetics_file)
+    if "scene" in kinetics_file:
+        assert pyro.kinetics_type == "SCENE"
+    if "jetto" in kinetics_file:
+        assert pyro.kinetics_type == "JETTO"
+    if "transp" in kinetics_file:
+        assert pyro.kinetics_type == "TRANSP"
+    assert isinstance(pyro.kinetics, Kinetics)
```
ZedThree: Could add `expected_type` to the parameters? `("scene.cdf", "SCENE", "SCENE"), ("scene.cdf", None, "SCENE")`, then this would be `assert pyro.kinetics_type == expected_type`

ZedThree (commented):
I'm not 100% convinced about the split between `GKInputReader`/`GKInputWriter`/`GKOutputReader`, although it is true that there's very little, if any, overlap between them.
Even so, I would be in favour of keeping them all in the same file, as they do logically belong together. That would reduce any potential explosion in the number of files.

Could you add one of the examples converted to use `PyroAlt`? It's currently a bit tricky to see what effect the change in API would have on the end user

pyrokinetics/gk_code/GKInputReaderGS2.py
d7919: Yes, technically an empty file is a valid GS2 input. We do expect the file name to end in `.in` (or possibly `.list` but I don't think that's supported here), although I'm not sure if we actually enforce this.

pyrokinetics/gk_code/GKInputReaderGS2.py
d7919: Using `local_eq = T` and `iflux /= 0` is an error in GS2 (or at least it's not correct). The Fourier representation is actually activated by `local_eq=T`. If you want to halt if we're using a non-Miller equilibrium then you would need to test `geoType` (https://gyrokinetics.gitlab.io/gs2/page/namelists/index.html#theta_grid_parameters-geotype) as well as `local_eq` I think.

pyrokinetics/local_geometry/LocalGeometryMiller.py
LiamPattinson: I agree, as it makes it clearer that the user shouldn't modify the dict, but it's still easy to break things by accident:

```python
my_data = DEFAULT_MILLER_INPUTS # looks okay, but user should have copied
my_data[key] = value # modifies the original!
```

Originally, the code used a function call to get inputs, creating a new dict every time. I think it might be a good idea to go back to a `get_default_miller_inputs` function just for safety.

pyrokinetics/gk_code/GKInputReaderGS2.py
LiamPattinson: An issue here is that `verify` must be able to uniquely identify a file type in some way, so testing for a `.in` extension could return a false positive if any other code uses the same convention. This isn't too serious, as the user can override file type inference by passing `gk_input_type="GS2"`, but it could lead to some confusion if a user tries inputting a GENE file and Pyrokinetics thinks it's looking at a GS2 file. Are we aware of any other codes that use `.in`?

d7919: I don't know of any codes which use that, but wouldn't like to say that there aren't any! Could it work by eliminating possibilities, rather than trying to guess the right option? Each code type checks essential conditions of the input file and then if there's more than one that doesn't complain we raise an error saying it is ambiguous so the user has to specify?

~~~~
possible_codes = [
       x for x in GKInputReaders if x.compatible_input(input)
]

if len(possible_codes) > 1:
   raise ....
elif len(possible_codes) < 1:
   raise .... 
else:
   reader = possible_codes[0]
~~~~

d7919: In the first branch we could even go  over the possible codes and ask each to look for optional things which may give a hint. For example GS2 could check if any of its namelists are found, if so it can be quite confident that it is probably a GS2 input.

bpatel2107: Even if nothing currently has a ```.in``` file (which `STELLA` might) I wouldn't want to make sure we can integrate a new code that could have this suffix. This is a tricky problem overall. Bar running the codes themselves it will be difficult to 100% guarantee the correct code is being read. This may just be a few edge cases though

pyrokinetics/gk_code/GKInputReaderGS2.py
bpatel2107: `bishop` sets the magnetic shear and equilibrium pressure gradient are defined. With the current method of writing files this  should always be fine.

`irho` sets which radial co-ordinate we use. Setting it to use forces the minor radius to be used which is the only supported option at the moment. You could imagine extending this in the future to other coordinates but I don't see that happening soon https://gyrokinetics.gitlab.io/gs2/page/namelists/index.html#theta_grid_eik_knobs-irho

pyrokinetics/gk_code/GKInputWriterGS2.py
bpatel2107: Maybe even include the pyrokinetics commit

pyrokinetics/gk_code/GKInputReaderGS2.py
ZedThree: Yes, I think this should be a "best effort guess", and the user can always override that if it's wrong. In the factory maybe we want some output like "gk_code not specified, guessing filetype as GS2; set `gk_code` to override"

I like David's suggestion of turning this around so that the function says "this is consistent with a GS2 input file"

pyrokinetics/pyro_refactor.py
bpatel2107: `psi_n` isn't usually stored in the GK input file but maybe we can write it in the header of the file?

pyrokinetics/gk_code/GKInputReaderGS2.py
bpatel2107: Yes that check should be with `geoType`, looks like a bug! We do need `iflux = 0` as well but that's separate to checking if this is Miller or Fourier

LiamPattinson: Thanks for this. Just to check I'm understanding correctly, would the following work without touching `iflux`?
```python
if gs2_eq not in ["eik", "default"]:
    raise NotImplementedError(f"GS2 equilibrium option {gs2_eq} not implemented")

local_eq = self.data["theta_grid_eik_knobs"].get("local_eq", True)
if not local_eq:
    raise RuntimeError("GS2 is not using local equilibrium")

geotype = self.data["theta_grid_parameters"].get("geotype", 0)
if geotype != 0:
    raise NotImplementedError("GS2 Fourier options are not implemented")
```

bpatel2107: The key is `geoType`. Not sure if this is case sensitive, but to be safe...

Otherwise yes, though I would also force `iflux=0` after checking if it is a local equilibrium

pyrokinetics/pyro_refactor.py
LiamPattinson: I think I may have to change that comment, as `psi_n` actually _isn't_ inferred from the GS2 input file, and it isn't set in the `LocalGeometryMiller` object unless the user creates it from an `Equilibrium` object (I think that's true of both `Pyro` and `PyroAlt`). Is there some way to infer `psi_n` from a GS2 input file?

bpatel2107: Not the way we run it. Might be best to set it to None and then change it if it gets set by the user

pyrokinetics/gk_code/GKInputReaderGS2.py
ZedThree: Ah, but this when reading the file. Should we instead error if these options are not what we expect/need? If the input file has these set to non-default values, are we changing the meaning of some of the other options?

I guess these should really be set in the writer?

pyrokinetics/gk_code/GKInputReaderGS2.py
LiamPattinson: Going forward, I think you're right that a 'best effort guess' would be a better system. However, if you try to use a blank input file with Pyrokinetics and specify `gk_input_type="GS2"`, it currently doesn't know what to do with it, as it requires an input file to include, at a minimum: `kt_grids_knobs`, `theta_grid_knobs`, `theta_grid_eik_knobs` (plus some others I should add to the `expected_keys` list).

For now, I think it would be better for `verify` to test "can Pyrokinetics parse this as a GS2 file", rather than "is this a valid GS2 input". In future, I think it would be better if Pyrokinetics could fill in GS2 defaults automatically so that a blank input file would work as expected, but that's a separate PR.

pyrokinetics/gk_code/GKInputReaderGS2.py
bpatel2107: It shouldn't really matter what they are in the template. As long as they are set correctly when we write the file then we should be fine. But agreed that this could be moved to the writer.

Only caveat I guess is when reading in existing data that doesn't have this set. Then pyrokinetics can't deal with that output correctly

LiamPattinson:
Thanks for all the feedback. I've implemented most of your recommendations in my latest commit. Please let me know if you spot anything else that should be updated, or if you have any objections to the latest changes. In the meantime, I'm going to work on similar updates for CGYRO.

LiamPattinson:
I've pushed some further updates for CGYRO bits and refactored such that GKInputReader and GKInputWriter are now a single class: GKInput. The purpose of a GKInput is to store the contents of a gyrokinetics input file in a dict-like form, and to facilitate its translation either to/from disk, or to/from the Pyrokinetics objects LocalGeometry, LocalSpecies, and Numerics. It's similar enough to GKCode classes that it might be a good idea to simply rename GKInput to GKCode once refactoring is complete. I haven't yet written GKOutputReaderCGYRO, as I'd like to work on GENE inputs first. As before, let me know if you think something could be done better or if you spot any bugs/errors. I've marked some lines with #FIXME to denote places where I think there may be a bug or something could be done a better way.

LiamPattinson:
I've added refactored GENE bits, and one unit test is currently failing. It isn't possible to read a GENE input file and write it back out, as `Rgeo` isn't set in the LocalGeometry. This problem also applies outside of this refactor, so I'll raise it in a separate issue.

bpatel2107:
#75 should address this issue

LiamPattinson:
Thanks @bpatel2107, everything seems to be working now. I'll be moving on to GKOutput-related tasks tomorrow.

bpatel2107:
@LiamPattinson Has there been any progress with the GKOutput for CGYRO and GENE?

LiamPattinson:
@bpatel2107 Sorry for the lack of updates, I was away on leave for a couple of weeks. I've made some updates to `GKOutputReaderGS2`, which now simply returns an xarray dataset, as I couldn't think of anything a `GKOutput` class could offer that isn't already a part of xarray. Since it'd considerably bloat the repo if I added one GS2 test file for each possible path through the reader class, I opted to 'monkey patch' it in testing (the function that processes raw data from disk is replaced on-the-fly with a function that returns ideal GS2-like info). I may be overdoing the parametrisation, as it currently generates about 150 unit tests, many of which will actually be redundant. There are other ways to parametrise a pytest fixture, so I'll play around with those to see if I can find a better way of doing things.

I haven't gotten on to CGYRO and GENE outputs yet, and a significant roadblock there is that I don't currently have access to the codes or any of their output files. I should be able to make some progress by adapting what's already available in GKCode classes, but I won't be able to test very rigorously.

There are a couple issues I wanted to flag up before wrapping up GS2 outputs:
- the functions `load_eigenvalues/eigenfunctions` in `GKCode` call `squeeze` over `kx` and `ky`, which drops those dimensions. It fails if either of these dimensions have a length greater than 1. Is this the desired outcome? They're only called for linear runs, but it's possible to have a linear run with more than one `ky`/`kx`.
- Previously, the number of fields is set depending on whether `fphi`, `fapar`, and `fbpar` are greater than zero. In my version, I'm assuming all three are present, so `nfields=3`, but I'm setting the corresponding slice of `data_vars['fields']` to zeros if `phi_t`, `apar_t` or `bpar_t` aren't present in the output file. Is this okay, or should I go back and make `nfields` variable again?

Let me know if there are any other changes you'd like me to make.

bpatel2107:
I have attached some CGYRO output files. 
[CGYRO_linear.zip](https://github.com/pyro-kinetics/pyrokinetics/files/8632055/CGYRO_linear.zip)

To gain access to the source code you need to follow the instructions below 
https://gafusion.github.io/doc/download.html

For gene I will email the files and the source code can be obtained here http://genecode.org/

There are cases where multiple `kx` and `ky` are simulated linearly (see #64) so removing the squeeze is wise. We may need to split up scenarios where we have multiple `kx` and `ky` with the nonlinear term turned off and on.

I think its best to not have the apar and bpar if they're not in the simulation. There may be some analysis done by comparing different fields that could give misleading answers if they are set to 0. Removing them entirely seems the safest.

LiamPattinson:
Thanks @bpatel2107, I'll try to make some progress early next week. I'll start on reverting the field changes to the GS2 reader, and I'll update eigenvalues/functions to work with multiple `kx`/`ky`. For now, I'll continue skipping eigenvalues/functions for nonlinear runs. I also received your email with GENE data, so cheers for that. I'll make a start on CGYRO/GENE once everything seems to be working for GS2.

LiamPattinson:
Apologies for the delay since my last update. I've updated the GS2 output reader to only include fields when `["fphi","fapar","fbpar"] > 0`, and to include eigenvalues/eigenfunctions for all `kx`/`ky`. The tests appear to be failing because f90nml v1.4 doesn't agree with how we handle species in GENE input files. I'll raise this in a separate issue/PR.

bpatel2107:
Is there any update on this?

LiamPattinson:
I took some time off around the jubilee bank holidays, but I've made some slow progress. I had to rethink how CGYRO outputs would work, as we expect a collection of files rather than a single file, and it took me a while to settle on a design. I'm hoping to have a fix for f90nml 1.4 pushed today, and I should have a draft of my CGYRO output reader and some basic tests ready later. As a lot of CGYRO outputs are uncommented lists of numbers which have different meanings depending on the type of run (linear/nonlinear? ntheta_plot==ntheta_grid?), it might be challenging to design decent tests, and we'll have to carefully construct the mock outputs.

LiamPattinson:
I've now added `GKOutputReaderGENE`, and confirmed by hand that it gives the same results as `GKCodeGENE` with the output files I was given (also confirmed the same for CGYRO, though eigenvalues were about 0.01% different when `ntheta_grid==ntheta_plot` -- maybe just floating point errors?). As the GENE files come to about 75MB, I don't think they should be in the git repo, and we'll need smaller ones.

I have an idea for testing these classes, but it'll require the example output files in PR https://github.com/pyro-kinetics/pyrokinetics/pull/84. We could pickle the xarray Datasets currently produced by `GKCode`, save those in the testing folder, and use them for 'golden answer' tests with `GKOutputReader`. I'll wait until PR https://github.com/pyro-kinetics/pyrokinetics/pull/84 is done before writing those tests, as the output files I currently have access to are a bit too large, and will instead get on with replacing `GKCode` with `GKInput` and `GKOutputReader` to complete the refactor.

LiamPattinson:
Well, that's fun: a `pip` error for Python 3.7.
Otherwise, I'm finally done with this refactor (barring any feedback or further requested changes). Sorry for taking so long with it; I underestimated how long a refactor would realistically take, especially while balancing it with other projects. I tried to keep the behaviour of `Pyro` as similar as possible to the previous behaviour, but I've made some changes, which are shown in the `examples` folder:

- Setting `gk_code` no longer converts the underlying `GKCode`/`GKInput`. You can do this explicitly by calling the function `convert_gk_code`.
- The constructor no longer takes some arguments. `local_geometry` is only set after creating one via `GKInput`/`Equilibrium`. I also wasn't able to determine what `linear` and `local` were doing, but can add them back in if needed.

I wasn't able to fully update and test `PyroScan`, so that will require some further work. The new `Pyro` works similarly enough that it's mostly compatible, but it definitely won't work for `gk_outputs` anymore. I also haven't yet included tests for how `Pyro` handles `gk_output`. 
As always, let me know if you want any further changes. I know there are several PRs waiting on this, so I'll try my best to get to any feedback ASAP, but I'm on leave tomorrow and this Friday afternoon.

LiamPattinson:
I've now merged in PR https://github.com/pyro-kinetics/pyrokinetics/pull/84. Some issues:
- You now must explicitly provide gk template files. Previously, you could provide no gk_file, set `pyro.gk_code = "GS2"`, then write. I may need to rethink how things work to restore that functionality. (EDIT: It looks like I changed the behaviour when passing in `template_file` to `write_gk_file`. It'll take me a little while to work out how that interacts with any previously read `gk_file` and with `local_geometry`/`local_species` obtained from other sources.)
- `example_JETTO_1dscan_complicated` is now broken, as `Pyro` no longer contains the attribute `gs2_input`. I'm not really sure how to fix this, as I haven't made much progress in figuring out how PyroScan interacts with the rest of the project.

bpatel2107:
It looks like if no template file is specified then pyrokinetics loads in the templates which is expected behaviour

https://github.com/PlasmaFAIR/pyrokinetics/blob/e5038d1157ec064b65d4bb48251ce2062dbd99dc/pyrokinetics/gk_code/GKInputGS2.py#L306-L325

Here is check if `self.data` exists, which only gets loaded from a template file. If it is not there then it looks at the template file. If not specified then it uses the default. So I think not setting a template should be fine? 


In this new format `gs2_input` is essentially `pyro.gk_input.data` I think so a switch there should fix this. Either way that scan is broken as `beta` would get overwritten in the current way it is set up. This particular scan  will be partially addressed in #85

bpatel2107:
In the `set` method of a `GKInput` we have the following lines

https://github.com/PlasmaFAIR/pyrokinetics/blob/e5038d1157ec064b65d4bb48251ce2062dbd99dc/pyrokinetics/gk_code/GKInputGS2.py#L333-L352

This is only valid for Miller geometry. I think this needs to be pulled out into a `set_local_geometry_miller` method and then captured in a if statement. When we extend to Fourier geometry/S-alpha this will be different for each code.

bpatel2107:
If we change `gk_code` and the change back to the original code, will the input file still be stored?

I think we should try to ensure that the data is stored somewhere. It might be redundant but adding in a `gs2_input` and `cgyro_input` attribute for each code would solve this. When we switch codes we could check to see if this exists yet

LiamPattinson:
I've been working on some further updates to `Pyro`:

- Setting `Pyro.gk_code` with a string again switches the internal state to a different `GKCode`/`GKInput`
- Added explicit rules that specify the scenarios in which each attribute (local_geometry, gk_file, run_directory, numerics, etc) may or may not be overwritten
- If the user switches to a new `gk_code`, makes edits, and then switches back, the original data is preserved. 

It achieves this by keeping full records for all gyrokinetics codes, using `gk_code` as a context switch, and using properties to redirect attributes such as `.local_geometry` and `.gk_file` to the appropriate record depending on the current context. Currently having some trouble getting it working with `PyroScan`, which I'll hopefully have resolved tomorrow morning. I likely won't have time to implement unit tests for all of the changes I've made, so that may need to wait until Monday.

LiamPattinson:
`Pyro` looks a lot different now, but it should behave very similarly to how it did before I started messing with things. I've reverted the examples to how they were before, and it's once again possible to set `pyro.gk_code = "GS2"` without having first given it a `gk_file`. However, I've formalised what should happen under those circumstances, and tried to minimise the circumstances in which the user can find themselves working with uninitialised data. If the user switches to a new `gk_code`, makes some changes, and then switches back, the original `gk_file`, `gk_input`, `local_geometry`, `local_species`, and `numerics` will be unchanged (unless they call `convert_gk_code(original_gk_code)`, which will use data associated with the current `gk_code` to overwrite the chosen destination). The usual attributes are all properties now, so when you access `pyro.local_geometry`, internally `Pyro` will redirect to the version of `local_geometry` corresponding to the current `gk_code`.


> In this new format `gs2_input` is essentially `pyro.gk_input.data` I think so a switch there should fix this. Either way that scan is broken as `beta` would get overwritten in the current way it is set up. This particular scan will be partially addressed in #85

I've added `gs2_input`, `cgyro_input` and `gene_input` as properties that link to the underlying `gk_input.data`. However, it feels potentially dangerous to allow edits to the data directly, as that could bring it out of sync with the associated `local_geometry`, `local_species` and `numerics`. I've been trying to think of ways to update `add_flags` to allow the user to change `gk_input.data`, `local_geometry`, `local_species`, and `numerics` in a self-consistent manner, and in such a way that they won't be overwritten at the write stage, but so far I haven't thought of anything.

> In the `set` method of a `GKInput` we have the following lines
> 
> https://github.com/PlasmaFAIR/pyrokinetics/blob/e5038d1157ec064b65d4bb48251ce2062dbd99dc/pyrokinetics/gk_code/GKInputGS2.py#L333-L352
> 
> This is only valid for Miller geometry. I think this needs to be pulled out into a `set_local_geometry_miller` method and then captured in a if statement. When we extend to Fourier geometry/S-alpha this will be different for each code.

Would it be alright if we paused this until the PR that introduces the new geometries? We'll likely have to make substantial changes throughout the code, and it'll take quite a lot of effort to pull out everything Miller related. We can pull those lines into a function quite easily, but we'd still be locked into Miller geometry in other parts of the function, and it'll add a lot of complexity to make the whole function geometry agnostic.

I'll try to add some further unit tests on Monday, though I could do that in a separate PR if you need this merged ASAP. As always, let me know if you'd like any changes.

bpatel2107:
Overall this looks good.

As for the changing `gs2_input`/`cgyro_input` being dangerous. I agree that these could become out of sync if the user is not careful. Whenever we write files, we need to load the numerics, geometry and local_species after we set the `code_input` This was the input file always uses the the pyro object value. 

Any user should change a parameter in the correct way (i.e. via `pyro.local_geometry.kappa` as opposed to `pyro.cgyro_input['KAPPA']`) so we shouldn't allow for scenarios where they change a parameter in a non-standard way. Being out of sync will then just have no impact. The `add_flags` method was designed to access parameters not directly captured in Pyro somehow. If you want (in a future PR) we can add a check to see if the parameter exists in pyro and raise an error if they try and change a parameter in a non-standard way.


Happy to delay the geometry changes until we have other options in.

pyrokinetics/gk_code/GKOutputReaderCGYRO.py
```diff
@@ -0,0 +1,406 @@
...
+        }
+        # Read in files
+        raw_data = {}
+        for key, cgyro_file in expected_files.items():
+            if not cgyro_file.path.exists():
+                if not cgyro_file.required:
+                    continue
+                raise RuntimeError(
+                    f"GKOutputReaderCGYRO: The file {cgyro_file.path.name} is needed"
+                )
```
ZedThree: This is perhaps a bit clearer:

```suggestion
                if cgyro_file.required:
                    raise RuntimeError(
                        f"GKOutputReaderCGYRO: The file {cgyro_file.path.name} is needed"
                    )
```

pyrokinetics/gk_code/GKOutputReaderCGYRO.py
```diff
@@ -0,0 +1,406 @@
...
+            kx = [0.0]
+            nkx = 1
+        else:
+            # Output data actually given on theta_plot grid
+            ntheta = ntheta_plot
+            theta = [0.0] if ntheta == 1 else theta_grid[:: ntheta_grid // ntheta]
+            # TODO Is this correct? for nkx = 5, this linspace call gives
+            #      [-2., -1.25, -0.5, 0.25, 1.]. Was it intended  to give
+            #      [-2, -1, 0, 1, 2] instead?
+            kx = 2 * pi * np.linspace(-int(nkx / 2), int(nkx / 2) - 1, nkx) / length_x
```
ZedThree: Does this need fixing? @bpatel2107

pyrokinetics/local_geometry/LocalGeometryMiller.py
```diff
@@ -1,10 +1,33 @@
...
 from scipy.optimize import least_squares  # type: ignore
 from ..constants import pi
-import numpy as np
 from .LocalGeometry import LocalGeometry
 from ..equilibrium import Equilibrium
 from ..typing import Scalar, ArrayLike
-from typing import Tuple
+
+
+def get_default_miller_inputs():
```
ZedThree: Just a little bike-shedding:

```suggestion
def default_miller_inputs():
```

although even better would be to have these as default argument to `LocalGeometryMiller.__init__` -- that's for another PR though

pyrokinetics/local_species.py
```diff
@@ -196,6 +201,16 @@ def update_pressure(self):
...
 
+    def normalise(self):
+        # Normalise to pyrokinetics normalisations and calculate total pressure gradient
+        te = self["electron"].temp
+        ne = self["electron"].dens
+        for name in self.names:
+            species_data = self[name]
+
+            species_data.temp = species_data.temp / te
+            species_data.dens = species_data.dens / ne
```
ZedThree, line 212: I was worried this might not be idempotent, that is if you apply if twice the values shouldn't change, but it looks like it is, because the electron temperature/density will be 1 after the first application. Is there a test for this already? It might depend if `te/ne` get copied on L206/7 or just referenced.

I guess #85 might handle this differently too.

pyrokinetics/pyro.py
```diff
@@ -1,339 +1,1423 @@
...
+        ``None``
+
+        Raises
+        ------
+        RuntimeError
+            If no path is provided, and no ``gk_file`` exists.
+        Exception
+            Various errors may occur while processing a gyrokinetics output.
         """
+        if path is None:
```
ZedThree, line 870: Is it possible to move all this logic into the respective `GKInput` readers? This current method is not very extensible. They already have some related handling. Do they need a (static) method to return the appropriate `gk_output_file`?

pyrokinetics/pyro.py
```diff
@@ -1,339 +1,1423 @@
...
-            setattr(new_pyro, key, deepcopy(value))
+            setattr(new_pyro, key, copy.deepcopy(value))
 
         return new_pyro
+
+    # Add properties that allow direct access to GKInput.data
+    # TODO This feels dangerous... could use a refactor
+    # TODO Not sure how to automate generation of these when new gk_codes are added
+    @property
+    def gs2_input(self) -> Union[f90nml.Namelist, None]:
```
ZedThree, line 1367: I think this is probably fine as a stop-gap for now, but we do really need a better overall design, I feel.

This might mean, for instance, that the code-specific portions live completely separately from the code-independent part, and the API looks more like, for example:


```python
gs2 = GS2(input_file)  # Read in, no conversion
pyro = gs2.to_pyro()  # Normalise to canonical form
CGYRO.write(pyro, cgyro_template)  # Convert and write to CGYRO
GENE.write(pyro, gene_template, dict_of_extras)
```

Obviously this would be a significant change to user code, but would hopefully end up with a lot cleaner, clearer, and more extensible library

ZedThree (approved):
LGTM. I think the big changes now are really in how Pyrokinetics works at a higher level.

pyrokinetics/local_species.py
bpatel2107: I think they get copied here as `self["electron"].temp` is a float so the standard behavior would be to copy. After it has been applied `te/ne` will be one but the output will be the same.

pyrokinetics/gk_code/GKOutputReaderCGYRO.py
bpatel2107: We rarely use an odd number of `kx`, but yes good spot, it should be 


```suggestion
            # TODO Is this correct? for nkx = 5, this linspace call gives
            #      [-2., -1.25, -0.5, 0.25, 1.]. Was it intended  to give
            #      [-2, -1, 0, 1, 2] instead?
            kx = 2 * pi * np.linspace(-int(nkx / 2), int((nkx+1) / 2) - 1, nkx) / length_x
```

bpatel2107:
Overall happy with this too

bpatel2107 (approved):

pyrokinetics/pyro.py
LiamPattinson: I've moved the logic into the `GKOutputReaders`. The way in which `GKOutputReaders` are accessed from `Pyro` feels a bit clunky though, so it might be a good idea to return to this later.

pyrokinetics/pyro.py
LiamPattinson: Might it be worth raising this as an issue? I imagine it'll require another large refactor, but it should be easier to repurpose things now that the functionality of `GKCode` and `Pyro` are separated.

LiamPattinson:
Thanks for the comments, I've made a few small changes. There are only a few outstanding issues I was planning on doing as part of this work:

- Better unit tests for `GKOutputReaderCGYRO`, use new output examples
- Better unit tests for `GKOutputReaderGENE`, use new output examples
- Test new functionality in `Pyro`

I can have these done in the next few days, or you can merge now and I'll submit a new PR.

bpatel2107:
I can wait for the tests to be done

LiamPattinson:
I've added some 'golden answer' tests for the `GKOutputReader`s. I'm not able to validate that they're returning anything sensible from an analytical standpoint, but I was able to compare them against the results using `GKCode` methods. The golden answers are stored in netcdf4 files generated using `GKCode` methods and Xarray, which adds a few MB to the repo (As an aside, the output files generated by `GKOutputReader`s can't be written to netcdf4 directly, as netcdf4 doesn't support complex numbers. I had to split them into real and imaginary parts to write to disk, and recombine them for the tests. Should I raise an issue about this?). I've also added some tests for `Pyro`.

I'm happy to call this done, unless there are any last-minute changes you'd like to see.

bpatel2107:
Seems like most of the data is in the GENE golden answer (1MB) compared to CGYRO/GS2 (50kB). Not sure why this would be so much larger than the other. What outputs did you save? 

If you raise another issue (and assuming it easy to remove the larger cached files from a different merge) then happy to merge. 

Thanks for putting in all the hard work!

LiamPattinson:
I read in the contents of `templates/outputs/GENE_linear` using the `GKCode` method, and split it into real and imaginary parts so that I could write it as a netcdf4. It looks like the file `field_0001` is already about 700kb, mostly due to a lot a theta values, and that's carried across into the Xarray Dataset. The eigenfunctions take up the same amount of space:

```python
>>> import xarray as xr
>>> ds = xr.open_dataset("gene_linear_output_7d551eaa.netcdf4")
>>> ds["fields"].nbytes
709632
>>> ds["eigenfunctions"].nbytes
709632
```
The size is inflated a little beyond what's necessary since I added a `ReIm` coordinate to every data_var, regardless of whether they needed it, and this inserted a lot of unnecessary zeros for the non-complex data_vars. However, those data_vars are all negligible in size compared to the fields and eigenfunctions anyway, so the best way to reduce the size of this file would probably be to produce new reference data with fewer theta values.

bpatel2107:
Yeah for linear runs there is a bit of redundancy as the fields and eigenfunctions are essentially the same thing but slightly rearranged and shifted in the complex plane. 

@dake0795 has run GENE for the same case as GS2/CGYRO. I've emailed them to you

LiamPattinson:
@bpatel2107 @dake0795 Thanks for the updated GENE files, I've merged those in now. I've also received some CGYRO linear/nonlinear files from Dr Jeff Candy at General Atomics, so I'll look into getting those into the project soon, but I think that'd be best saved for the next PR.

I'd also prefer to try removing the large files in a new branch after I've had some time to read into things. I may have already complicated things quite a lot by replacing the GENE files with alternatives using the same names.

bpatel2107:
Okay sounds good, @LiamPattinson are you happy for me to merge now?

LiamPattinson:
@bpatel2107 Yup, I'm happy if you are.

Issue #72, opened 2022-03-25
Auto file detection not working


bpatel2107:
When I try to auto-detect a file it fails here

https://github.com/pyro-kinetics/pyrokinetics/blob/6c59544c36a8f8140cede0d3a024fedecb096a00/pyrokinetics/pyro.py#L177

Where exactly does the inference take place. We need to implement that before we get to the `read_gk_file` method. @LiamPattinson do we need to add a check if `gk_file is not None` and `gk_code is None` to try and infer the file type when a pyro object is initialised?

ZedThree:
Can you post the file that fails? Maybe replacing some of the values if they're sensitive

bpatel2107:
It's just this


```python
from pyrokinetics import Pyro

# Point to CGYRO input file
gk_file = 'cgyro_input.txt'

# Load in file
pyro = Pyro(gk_file=gk_file)
```

[cgyro_input.txt](https://github.com/pyro-kinetics/pyrokinetics/files/8350782/cgyro_input.txt)

LiamPattinson:
Apologies for that, it looks like I didn't push changes to the Pyro object. It certainly works with the refactored version I'm working on, but I'll put some effort into getting it working in a PR from the current unstable branch.

PR #73, opened 2022-03-25
Fixed file inference for GKCode

LiamPattinson:
Issue https://github.com/pyro-kinetics/pyrokinetics/issues/72 revealed that file type inference was not implemented for `Pyro` objects when reading `gk_files`. This pull request aims to correct that oversight.

### DIFF
--- a/pyrokinetics/pyro.py
+++ b/pyrokinetics/pyro.py
@@ -67,7 +67,7 @@ class Pyro:
 
         # Read gk_file if it exists (not necessarily load it)
         if self.gk_file is not None:
-            self.read_gk_file()
+            self.read_gk_file(self.gk_file, gk_code)
 
         self.base_directory = Path(__file__).dirname()
 
@@ -83,7 +83,7 @@ class Pyro:
         """
 
         if value is None:
-            self._gk_cde = GKCode()
+            self._gk_code = GKCode()
             return
 
         try:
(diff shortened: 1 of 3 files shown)


LiamPattinson:
Okay, so it looks like that broke _everything_. I can't understand how my changes could have broken this so thoroughly, and it runs without issue on my local machine. Any ideas?

d7919:
The errors all seem to be to do with not being able to find the example/template files. I can't see how your changes could have broken that so maybe it's worth trying to rerun the actions -- maybe it was some transient issue with the CI environment.

bpatel2107:
I tried re-running the jobs but it had the same error. No idea why it can't see the template files...

ZedThree:
GitHub Actions were unstable yesterday. Reran the tests today and they all passed.

@LiamPattinson Please could you add a test that fails without this change?

LiamPattinson:
@ZedThree Sorted that now, also added tests for equilibrium/kinetics file type inference via `Pyro`

pyrokinetics/pyro.py
```diff
@@ -83,7 +83,7 @@ def gk_code(self, value):
         """
 
         if value is None:
-            self._gk_cde = GKCode()
+            self._gk_code = GKCode()
```
ZedThree: Is it perhaps more useful/sensible to set this to `None`? Is the `GKCode` base class useful by itself? I'm just thinking that if we store `None` here, then the conditional around L192 could be:

```diff
-        # if self.gk_code is still None, we couldn't infer a file type
-        # Must also check against GKCode, and due to inheritance we need
-        # to check type directly rather than using isinstance
-        if self.gk_code is None or type(self.gk_code) == GKCode:
+        if self.gk_code is None: 
             raise ValueError("Could not determine gk_code from file type")
```

If it is indeed useful to have the base class here, then ignore this!

pyrokinetics/pyro.py
```diff
@@ -171,21 +171,35 @@ def read_gk_file(self, gk_file=None, gk_code=None):
...
+            raise ValueError("Please specify gk_file")
+
+        # if gk_code is not given, try inferring from possible GKCodes
+        if gk_code is None:
+            for key, gk in gk_codes.items():
+                try:
+                    gk.verify(self.gk_file)
+                    self.gk_code = key
+                    break
+                except Exception:
```
ZedThree, line 180: Not for this PR, and this is just a suggestion, but could we wrap any exception raised in `gk.verify` in our own type, e.g. `VerificationException`, then we can be a bit more explicit about what we're catching here?

pyrokinetics/pyro.py
```diff
@@ -171,21 +171,35 @@ def read_gk_file(self, gk_file=None, gk_code=None):
...
-            # If equilibrium already loaded then it won't load the input file
-            if hasattr(self, "eq"):
-                template = True
-            else:
-                template = False
+        # If equilibrium already loaded then it won't load the input file
+        if hasattr(self, "eq"):
+            template = True
+        else:
+            template = False
```
ZedThree: If we're touching this:

```suggestion
        # If equilibrium already loaded then it won't load the input file
        template = hasattr(self, "eq"):
```

ZedThree (approved):
LGTM, thanks @LiamPattinson!

A couple of comments to look at, that may or may not make sense

pyrokinetics/pyro.py
LiamPattinson: That's a good idea. It'll require a bunch of small changes throughout the code, but ultimately it'll easier to understand the control flow. It'll also give us the option of raising exceptions that shouldn't be caught during verification (although right now I can't think of an application for that).

pyrokinetics/pyro.py
LiamPattinson: I'd considered this myself, but assumed that the `GKCode` base class must be used somewhere. I've now set it to `None`, and all the tests are passing just fine, but since coverage is a bit patchy there's a good chance I'm missing some use case. @bpatel2107 please let me know if you'd like me to revert the change.

bpatel2107: THe `GKCode` class on it own doesn't do anything so setting this to `None` should be fine. The only issue is if someone tries to write a file without setting `GKCode` then it will probably fail without a useful warning. Though I think that is the case now so could be a separate issue. Fine with this for now.

PR #76, opened 2022-03-30
Fixed how rho is calculated from the JETTO kinetic data

dake0795:
Fixed how rho = r/a is calculated from netCDF profiles.


pyrokinetics/kinetics/KineticsReaderJETTO.py
```diff
@@ -4,9 +4,12 @@
 from ..species import Species
 from ..constants import electron_mass, hydrogen_mass, deuterium_mass
 
+import matplotlib.pyplot as plt
+
```
bpatel2107: Can you remove the unnecessary import?

pyrokinetics/kinetics/KineticsReaderJETTO.py
```diff
@@ -33,10 +36,13 @@ def read(self, filename: PathLike) -> Dict[str, Species]:
...
 
-            rho = kinetics_data["RMNMP"][-1, :].data
+            Rmax = kinetics_data["R"][-1, :].data
+            Rmin = kinetics_data["RI"][-1, :].data
+
+            r = (Rmax - Rmin) / 2
+            rho = r / r[-1]
             rho_func = InterpolatedUnivariateSpline(psi_n, rho)
 
-            # Electron data
```
bpatel2107, line 39: Any reason to remove the comment?

pyrokinetics/kinetics/KineticsReaderJETTO.py
```diff
@@ -33,10 +36,13 @@ def read(self, filename: PathLike) -> Dict[str, Species]:
             psi = psi - psi[0]
             psi_n = psi / psi[-1]
 
-            rho = kinetics_data["RMNMP"][-1, :].data
+            Rmax = kinetics_data["R"][-1, :].data
+            Rmin = kinetics_data["RI"][-1, :].data
+
+            r = (Rmax - Rmin) / 2
+            rho = r / r[-1]
```
bpatel2107, line 41: We need to check to see if the last closed flux surface is actually the last element. I think data stop just shy of it so we may need to slightly tweak it

ZedThree:
The failing tests are "golden answer" tests:

```python
assert np.isclose(species.get_norm_dens_gradient(0.5), midpoint_density_gradient)
```

where `midpoint_density_gradient` is a constant defined here: https://github.com/pyro-kinetics/pyrokinetics/blob/991d06a026880c55adbe4a1e158e60416dfc4d65/pyrokinetics/tests/kinetics/test_kinetics.py#L115

(and similarly below for the other species).

This is presumably expected because the definition of `rho` has changed, but we need to double-check that it's changed in an expected way. If so, the tests will need updating with the new number

PR #79, opened 2022-05-10
Enable recreating `PyroScan` instances

ZedThree:
- Save/restore `parameter_map` dict to/from JSON file
- Pull out a method for formatting the individual run directory names

Saving the `parameter_map` is essential in order to know what parameters the scan was over. Latter method enables iterating over the output files in a scan afterwards.


bpatel2107:
Looks good!

Issue #80, opened 2022-05-17
Unit tests fail GENE writing for f90nml v1.4


LiamPattinson:
`GKCodeGENE.write()` fails on the latest version of f90nml. It doesn't seem to like `iSp=0` (and may fail for `iSp>0` too)

```python
            if name == "electron":                                        
>               gene_input[species_key][iSp]["name"] = "electron"
E               KeyError: 0
                                                                                                                                                               
pyrokinetics/gk_code/GKCodeGENE.py:157: KeyError
```

bpatel2107:
Fixed in #82

PR #82, opened 2022-06-08
Fix for f90nml v1.4

LiamPattinson:
f90nml v1.4 changed how it would support repeated keys in Fortran namelists, which are used by GENE to represent multiple species. Rather than collecting them into a list by default, the user is required to explicitly declare them as 'cogroups'. When converting to and from dict, repeated namelist keys are listed as separate entries with keys '_grp_entry_0', '_grp_entry_1', etc. The preferred way to deal with cogroups is to build a `f90nml.Namelist` object first and then explicitly add cogroups, but as GKCode converts to dict after reading, we instead must prepend `'_grp_'` and append a number to the species key. 

v1.4 appears to have bugs, but 1.4.1 and 1.4.2 behave as expected. 1.4 has been excluded in setup.cfg.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,7 +28,7 @@ packages = find:
 include_package_data = True
 install_requires =
     numpy >= 1.20.3
-    f90nml >= 1.3, != 1.4
+    f90nml >= 1.3, != 1.4.0
     scipy >= 1.6.3
     netCDF4 >= 1.5.6
     path >= 15.1.2
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,7 +28,7 @@ packages = find:
 include_package_data = True
 install_requires =
     numpy >= 1.20.3
-    f90nml >= 1.3
+    f90nml >= 1.3, != 1.4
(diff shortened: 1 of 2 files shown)


LiamPattinson:
Fixes Issue https://github.com/pyro-kinetics/pyrokinetics/issues/80

setup.cfg
```diff
@@ -28,7 +28,7 @@ packages = find:
 include_package_data = True
 install_requires =
     numpy >= 1.20.3
-    f90nml >= 1.3
+    f90nml >= 1.3, != 1.4
```
bpatel2107: Does this need to be 1.4.0 or will 1.4.1 not be included here.

LiamPattinson: On my system it only excludes 1.4, and will install 1.4.2 by default. f90nml doesn't seem to use a patch number on new minor releases, so there is no v1.4.0. However, it looks like pip behaves the same either way, and 1.4.0 is more explicit, so I'm happy to change it.

bpatel2107 (approved):

PR #88, opened 2022-06-21
Enhancements/tglf gk code

bpatel2107:
Added in TGLF as an option for a `GKCode`

- [x] Reads in input files
- [x] Reads in output data
- [x] Added examples
- [x] Added test

PR #91, opened 2022-07-11
Add TGLF into pyrokinetics

bpatel2107:
Supercedes #88 

Added in TGLF as an option for a `GKCode`

- [x] Set up GKInputTGLF
- [x] Set up GKOutputReaderTGLF
- [x] Added examples
- [x] Added test


bpatel2107:
Ready for review

pyrokinetics/gk_code/GKOutputReaderTGLF.py
```diff
@@ -0,0 +1,383 @@
...
+        return {
+            "input": TGLFFile(dirname / "input.tglf", required=True),
+            "run": TGLFFile(dirname / "out.tglf.run", required=True),
+        }
+
+    def verify(self, dirname: PathLike):
+        dirname = Path(dirname)
+        for f in self._required_files(dirname).values():
+            if not f.path.exists():
+                raise RuntimeError
```
ZedThree, line 34: Is this error ever seen by the user? Could perhaps do with a more verbose message?

pyrokinetics/gk_code/GKOutputReaderTGLF.py
```diff
@@ -0,0 +1,383 @@
...
+        Args:
+            raw_data (Dict[str,Any]): Dict containing TGLF output.
+            gk_input (GKInputTGLF): Processed TGLF input file.
+
+        Returns:
+            xr.Dataset: Dataset with coords and attrs set, but not data_vars
+        """
+
+        if gk_input.is_linear():
+            f = raw_data["wavefunction"].split("\n")
```
ZedThree: No need to change, but `str.splitlines` is a bit more explicit

pyrokinetics/pyroscan.py
```diff
@@ -253,6 +253,12 @@ def load_gk_output(self):
 
         # xarray DataSet to store data
         ds = xr.Dataset(self.parameter_dict)
+        if self.base_pyro.gk_code == "TGLF":
+            nmode = self.base_pyro.gk_input.data.get("nmodes", 2)
+            nmode_coords = {"nmode": list(range(1, 1 + nmode))}
+            ds = ds.assign_coords(nmode_coords)
+        else:
+            nmode = np.nan
```
ZedThree, line 264: We should really try to not have GK code specific bits outside of the readers/writers. Instead, perhaps this needs to be a "capability" of the GK codes, or some property: `self.base_pyro.gk_code.has_nmodes`

Or perhaps the TGLF data needs to be transformed into something much more like the other codes? I'm not exactly sure what `nmodes` is here, and why this needs to be handled differently

ZedThree (approved):
LGTM, but I think we should do something about the bit in pyroscan

pyrokinetics/gk_code/GKOutputReaderTGLF.py
bpatel2107: If I'm honest I just copied this over from `GKOutputReaderCGYRO`. @LiamPattinson  thoughts?

pyrokinetics/gk_code/GKOutputReaderTGLF.py
bpatel2107: That is better, makes looking for the eigenvalues in `out.tglf.run` a bit cleaner

pyrokinetics/pyroscan.py
bpatel2107: TGLF is an eigensolver and you can request any number of modes for it to find. However, it might not find any modes and in that scenario the outputted data changes. 
 
Mostly we handle initial value solvers so it could be good to distinguish between the two. Add a property like 
`pyro.gk_code.solver_type = 'initial_value'`

Probably best for a separate PR

ZedThree: Ah, ok, that makes sense. GS2 can also run as an eigensolver, so this would be a good thing to generalise.

Agreed, another PR is fine. Could you leave a big `FIXME/TODO` or something here just explaining what this bit is doing and that it needs generalising?

bpatel2107: Raised #93

pyrokinetics/gk_code/GKOutputReaderTGLF.py
LiamPattinson: The `verify` function is mostly used for file-type inference, so the raised exception is just there to tell the output reader factory, "This path doesn't look like a directory containing TGLF files, try a different output reader". It might be worth adding more precise error messages to the verify functions throughout Pyrokinetics in case users find another use for them, but I don't think it's a high priority.

Ideally, we could have something like the following for each class with a verify function:
```python
# Maybe @staticmethod?
def verify(self, path, throw=True):
    try:
        # check that the file is valid, raise exception otherwise
        return True
    except Exception as exc:
        if throw:
            raise VerifyPathException(
                f"The path {path} is not a ... file"
            ) from exc
        return False
```

PR #95, opened 2022-07-15
New GKCode GFTM

bpatel2107:
Add in new GKCode GFTM

This has the same input/output format as TGLF. This results near identical code for pyro. Literally I just copied all the TGLF files and replaced "tglf" with "gftm" which did the job. It is a different eigensolver. 

I am leaning toward this being a separate GKCode even with all the redundancy as they have completely different source codes.

The identical input files makes the auto-verification of gk_code quite difficult. Might need to look at the file prefix too?

Thoughts @ZedThree @LiamPattinson @d7919


LiamPattinson:
If these classes are to be identical besides file extensions, then the following should work and it won't require any duplicated code:

```python
class GKInputGFTM(GKInputTGLF):
    
    der verify(self, filename: Path):
        # Check file extensions for GFTM, also check for expected keys
        # Note: also check file extensions in the TGLF verify!
    
    # No other methods needed

# Similar for the output reader...
```
This inheritance hierarchy might not make much sense, as then `isinstance(gk_input, GKInputTGLF)` will return `True` for `GKInputGFTM`, so perhaps instead it would be better to do the following:

- rename `GKInputTGLF` to something horrible like `GKInputTGLFBase`
- set its `verify` function to an `@abstractmethod` so you can't instantiate the class directly
- Have both `GKInputTGLF` and `GKInputGFTM` inherit `GKInputTGLFBase` and implement their own `verify` functions
- Go through a similar process for the output readers

The simplest option might be to note in the docs that TGLF bits work just fine for GFTM, and to register TGLF classes to the factories twice with different keys:

```
gk_inputs["TGLF"] = GKInputTGLF
gk_inputs["GFTM"] = GKInputTGLF
```

I think this would just require updating the `verify` method for the output reader to also check for GFTM file extensions. This would allow the user to pass `gk_code="GFTM"` to `Pyro` calls without issues.

There could be some weird side effects to this. File type inference could take longer, as if it fails on TGLF it'll check again on GFTM, even though it'll be looking for the exact same info. Setting up a `Pyro` instance with a TGLF input and then converting to GFTM will go through a complicated conversion process that might lose info, despite the two being identical (though that'll be the case regardless of which of these methods we choose).

bpatel2107:
The output files for tglf are of the form `out.tglf.extension` and for GFTM it is `out.gftm.extension` so we could have the code name be used in the Readers and then just change the `code_name` accordingly. You could  look for `tglf` or `gftm`  in the verify too to distinguish between them.

bpatel2107:
Not sure this is even worth pursuing. That code isn't used at the moment is likely to change. Can add it in once relevant

PR #97, opened 2022-08-01
Normalisation refactor

ZedThree:
Refactor of #85

Replaces `NOCOS` with `NormalisationConvention`. The different codes' conventions are described as modifications to the base `pyrokinetics` convention:

```python
NORMALISATION_CONVENTIONS = {
    "pyrokinetics": NormalisationConvention("pyrokinetics"),
    "cgyro": NormalisationConvention("cgyro", bref_type="Bunit"),
    "gs2": NormalisationConvention("gs2", vref_multiplier=np.sqrt(2)),
    "gene": NormalisationConvention("gene", lref_type="major_radius"),
}
```

I switched these to named conventions as being a bit more user-friendly. It's now a bit easier to add new ones, so I think it would be fine to have e.g. `stella` be the same as `gs2` in practice.

I've renamed `LocalNorm` to `Normalisation`, but if pyrokinetics gains the ability to handle global codes, the original name might be better.

`Normalisation.from_kinetics` and `from_local_geometry` are class methods that return a new instance instead of modifying the existing one. I'm not completely happy about this, because really we need both at the same time. I think this is sort of just a symptom of how we can create a `pyro` object piecemeal from kinetics and geometry.


pyrokinetics/local_species.py
```diff
@@ -61,33 +60,20 @@ def from_dict(self, species_dict, **kwargs):
...
+        cls, kinetics: Kinetics, psi_n: float, local_norm: Normalisation
+    ):
         # TODO this should replace from_kinetics
         local_species = cls()
-        local_species.from_kinetics(kinetics, psi_n=psi_n, lref=lref)
+        local_species.from_kinetics(kinetics, psi_n=psi_n, local_norm=local_norm)
         return local_species
 
-    def from_kinetics(self, kinetics, psi_n=None, lref=None):
+    def from_kinetics(self, kinetics, psi_n, local_norm):
```
ZedThree: I'm not actually convinced that passing in `local_norm` is correct here -- previously it was generated to be consistent with `kinetics`, although `lref` had to be passed in.

Maybe it should be optional, and if `None`, then it uses `kinetics` for the normalisation?

pyrokinetics/pyro.py
```diff
@@ -1338,12 +1322,20 @@ def load_local_species(self, psi_n: float, a_minor: Optional[float] = None) -> N
                     "load_global_equilibrium."
                 )
 
+        self.local_norm = Normalisation.from_kinetics(
+            self.kinetics,
+            psi_n=psi_n,
+            convention=self.local_norm.nocos.name,
+            lref=a_minor,
+            bref=self.local_norm.bref,
```
ZedThree: Here we need to reuse `bref` from that calculated during `load_geometry` above. Looking at this with fresh eyes, this does put a dependency on the order in which `load_geometry` and `load_local_species` should be called.

pyrokinetics/local_species.py
bpatel2107: `lref` isn't stored in `kinetics` so how else would we get it? We could directly pass `lref` by doing

```python
local_species.from_kinetics(kinetics, psi_n=psi_n, lref=local_norm.lref)
```

pyrokinetics/pyro.py
bpatel2107: Yes this is true. We should add a check to make sure `local_geometry` is in before `local_species` is loaded

bpatel2107:
We'll need to add in TGLF as well as the outputs.

I'm also not sure about how to actually implement these changes when writing to a file. Do we store everything internally using the pyrokinetics convention and then convert when writing the files or do we store everything in the requested convention and then write the data as is.

The latter requires the geometry/local_species to be re-calculated. Like right now these changes have no impact on `local_geometry.Rmaj` which technically should be `Rmajor / Lref`

bpatel2107:
> 
> I'm also not sure about how to actually implement these changes when writing to a file. Do we store everything internally using the pyrokinetics convention and then convert when writing the files or do we store everything in the requested convention and then write the data as is.
> 

@ZedThree  @d7919 thoughts?

d7919:
I'm probably not familiar enough with the current code structure to make an informed comment, but my uninformed feeling is that always storing in pyrokinetics and then converting to the requested convention on write might be the cleanest.

bpatel2107:
I think I agree. This will require adding the normalisations to the readers/writers but I think it is worth it in the long run.

Only tricky detail would be setting values becomes awkward as you need to do it in pyro conventions and then let pyro change it for you. E.g setting beta in CGYRO requires knowing to multiply by `(B0/Bunit)**2`

d7919:
If you're using a setter, rather than directly accessing variables I suppose it may be possible to add an optional 'convention' input to specify what convention you're using when you pass something?

bpatel2107:
Would that be quite a large refactor?

d7919:
I would guess it could be, although maybe someone knows some magic to achieve this nicely

bpatel2107:
@ZedThree if we want to store everything internally with one normalisation (pyrokinetics by default), do you know of a clever way of reading/writing the data and always applying the correct normalisation? 

Or would we need to go through each variable and apply the normalisation (which would could be done using the `NormalisationConventions`) Something like 

```python
gs2['rmaj'] = miller.rmaj * NormalisationConvention.convert("pyrokinetics", "gs2", "lref")
```

ZedThree:
It's worth looking at [pint](https://pint.readthedocs.io/en/stable/index.html) for handling units. Then we can rely on that to do the conversions for us. There's a pint plugin for xarray too.

ZedThree:
I've spent today playing about with `pint`, and I think I can get it working like this:

```python
# Add units to Rmaj
Rmaj = pyro.local_geometry.Rmaj * ureg.metre

# Convert to GENE normalisation:
print(Rmaj.to("gene_lref"))
# 1.0 gene_lref

# In a given context, we can drop the convention name from the reference values
with ureg.context("gs2"):
    print(Rmaj.to("lref"))
# 1.233172498241725 lref
```

With this approach, we always create all the normalisation conventions, and then it doesn't matter what convention we store values in, we can always trivially convert to some another convention:

```python
Rmaj.to("gs2_lref").to("pyrokinetics_lref").to("gkdb_lref").to("gene_lref")
```
(although from my testing this does incur some rounding errors)

What I haven't been able to get working nicely is where something needs to be normalised to some combination of reference values, doing that without repeating the normalisation. That is, this works:

```python
with ureg.context("gene"):
    # Normalised to GS2
    nu = 5.677 * ureg.gs2_lref / ureg.gs2_vref
    print(nu.to("lref / vref"))
# 3.255217905458966 lref / vref
```

But getting something like `nu.to("gs2")` seems much more difficult.

We would need to make sure things were in the correct units first, before we could use this to normalise them, but I'm not sure how far we'd need to go. It would probably be nice if _everything_ had units, then we could also handle converting growth rates into say, Alfvén wave transit times or whatever.

bpatel2107:
I think we can add units to everything with relative ease, just will take time.

Do you think we could make a light wrapper around pint that handles cases like `nu.to("GS2")` that could handle these cases. Most inputs are going to be a combination of reference values so it would be good in the long run

ZedThree:
I had to explore a few dead-ends and investigate several libraries, but I now have something that mostly works, it just has lots of knock-on effects that need to be fixed. But we can now have things like:

```python
In [3]: pyro.local_species.electron.nu
Out[3]: 0.050767795211818026 <Unit('vref_pyrokinetics_run0000 / lref_pyrokinetics_run0000')>

In [4]: pyro.local_species.electron.nu.to(pyro.norms.gs2)
Out[4]: 0.03589825226016644 <Unit('vref_gs2_run0000 / lref_gs2_run0000')>

In [5]: pyro.local_species.electron.nu.to_base_units()
Out[5]: 20311.324112597606 <Unit('1 / second')>

In [6]: pyro2.local_species.electron.nu.to(pyro.norms)
Out[6]: 0.0183350430079753 <Unit('vref_pyrokinetics_run0000 / lref_pyrokinetics_run0000')>

In [7]: pyro.local_species.electron.nu.to(pyro2.norms)
Out[8]: 0.03942599128526476 <Unit('vref_pyrokinetics_run0001 / lref_pyrokinetics_run0001')>
```
(where `pyro` and `pyro2` are different simulations)

We could in principle define other units and conventions, like the cyclotron frequency and Larmor radius.

Each `Pyro` object gets a unique name which is used in turn to give each combination of simulation and convention a unique name for each unit. The stem can be set manually, taken from one of the input files, or the generic fallback "run". We create these units for all the conventions at once through a top-level `SimulationNormalisation` object, owned by each `Pyro` instance, which has a `ConventionNormalisation` for each convention. The `SimulationNormalisation` object has `lref`, `bref`, etc members which point to the current `default_convention`'s specific units. So `pyro.norms.lref` points to `pyro.norms.pyrokinetics.lref`, which points to the actual `lref_pyrokinetics_<run_name>`, by default.

It's still failing a couple of tests, mostly issues around PyroScan and copying these normalisation objects. There's also lots of places that now need the normalisation and units adding. Hopefully I'll be able to finish it off next week.

Some important caveats and downsides:

- I've picked `pint`, but there is also `astropy.units` and `unyt`. `plasmapy`, which would be really interesting/useful to incorporate, uses `astropy.units`. Unfortunately, that doesn't play nice with `xarray`, whereas `pint` does. If we do end up using `plasmapy`, we can write something to wrap their functions and convert the unit systems
- It's not possible to convert dimensionless quantities like `beta` between normalisations. I _think_ it would be possible to re-jig things to make this possible, by giving it a new dimension like `[beta_gs2]` and having explicit conversion functions. Currently, the different normalisations have their own `beta` value, like `norms.gs2.beta` and `norms.cgyro.beta`, that need to be accessed explicitly.
- It's not currently possible to convert between different codes without having these normalisation structures created from the geometry/kinetic data. I think this would be possible in the future, similar to the dimensionless quantities, by using parameterised contexts, but this has some severe footguns
- When a variable becomes a `Quantity` (i.e. has some units), they might need to be converted back to a normal number with `value.magnitude` or `value.m` for short

Still needs some specific tests for the normalisation stuff, and some proper documentation on how this works. It might also be nice to have something like `pyro.norm_to(pyro.norms.gs2)` or similar, to convert a whole simulation to a different convention, but that can come later if needed.

@bpatel2107 Do you want to have a play with this version and see if if breaks in any unexpected ways? It would also be great if we could add units to as many things as possible, but that can probably also be done piecemeal. E.g. adding units to the geometry and everything downstream from that.

ZedThree:
Ok, I have spent way too long on this, but it does now work and pass all the tests. There's probably more places to add units, but the basics are covered. This is ready for review!

There's a bunch of docs, but basically, there's now a set of base "simulation units" that are defined in terms of fictional dimensions, such as `lref_minor_radius = [lref]`, or other simulation units, such as `lref_major_radius = <aspect ratio> lref_minor_radius`. Then there are a set of conventions that collate the appropriate units, e.g. GS2 has `lref = lref_minor_radius`, GENE has `lref = lref_major_radius`.

We can then provide physical reference values to get physical units that have unique names per simulation, and are defined in SI units, so `lref_minor_radius_run1234 = 1.5 metres`. We update each convention's units for a given simulation to the physical unit instead of the simulation unit.

So now we can use the conventions to convert units:

```python
growth_rate_gene = growth_rate_gs2.to(norms.gene)
```

It could definitely still use some polish in places, and I'm sure this is not the most elegant solution, but it's a decent start at least. 

Better names for anything/everything also welcome. For example, the velocity units are named according to the methods in the [plasmapy docs](https://docs.plasmapy.org/en/stable/api/plasmapy.formulary.speeds.thermal_speed.html#thermal-speed-notes), `vref_nrl` and `vref_most_probable` -- which are not really very good names!

---

The latest version of numpy and pint don't support Python 3.7, so rather than fight the CI, I've just dropped it...

pyrokinetics/gk_code/GKInputGENE.py
```diff
@@ -120,8 +129,6 @@ def get_local_geometry_miller(self) -> LocalGeometryMiller:
```
bpatel2107, line 129: Here we are reading in the data and working out the pyro normalisation. This should be fine when we are writing out new files and running new simulations. But if we wanted to load in an existing simulations that has `lref=major_radius` (which corresponds to `self.data["geometry"]["major_r"] = 1.0` then pyro wouldn't know how to normalise the outputs. I think when we add in the normalisation for the outputs we need to check the input file to see which normalisation was used.

bpatel2107:
Overall it looks pretty good so far. We need to add in the output normalisations too. Do you want to leave that for a separate PR or include it in this one?

pyrokinetics/gk_code/GKInputGENE.py
ZedThree: I was a bit confused by this -- if I understand correctly, GENE input files always (sometimes?) have both `major_r` _and_ `minor_r` set, and then Pyrokinetics will renormalise it so that `major_r == 1.0`? I don't have access to the GENE manual -- are these in metres, or normalised units?

With this PR, reading in any GENE file will result in lengths being defined in terms of `lref_major_radius`, and writing back out to GENE (_should_) be idempotent (i.e. not change anything physically), and to convert to other codes we need either a) both `Rmaj` and `a_minor`, or b) the aspect ratio. In practice, we seem to always have `Rmaj` set and only need to supply `a_minor`.

We currently only set `Rmaj` in the Miller geometry from GENE input files, but maybe we should also set `a_minor`? This way, we'd always have the correct aspect ratio used in GENE and so wouldn't need to supply it. How many other codes include the aspect ratio? I believe GS2 can take it as an option (at least two different ways, I'm sure).

I did consider adding the following to `LocalGeometry`:

```diff
 
+    @property
+    def aspect_ratio(self) -> float:
+        return self.Rmaj / self.a_minor
+
+    @aspect_ratio.setter
+    def aspect_ratio(self, value: float):
+        # TODO: Does it make sense to change the aspect ratio if both
+        # R and r are already set?
+        if hasattr(self, "Rmaj"):
+            self.a_minor = self.Rmaj / value
+        elif hasattr(self, "a_minor"):
+            self.Rmaj = self.a_minor * value
+        else:
+            raise RuntimeError(
+                "Cannot set aspect_ratio: missing both minor and major radius"
+            )
+
+    @property
+    def inverse_aspect_ratio(self) -> None:
+        return 1.0 / self.aspect_ratio
+
+    @inverse_aspect_ratio.setter
+    def inverse_aspect_ratio(self, value: float) -> None:
+        self.aspect_ratio = 1.0 / value
+
     # TODO replace this with an abstract classmethod
```

ZedThree:
I think it would be better to get this in so people can play around with it and find the bugs, and then add more units after.

To convert between simulation and physical units, we need to have a pint `Context` with an explicit transformation, and pint won't transform between mixed units. For example, there's a context (automatically enabled in this PR) to convert between `eV` and `K`, but pint can't convert `eV / m` to `K / m`. So I've add to explicitly add some transformations for what I think are the most common mixed units: `[vref] / [lref]` and `[nref] * [tref]` for instance. If people regularly use other mixed units, we'll need to add them.

bpatel2107:
Can we add in a setter for beta in the instance where we are loading in a local gk file. 

For example we set `miller.B0` from a local gk file. This sets `bref` but we have no way of setting `nref` or `Tref` so the line below will fail. 

https://github.com/pyro-kinetics/pyrokinetics/blob/e16df125fad7ece5b7e756d42135e30727e4c83c/pyrokinetics/normalisation.py#L642-L651

We need to try the method above and if that fails then check if only `bref` exists and then define beta as `beta = 1 / bref**2`

ZedThree:
In that case, I think we'd still need to keep `nref` and `tref` in there, as otherwise you wouldn't be able to tell the difference between the betas for simulations that did or didn't have nref/tref. For instance, you'd read in a GK file, get one dimensionless value for beta, then add the physical reference values and get a different, but still dimensionless, beta value.

I think I can probably hack this for now so that it works like you want, but with ugly units.

bpatel2107:
Technically if beta is already defined then `bref` should be calculated when setting `tref` and `nref`. Maybe we can add a check to see if the value of beta is consistent with this or not.

ZedThree:
Ah, good point. Hmm, that probably complicates things a bit, as pint doesn't really handle dimensionless quantities very well.

I rather suspect that means we'll need a `[beta]` dimension. Not sure about transformations to/from `bref`, hopefully just the consistency check as you suggest will do

bpatel2107:
I think we can have a beta scale factor out the front that gets set such that when we load in the reference values the actual value used in simulations doesn't change and a warning is printed if it is not 1

ZedThree:
It looks like only `input.gene` in the example templates has `beta` set -- do you know which kinetic and geometry files are consistent with that simulation so that I can test this?

Setting beta from a GK simulation might be a little ugly. We need to set it for all the normalisation conventions (which at the moment is really just the choice of B0 luckily; but in general we also have choices of nref/tref too), which means we need to know which convention we're setting it for.

I'm not sure about the scale factor, my plan was to check for consistency when we update any of the physical reference values:

```python
    def _check_beta_consistency(self):
        # If beta hasn't been set, it can't be inconsistent
        if self.beta is None:
            return

        # If we've not set all the reference values, then it can't be inconsistent
        if not all(map(_is_set, [self.bref, self.nref, self.tref])):
            return

        # Check that explicitly set value agrees with calculated value
        if np.isclose(self._beta, self._calc_beta()):
            return

        warnings.warn(
            f"Explicitly set value of beta ({self._beta}) is inconsistent with "
            f"value from physical reference values ({self._calc_beta()})"
        )
```

bpatel2107:
There will be cases when people want to do a beta scan so we need some way to modify the beta used in a simulation. 

We can figure out which species is the reference species by their values. i.e. if `local_species.electron.temp = 1.0` then tref = te and so on.

This is why a scale factor might be best as it separates the reference beta which we can define as `beta_ref = 8*pi*nref*tref / bref**2` and in a simulation we do `gk_code.beta = beta_multiplier * beta_ref` if it is defined.

This way `beta_ref` becomes a unit and the GK `beta` is just relative to that value.

I do like the idea of a warning when not consistent though

ZedThree:
Sure, we definitely want to support people running beta scans.

Where does all this live though? I'm not sure about putting `beta_multiplier` into the norm classes, as then they're not purely reference values and units. Having it elsewhere though will make it really difficult to convert beta between conventions -- we can't make it a simple unit, because it's dimensionless and therefore impossible to convert automatically. We can do it manually, of course, provided we do know which conventions we converting to/from.

It might be a bit easier in some ways if we did give beta its own dimension, but working out all the different conversions becomes difficult then too, because pint is very awkward for converting dimensions. We'd also need a different dimension for each combination of `(nref, tref, bref)`.

I'll see if I can get it work with separate dimensions, as that will give us the most flexibility in terms of how to store the actual simulation value.

bpatel2107:
To be honest its not simple. Right now beta sneakily lives inside the `local_geometry` via `B0` but its not the best place for it.  

There may be need to have a new object that stores these kind of parameters that need reference values. For now we can think about storing `beta_multiplier` in the base pyro or the numerics object?

ZedThree:
Yes, maybe in numerics might make most sense?

I _think_ I can make it work by defining a dimension and setting up the "other" conventions as units, basically the same as I've done for all the other reference values. Then we probably only need to hardcode changing the units from dimensionless to `[beta_ref]`

ZedThree:
I think the `beta_multiplier` is a bit too complicated. Here's what I think are all the possible situations, and what variables _could_ be defined in each case:

| | `beta_ref` | `1/B0**2` | `beta` | `beta_multplier` |
|--|--|--|--|--|
| geometry + kinetic | :heavy_check_mark: | :heavy_check_mark: | :x: | :x: |
| GK + geometry + kinetic | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| GK + geometry | :x: | :heavy_check_mark: | :heavy_check_mark: | :white_check_mark: |
| GK + kinetic | :x: | :x: | :heavy_check_mark: | :x: |
| GK | :x: | :x: | :heavy_check_mark: | :x: |

where `beta` is the value actually used in the simulation.

We can always set `beta` for a GK simulation, even if it is done so arbitrarily. `beta_multiplier` can only really be set in the "full physics" case. I think it would be easiest to set `numerics.beta` from the GK simulation and warn if it's not consistent with `beta_ref`. If we have units like `beta_ee_B0`/`beta_ee_Bunit`, we can then automatically convert (or error!) to the correct convention.

Is `1/B0**2` really used very often as an approximation to `beta`? It feels very off to me: the units are wrong, it looks like it's saying that `n*T ~ 1/(2*mu0)`, and cannot be kept consistent between a GK simulation and `beta_ref`.

bpatel2107:
The `1/B0**2` was not physically motivated and I am open to changing this. I set it to that as it was the closest thing I could come up with that was sort of physically relevant though I do see how it isn't great.

I am fine with setting the GK `beta` in the `numerics` and having this warning pop up. In the case where we do have `nref`, `tref` and `bref` and we use a template file for a GK input we should set `beta` to `beta_ref` by default.

ZedThree:
Looking at how GKDB/IMAS handles it, they have `beta_reference` in `species_all`, which I think corresponds to our `LocalSpecies`.

I'm just going to stick it into `numerics` for now, and we should have a bigger project of renaming and moving things to better align with IMAS later.

bpatel2107:
I think numerics is better as you can't get beta from the local species data alone. So you wouldn't be able to fully load in `LocalSpecies` from just a SCENE/TRANSP output.

This along with `rho_star` (which isn't stored yet) are the only terms I can think of that need both local species data and local geometry data so shouldn't really be stored in either one alone.

Once this is in we can discuss how handle it going forward

ZedThree:
Now you can change `pyro.numerics.beta` however you like, and that's what will get written to file -- unless it's zero, in which case it will fallback to the reference beta value. Is that actually what we want though? How do we handle electrostatic cases? Set it to `None` maybe?

bpatel2107:
I think if `pyro.numerics.beta =  0.0` then the GK input file should be 0. You can imagine doing a scan from 0 to 1 and you would want this behaviour.

If it is `None` then it should try and set it to the reference value and if it can't do that then set it to 0

ZedThree:
The last commit actually caught a bug: collision frequencies were not being converted properly. Unfortunately, this means we basically always need both geometry and kinetic files in order to convert GK files. This is because `nu ~ vref / lref`, and `lref` comes from the geometry, while `vref` comes from the kinetics. In principle we should be able to convert `nu` if we know the aspect ratio alone, but unfortunately pint is very annoying when it comes to mixed units and I can't make it work.

The workaround is to just zero the collision frequency for all species, or to convert to the correct units manually.

Similarly for `beta`, where we can read it in from a GK file, but cannot convert it between conventions without the geometry (for `bref`)

bpatel2107:
We do know the relative normalisations between them. So do you think we could use a temporary set of normalisations whilst converting. It shouldn't matter what they actually are as we need a placeholder.

So if bref is None, then temporarily use an arbitrary bref, do our conversions and the set bref back to None. Or just have a conversion normalisation object that uses what it can for the reference values but is completely separate.

Once normalised again it doesn't actually matter what the reference values are so this should work

ZedThree:
Ok, I got it working to convert frequencies when we only know the aspect ratio

bpatel2107:
Can you do something similar for `bref` to convert from B0 normalisation to Bunit normalisation?

ZedThree:
That already works! Although currently you need to call `set_bref` with a `LocalGeometry` -- would be trivial to add `B0` and `bunit_over_b0` arguments though

bpatel2107:
Okay awesome! Is this good to merge now?

ZedThree:
Yep! It needs some real-world testing now, and then we can start to add units to the rest of the code

PR #98, opened 2022-08-08
Enable reading more kinds of GS2 simulations

### LINKED ISSUE #94: gs2 code calculated frequency and growth rates to be included into pyro structure
There are some cases when running gs2 that, due to selection of frequency of mode output frequencies, the Pyrokinetics and gs2 calculated linear mode frequencies don't agree.  Could the gs2 calculated values be included for the possibility of cross checking or using this other calculation?

ZedThree:
- enables "range" grids, as well as "box" for linear simulations
- enables reading `omega_average` if the time-dependent fields aren't present

Closes #94 

The normalisation for the growth rate and frequency probably want checking by an expert!

This let me read and do some basic analysis on more kinds of GS2 runs, but it almost certainly has some rough edges, and will need some work to generalise a bit more.

### DIFF
--- a/pyrokinetics/gk_code/GKInputGS2.py
+++ b/pyrokinetics/gk_code/GKInputGS2.py
@@ -217,7 +217,8 @@ class GKInputGS2(GKInput):
             "nky": 1,
             "nkx": 1,
             "ky": self.data["kt_grids_single_parameters"]["aky"] / sqrt2,
-            "kx": self.data["kt_grids_single_parameters"].get("akx", ky * shat * theta0) / sqrt2,
+            "kx": self.data["kt_grids_single_parameters"].get("akx", ky * shat * theta0)
+            / sqrt2,
             "theta0": self.data["kt_grids_single_parameters"].get("theta0", 0.0),
         }
 
--- a/pyrokinetics/gk_code/GKInputGS2.py
+++ b/pyrokinetics/gk_code/GKInputGS2.py
@@ -217,7 +217,7 @@ class GKInputGS2(GKInput):
             "nky": 1,
             "nkx": 1,
             "ky": self.data["kt_grids_single_parameters"]["aky"] / sqrt2,
-            "kx": 0.0,
+            "kx": self.data["kt_grids_single_parameters"].get("akx", ky * shat * theta0) / sqrt2
             "theta0": self.data["kt_grids_single_parameters"].get("theta0", 0.0),
         }
 
--- a/pyrokinetics/gk_code/GKInputGS2.py
+++ b/pyrokinetics/gk_code/GKInputGS2.py
@@ -212,6 +212,99 @@ class GKInputGS2(GKInput):
         local_species.normalise()
         return local_species
 
+    def _read_single_grid(self):
+        return {
+            "nky": 1,
+            "nkx": 1,
+            "ky": self.data["kt_grids_single_parameters"]["aky"] / sqrt2,
+            "kx": 0.0,
+            "theta0": self.data["kt_grids_single_parameters"].get("theta0", 0.0),
+        }
+
+    def _read_range_grid(self):
+        range_options = self.data["kt_grids_range_parameters"]
+        nky = range_options.get("naky", 1)
+
+        ky_min = range_options.get("aky_min", "0.0")
+        ky_max = range_options.get("aky_max", "0.0")
+
+        spacing_option = range_options.get("kyspacing_option", "linear")
+        if spacing_option == "default":
+            spacing_option = "linear"
+
+        ky_space = np.linspace if spacing_option == "linear" else np.logspace
+
+        ky = ky_space(ky_min, ky_max, nky) / sqrt2
+
+        return {
+            "nky": nky,
+            "nkx": 1,
+            "ky": ky,
+            "kx": np.array([0.0]),
+            "theta0": 0.0,
+        }
+
+    def _read_box_grid(self):
+        box = self.data["kt_grids_box_parameters"]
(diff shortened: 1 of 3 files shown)


pyrokinetics/gk_code/GKInputGS2.py
```diff
@@ -212,6 +212,99 @@ def get_local_species(self):
         local_species.normalise()
         return local_species
 
+    def _read_single_grid(self):
+        return {
+            "nky": 1,
+            "nkx": 1,
+            "ky": self.data["kt_grids_single_parameters"]["aky"] / sqrt2,
+            "kx": 0.0,
```
d7919: Should this be `kx = ky*shat*theta0`?

pyrokinetics/gk_code/GKInputGS2.py
```diff
@@ -212,6 +212,99 @@ def get_local_species(self):
...
+        if spacing_option == "default":
+            spacing_option = "linear"
+
+        ky_space = np.linspace if spacing_option == "linear" else np.logspace
+
+        ky = ky_space(ky_min, ky_max, nky) / sqrt2
+
+        return {
+            "nky": nky,
+            "nkx": 1,
```
d7919, line 242: This could be >1

pyrokinetics/gk_code/GKInputGS2.py
```diff
@@ -212,6 +212,99 @@ def get_local_species(self):
...
+            "ky": ky,
+            "kx": np.array([0.0]),
+            "theta0": 0.0,
```
d7919, line 245: This could be /= 0

pyrokinetics/gk_code/GKInputGS2.py
```diff
@@ -212,6 +212,99 @@ def get_local_species(self):
...
+    def _read_box_grid(self):
+        box = self.data["kt_grids_box_parameters"]
+        keys = box.keys()
+
+        grid_data = {}
+
+        # Set up ky grid
+        if "ny" in keys:
+            grid_data["nky"] = int((box["ny"] - 1) / 3 + 1)
+        elif "n0" in keys:
```
d7919, line 257: `n0` is not related to the number of modes, but sets the toroidal mode number corresponding to the first non-zero ky. The number of ky should always be set by ny or naky here I think.

pyrokinetics/gk_code/GKInputGS2.py
ZedThree: (Note that this is unchanged from existing implementation)

Maybe:

```suggestion
            "kx": self.data["kt_grids_single_parameters"].get("akx", ky * shat * theta0) / sqrt2
```

We also don't handle being able to set `n0` and `rhostar_single` here, but it looks like we probably actually _can't_?

pyrokinetics/gk_code/GKInputGS2.py
ZedThree: Good point, I left this for the time being, because I haven't quite got my head round how the `kx` grid is set. It looks like there's at least four different ways

pyrokinetics/gk_code/GKInputGS2.py
bpatel2107: We store `theta0`, this probably should be what is suggested. Maybe we should store one or the other

pyrokinetics/gk_code/GKInputGS2.py
bpatel2107: When we have a box with multiple `kx` and `ky` we should use the method in the nonlinear case. This is specifically for single mode analysis. Might need better logic for this though

ZedThree: Sorry, just to be clear and so we're all on the same page, I've rearranged things a little to be able to handle three different GS2 grids:

- `single`, which was what the linear case previously used
- `range`, which is new (to pyrokinetics) and another option for linear runs, and could have multiple `kx` modes
- `box`, which is what the nonlinear case previously used

bpatel2107: Is `nkx` basically set by `gs2['kt_grids_range_parameters']['ntheta0']`

pyrokinetics/gk_code/GKInputGS2.py
bpatel2107: We don't really need both `kx` and `theta0`

pyrokinetics/gk_code/GKInputGS2.py
bpatel2107: Yes, `n0` can be used to set `ky` via `ky=n0*drhodpsi*rhostar_single`

bpatel2107:
For linear runs we should decide which to store, `kx` or `theta0` @d7919 any thoughts in which is preferable?

`kx` is more consistent with the nonlinear data, but the `theta0` is generally more useful

d7919:
> 

What do other codes use? I suppose whichever we choose we should be able to add a getter to provide the other one?
I'd probably prefer theta0 but happy either way really.

bpatel2107:
Agreed about the getter/setter. Maybe we can keep both?

CGYRO doesn't quite have this. You can have a nonlinear grid with the nonlinear term turned off but that's not really the same. Not sure about GENE

Both can store `theta0` so maybe we should stick with that. Could even have a method to switch the coordinates. I think `assign_coords` in xarray might allow for this?

bpatel2107:
I think the `theta0` vs `kx` is a different issue so I'm going to merge this PR and open a new issue for that

PR #99, opened 2022-08-09
'supported types' lists now properties, not class-level vars

LiamPattinson:
We earlier implemented a plugin architecture for `GKInput`/`Equilibrium`/`Kinetics` etc, but on testing, I found it didn't work as intended for user-created plugins:

```python
import pyrokinetics as pk

# User creates a new GKInput class 
class MyGKInput( pk.gk_code.GKInput):
    # ... implement all the required methods here ...

# Register with gk_inputs factory
pk.gk_code.gk_inputs["MyGKInput"] = MyGKInput

# Try to read a file
pyro = pk.Pyro(gk_file="my_gkinput_file.txt")
``` 
The last step fails, as `Pyro.supported_gk_inputs` is defined at the class level, and is therefore instantiated back during the original `import`. It can't recognise the new plugin created by the user.

I've changed these class-level lists to be implemented as properties instead, meaning a new list is created from the factory classes each time the user accesses `pyro.supported_gk_inputs`, `pyro.supported_equilibrium_types`, etc. A simpler but less flexible fix could be to instead move the class-level lists inside the `__init__` function -- let me know if you'd prefer this.

### DIFF
--- a/pyrokinetics/equilibrium/Equilibrium.py
+++ b/pyrokinetics/equilibrium/Equilibrium.py
@@ -61,9 +61,6 @@ class Equilibrium:
         all readers, so only include this if necessary.
     """
 
-    # Define class level info
-    supported_equilibrium_types = [*equilibrium_readers]
-
     def __init__(
         self,
         eq_file: PathLike,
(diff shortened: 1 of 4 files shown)


bpatel2107:
This looks good to me! Should make development a bit easier

PR #103, opened 2022-08-10
field.h5 read capability added. 

ajaycj1:
If field binary file is absent, field .h5 file is read if present.
ntime is corrected to include the last entry.


ajaycj1 (commented):
I added a comment in pyro.py. Please ignore that.

PR #104, opened 2022-08-10
GENE field file .h5 read capability added. Corrections to open last field and nrg entries - Test files updated.

ajaycj1:


ajaycj1:
So far Pyrokinetics was not reading the last time step entry in field and nrg output files. It used to read only up to the second last entry. I corrected this. 
However the tests are not updated to account for this, hence the check failures.

pyrokinetics/gk_code/GKOutputReaderGENE.py
```diff
@@ -5,6 +5,7 @@
 import struct
 import csv
 import re
+import h5py #cj added
```
bpatel2107: Can you remove the cj added comment?

bpatel2107:
Looks like it's passing the tests! 

Can you remove the lines where you've written cj added/removed and then I can merge into unstable

ajaycj1:
> Looks like it's passing the tests!

Well, it is like passing a test where the questions were set by the same person!

bpatel2107:
Keep in mind this test is just to make sure the output files are being read in correctly. So we need to provide the right answer for to test against.

pyrokinetics/gk_code/GKOutputReaderGENE.py
```diff
@@ -339,10 +366,15 @@ def _set_fluxes(
                     ]
 
                 # Skip time/data values in field print out is less
-                if i_time != data.ntime - 1:
+                if i_time < data.ntime - 2:
```
bpatel2107, line 369: Isn't this going up to the third to last point? Is this because GENE always spits out final point regardless of time?

bpatel2107:
@ZedThree why isn't black/flake8 running for these cases. Is there a way to get them to run for merges outside of pyro?

Edit - Maybe it can't push to forked directories. Should we request people to branch off our unstable

pyrokinetics/gk_code/GKOutputReaderGENE.py
ajaycj1: Yes, GENE spits out the final step regardless, even if it is not integer multiples of istep_field or istep_nrg.

bpatel2107:
I'll merge and then this will probably fail black/flake8 but we can deal with that here

PR #107, opened 2022-08-12
GENE circular geometry read capability added

ajaycj1:


bpatel2107:
So this won't be able to write GENE inputs at `circular`. This would write it as `miller` and use the defaults, which is effectively the same thing. Do you see an issue with this @ajaycj1 

Essentially if you read in a file and then write it, it will be different, though functionally identically if I have understood correctly

ajaycj1:
You are right. 
Whether I see an issue with not writing circular as a different geometry and treating it as a case of miller - I think it is a choice.
1. In general, most of my choices were based on making the least amount of changes to the code as possible - this is probably not a good practice in the long term.
2. As far as I know, in most cases people use miller as the default analytic geometry, so I don't find a need to write circular explicitly as a different geometry.

bpatel2107:
I think I'll merge for now as this is an edge case and has no real impact on the actual output

ajaycj1:
I just saw in get_local_geometry_circular, I had written circular_data["local_geometry"] = "Circular" (I said the opposite earlier in this conversation). 
Again, it is a choice - correct it as required.

bpatel2107:
So it technically doesn't matter what you set that as, but for clarity I will change it to Miller

PR #112, opened 2022-08-18
Enhancements/LocalGeometry

bpatel2107:
Initial set up for Fourier geometry representation of equilibrium based off of 

_Plasma Phys. Control. Fusion 63 (2021) 012001 (5pp) https://doi.org/10.1088/1361-6587/abc63b 
Plasma Phys. Control. Fusion 51 (2009) 105009 J Candy
Phys. Plasmas, Vol. 5, No. 4, April 1998 Miller et al._ 

Necessary for interface with IMAS structure (like access via GKDB)

Used the following logic
image

Currently have a good fit of flux surface and following quantities
R
Z
dR/dθ
dZ/dθ
θR
dθR/dθ
image

Bpol fit isn't great at the moment
image

though I can't see any major flaws

To do

- [x] Base LocalGeometryFourier object 
- [x] Set up GKInputReader / writer with Fourier
- [x] Set up GKOutputReader with Fourier


bpatel2107:
@ajaycj1 @ZedThree 

This PR is needed to complete the GKBD interface

bpatel2107:
@dake0795 FYI

bpatel2107:
Just to be clear is does pretty well expect at the turning points in `theta`

image

pyrokinetics/local_geometry/LocalGeometryFourier.py
```diff
@@ -0,0 +1,662 @@
...
+    # plt.legend()
+    # plt.show()
+
+    # Plot dZR_dtheta
+    # plt.plot(theta, dZdtheta, label='Fit')
+    # plt.plot(theta, np.gradient(Z, theta), '--', label='Data')
+    # plt.xlabel('theta')
+    # plt.ylabel('dZ / dtheta')
+    # plt.legend()
+    # plt.show()
```
ZedThree: Commented out code

pyrokinetics/local_geometry/LocalGeometryFourier.py
```diff
@@ -0,0 +1,662 @@
...
+        )
+
+        thetaR = np.arccos(normalised_radius)
+
+        for i in range(len(theta)):
+            if R[i] < R_upper:
+                theta[i] = np.pi - theta[i]
+            else:
+                if Z[i] < 0:
+                    theta[i] = 2 * np.pi + theta[i]
```
ZedThree: We can get rid of a level of indentation:

```suggestion
            elif Z[i] < 0:
                 theta[i] = 2 * np.pi + theta[i]
```

The following is a bit shorter (and plausibly faster), but not 100% equivalent:

```suggestion
        theta = np.where(R < R_upper, np.pi - theta, theta)
        theta = np.where(Z < 0, 2 * np.pi + theta, theta)
```

pyrokinetics/local_geometry/LocalGeometryFourier.py
```diff
@@ -0,0 +1,662 @@
...
+
+        for i in range(len(thetaR)):
+            if Z[i] < 0:
+                thetaR[i] = 2 * np.pi - thetaR[i]
```
ZedThree: We can just use `np.where` here:

```suggestion
        thetaR = np.where(Z < 0, 2 * np.pi - thetaR, thetaR)
```

pyrokinetics/local_geometry/LocalGeometryFourier.py
```diff
@@ -0,0 +1,662 @@
...
+        self.theta = theta
+        self.b_poloidal = b_poloidal
+
+        self.q = float(q)
+        self.shat = shat
+        self.beta_prime = beta_prime
+        self.pressure = pressure
+        self.dpressure_drho = dpressure_drho
+
+        for n_moments in [4, 5, 6, 8, 10, 12, 14, 16]:
```
ZedThree: Is this something we might want to customise? Or is this a fairly standard set of moments?

pyrokinetics/local_geometry/LocalGeometryFourier.py
```diff
@@ -0,0 +1,662 @@
...
+
+            self.kappa = kappa
+            self.n_moments = n_moments
+            self.sym_coeff = sym_coeff
+            self.asym_coeff = asym_coeff
+
+            self.thetaR = self.get_thetaR(theta)
+            self.dthetaR_dtheta = self.get_dthetaR_dtheta(theta)
+
+            """
```
ZedThree: I'm guessing this is just here for debugging during development?

pyrokinetics/local_geometry/LocalGeometryFourier.py
```diff
@@ -0,0 +1,662 @@
...
+            )
+
+            plt.plot(self.theta, bpol_fit, label=f"N moments={n_moments}")
+
+        plt.plot(self.theta, self.b_poloidal, "--", label="Data", color="k")
+        plt.legend()
+        plt.xlabel("theta")
+        plt.title("Fit to poloidal field with different number of moments")
+        plt.ylabel("Bpol")
+        plt.show()
```
ZedThree: Just here for debugging? It might be nice to provide some separate plotting routines

ZedThree (commented):
LGTM. Do we a function for converting between Miller and Fourier?

pyrokinetics/local_geometry/LocalGeometryFourier.py
bpatel2107: All the plotting is definitely for debugging but I agree some routines might be nice

pyrokinetics/local_geometry/LocalGeometryFourier.py
bpatel2107: Quite a few functions use them so might be safer to keep them as attrs but not necessary

pyrokinetics/local_geometry/LocalGeometryFourier.py
bpatel2107: We can put it as a `kwargs` You only need the one but that is a choice. My only concern is that if codes only take 3 moments then we'll need to fit 3 moments. Not sure if you can just truncate a fourier series. There may be some thinking required here

pyrokinetics/local_geometry/LocalGeometryFourier.py
bpatel2107: We need theta to go from `0 -> 2 pi`, but `arcos` only return values between `0` and `pi` 

So we need to rotate some of the angles around depending on where it is on the flux surface. Here for example we need to rotate 3 of the 4 quadrants of a flux surface, firstly when
`R[i] < R_upper` which is the inboard side  (left quadrants)

and when both `R[i] > R_upper` and `Z[i] < 0.0` (bottom right quadrant)

The top right quadrant is already at the correct angle

bpatel2107:
Seems like a higher resolution EQDSK file works much better. Is likely an issue with a low resolution equilibrium not quite being a Grad-Shafranov solution.

image

pyrokinetics/local_geometry/LocalGeometryFourier.py
bpatel2107: This is good

ZedThree:
Ah, not terribly surprising the EQDSK needs more resolution to get a good agreement -- it's a pretty awful format really!

Does it make sense perhaps to give an estimate of the agreement somehow?

pyrokinetics/local_geometry/LocalGeometryFourier.py
bpatel2107: Used double conditional in where to do this

bpatel2107:
Lowered the threshold for poor quality fits which will warn the users

bpatel2107:
Turns out I was the problem, not the GEQDSK. Forgot that `kappa` was a function of `r`, so was missing a term in `dZdr`. Adding that in fixes the fit in the lower resolution GEQDSK

High res GEQDSK
image

Lower res GEQDSK
image

bpatel2107:
I can do the mapping from Fourier to CGYRO

@d7919 do you know how map these to GS2?

@ajaycj1 do you know how map these to GENE?

ajaycj1:
> @ajaycj1 do you know how map these to GENE?
Sorry, never did this.
I can of course learn how to do this, but I am a bit time-strapped - need to complete some things before I go on a short vacation. If I get time meanwhile, I can look into this, but unfortunately cannot guarantee at the moment.

d7919:
I think likely selecting geoType = 3 (https://gyrokinetics.gitlab.io/gs2/page/namelists/index.html#theta_grid_parameters-geotype)  and then table 7 of https://bitbucket.org/gyrokinetics/wikifiles/raw/master/JRB/GS2GeoDoc.pdf helps map the inputs to the quantities.

bpatel2107:
Well annoyingly it seems that the representation I've done is actually called "Miller Extended Harmonic" which is supposed to be better than Miller, but is definitely not Fourier/Miller. It is still used in GKDB/CGYRO so is necessary (thankfully not a complete waste of time). Will need to give it a new name. (also has the same flag as Miller so we need to add in a special check for distinguish between Miller and MXH).

Hopefully the fourier representation will be quick to implement.

This is what CGYRO actually uses:
image

This is what GENE uses
image

This basically looks the same as CGYRO

This is what GS2 uses which does look different to the other two.
image
image

Either we need to figure out the conversion between codes (no idea how trivial that is) or re-fit for GS2.  @ZedThree thoughts?

ZedThree:
One option (not that I'm necessarily suggesting it) is to implement the CGYRO/GENE Fourier equilibrium in GS2 -- I don't imagine many of us have time to do that.

Another option is that we just raise an exception if trying to write a Fourier eqm. to GS2.

Maths is not my strong suit, but I _think_ the GS2 version can be turned into the general one, but not vice-versa. It only has parameters for two modes, _m_ and _n_, but if you expand the whole thing out, you get terms for the following modes: 0, 1, _m - 1_, _m + 1_, _n - 1_, _n + 1_

So you can convert all GS2 Fourier equilibria to the general version, but only a subset (with quite strict conditions) the other way.

bpatel2107:
So we now have 3 different local geometry fits that work. The Miller extended haromic above (used for GKDB and CGYRO)

CGYRO (the poloidal field fit is quite slow, may need better initial guesses)
image
image


GENE
image
image

MXH
image
image


So now we are at the point where we need to map between the different fits. In theory all we need is the `R`, `Z` and `b_poloidal` to get the shaping coeffficients with the rest of the terms like `q` already being stored. We also need to be careful about `theta` as that can be different in different fits.

Essentially we can add a `load_from_local_eq` method to each class to do this which is straightforward.

The question now is should we store the `R`, `Z` and `b_poloidal` from the fit or the the actual data. It only is relevant when loading data from an `GlobalEquilibrium`

bpatel2107:
So I've re-vamped the LocalGeometry classes to all use the super `load_from_eq` method as part of that is common throughout the different methods. The only major differences is the shaping parameters which can be derived entirely from `R`, `Z` and `b_poloidal` so I've created a `get_shaping_coefficients` method for each subclass that calculates the relevant parameters. 

The neat thing is that now transferring between different geometries is trivial as we just need to re-calculate the shaping coefficients which can be easily done now.

When loading in from GK files it is mostly trivial though we'll need to either skip over unknown parameters like `psi_n` or calculate things like `f_psi` through loop integrals

pyrokinetics/local_geometry/LocalGeometry.py
```diff
@@ -26,14 +50,159 @@ def __init__(self, *args, **kwargs):
...
+        self.Rmaj = float(R_major / eq.a_minor)
+        self.Z0 = float(Zmid / eq.a_minor)
+        self.a_minor = float(eq.a_minor)
+        self.f_psi = float(fpsi)
+        self.B0 = float(B0)
+        self.q = float(q)
+        self.shat = shat
+        self.beta_prime = beta_prime
+        self.pressure = pressure
+        self.dpressure_drho = dpressure_drho
```
ZedThree, line 168: Could we combine these lines with the above and just assign directly?

pyrokinetics/local_geometry/LocalGeometry.py
```diff
@@ -26,14 +50,159 @@ def __init__(self, *args, **kwargs):
...
+        # Calculate shaping coefficients
+        self.get_shape_coefficients(self.R, self.Z, self.b_poloidal)
+
+        # Bunit for GACODE codes
+        self.bunit_over_b0 = self.get_bunit_over_b0()
+
+        if show_fit:
+            self.plot_fits()
+
+    def load_from_lg(self, lg, verbose=False, show_fit=False):
```
ZedThree: Is this method on the right class? It looks like a copy method rather than a constructor, but says it creates a `FourierCGYRO`

Also, let's expand those abbreviations :stuck_out_tongue: 

```suggestion
    def load_from_local_geometry(self, local_geometry, verbose=False, show_fit=False):
```

pyrokinetics/pyro.py
```diff
@@ -1093,6 +1093,27 @@ def local_geometry_type(self) -> Union[str, None]:
...
+        Returns
+        -------
+
+        """
+
+        # Check if already loaded and if show then switch geometries
+        if not isinstance(self.local_geometry, LocalGeometry):
+            raise ValueError("Please load local geometry before switching")
+
+        if local_geometry in self.supported_local_geometries:
```
ZedThree: If it's not supported, we should raise a `ValueError` instead, e.g.:

```suggestion
        if local_geometry not in self.supported_local_geometries:
            raise ValueError(f"Unsupported local geometry type. Got '{local_geometry}', expected one of: {self.supported_local_geometries.keys()}")
```

Then rest of method can be unindented one step too :)

ZedThree (commented):
Looks good -- needs some tests though! A simple one would be converting from Miller to Fourier, back to Miller and checking it's (roughly) the identity.

You've also already been testing by checking the fit -- you could convert that to a test by calculating e.g. the l2-norm of the difference and checking that's below some threshold.

bpatel2107:
I have been writing the tests and came across a small hiccup which I am trying to work through! May have some time next week to work on this

pyrokinetics/local_geometry/LocalGeometry.py
```diff
@@ -26,14 +50,159 @@ def __init__(self, *args, **kwargs):
             super(LocalGeometry, self).__init__(_data_dict)
 
     # TODO replace this with an abstract classmethod
-    @not_implemented
-    def load_from_eq(self, eq, psi_n=None):
+    def load_from_eq(self, eq, psi_n=None, show_fit=False, **kwargs):
```
LiamPattinson: `psi_n=None` by default, but this will lead to an error when its passed to `eq.get_flux_surface`. That function has `psi_n=1.0` by default, so for consistency it would be good to keep it the same here (although personally, I'd argue in favour of not having a default value, so the user has to specify `psi_n` explicitly).

pyrokinetics/local_geometry/LocalGeometry.py
```diff
@@ -26,14 +50,159 @@ def __init__(self, *args, **kwargs):
...
+        self.a_minor = float(eq.a_minor)
+        self.f_psi = float(fpsi)
+        self.B0 = float(B0)
+        self.q = float(q)
+        self.shat = shat
+        self.beta_prime = beta_prime
+        self.pressure = pressure
+        self.dpressure_drho = dpressure_drho
+
+        self.R = R
```
LiamPattinson: I'm not sure about storing the `Equilibrium`-derived `R`, `Z`, `b_poloidal`, etc and `psi_n` in the `LocalGeometry` instance itself. Since these attributes aren't set when a `LocalGeometry` is instantiated from a GK input file, it means the class will behave very differently depending on the context of how it was created, which makes it harder to test/document the class and could lead to unexpected behaviour for the users. It's also a little non-intuitive that `self.R` and `self.Z` are the RZ-coordinates of the flux surface we're using as a reference, rather that the RZ-coordinates of the fitted flux surface described by the `LocalGeometry` (similarly for `b_poloidal`, `f_psi`, etc). Perhaps the `flux_surface` function associated with each derived class could be used to set RZ when objects are initialised?

It would take a pretty extensive rewrite to this PR to change this, so it might be better to come back to this in a later refactor.

pyrokinetics/local_geometry/LocalGeometryMiller.py
```diff
@@ -5,30 +5,27 @@
...
-        "btccw": -1,
-        "ipccw": -1,
-        "beta_prime": 0.0,
         "local_geometry": "Miller",
     }
 
+    return {**base_defaults, **miller_defaults}
+
 
 def grad_r(
```
LiamPattinson: External functions `grad_r`, `flux_surface` and `b_poloidal` are defined for each class, but are only used internally. Would it make more sense for these functions to be part of the classes? I think they only need a `theta` grid, and otherwise these functions make use of attributes belonging to the classes.

There's a discrepancy in how `flux_surface` is used, with some functions calls using `(self.Rmaj, self.rho, self.Z0)` and others using `(self.Rmaj*self.a_minor, self.r_minor, self.Rmaj*self.a_minor)`, so perhaps that would need to take an extra argument of `normalised=False` to switch between each version.

pyrokinetics/local_geometry/LocalGeometryMiller.py
```diff
@@ -468,6 +423,47 @@ def get_bunit_over_b0(self):
 
         return integral * self.Rmaj / (2 * pi * self.rho)
 
+    def plot_fits(self):
```
LiamPattinson: The `plot_fits` functions are all very similar to each other, and it should be possible to write a single generic function for all cases. I had a play around, and found that this works for Miller, and it should be easily extendable to the other methods too:

```python
# Define these methods at the end of LocalGeometryMiller
# Ideally, the external grad_r/flux_surface/b_poloidal functions should be 
# defined here instead of being free functions.

    def grad_r(self, theta):
        return grad_r(
            self.kappa, self.delta, self.s_kappa, self.s_delta, self.shift, theta
        )

    def flux_surface(self, theta, normalised=False):
        Rmaj = self.Rmaj if normalised else self.Rmaj * self.a_minor
        rmin = self.rho if normalised else self.r_minor
        Z0 = self.Z0 if normalised else self.Z0 * self.a_minor
        return flux_surface(self.kappa, self.delta, Rmaj, rmin, theta, Z0)

    def get_b_poloidal(self, R, theta):
        """
        TODO: can just call this b_poloidal if no longer storing b_polidal from eq
        Instead of passing in R, should be able to derive this from the flux surface and theta
        """
        return b_poloidal(
            kappa=self.kappa,
            delta=self.delta,
            s_kappa=self.s_kappa,
            s_delta=self.s_delta,
            shift=self.shift,
            dpsi_dr=self.dpsidr,
            R=R,
            theta=theta,
        )

# Define the following at the bottom of LocalGeometry.py
def plot_equilibrium_to_local_geometry(
    eq: Equilibrium,                                                                                                                                                                                                                                                                                                           
    local_geometry_type: str,                                                                                                                                                                                                                                                                                                  
    psi_n: float,                                                                                                                                                                                                                                                                                                              
    axes: Optional[Tuple[plt.Axes, plt.Axes]] = None,                                                                                                                                                                                                                                                                          
):                                                                                                                                                                                                                                                                                                                             
    # Get flux surface at psi_n                                                                                                                                                                                                                                                                                                
    R, Z = eq.get_flux_surface(psi_n=psi_n)                                                                                                                                                                                                                                                                                    
    b_poloidal = eq.get_b_poloidal(R, Z)                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                               
    # Get theta grid from eq
    # NOTE: Copied from LocalGeometryMiller.
    #       Do other LocalGeometry classes need to calculate theta differently?                                                                                                                                                                                                                                                                                               
    r_minor = eq.rho(psi_n) * eq.a_minor                                                                                                                                                                                                                                                                                       
    kappa = (max(Z) - min(Z)) / (2 * r_minor)
    Zmid = (max(Z) + min(Z)) / 2
    Zind = np.argmax(abs(Z))
    R_upper = R[Zind]
    normalised_height = (Z - Zmid) / (kappa * r_minor)
    # Floating point error can lead to >|1.0|
    normalised_height = np.where(
        np.isclose(normalised_height, 1.0), 1.0, normalised_height
    )
    normalised_height = np.where(
        np.isclose(normalised_height, -1.0), -1.0, normalised_height
    )
    theta = np.arcsin(normalised_height)
    for i in range(len(theta)):
        if R[i] < R_upper:
            if Z[i] >= 0:
                theta[i] = np.pi - theta[i]
            elif Z[i] < 0:
                theta[i] = -np.pi - theta[i]

    # Fit LocalGeometry at same psi_n
    local_geometry = local_geometries(local_geometry_type)
    local_geometry.load_from_eq(eq, psi_n=psi_n)

    # Get flux surface and b_poloidal from local_geometry
    R_fit, Z_fit = local_geometry.flux_surface(theta)
    # TODO: with some refactoring, shouldn't need to pass in R_fit
    b_poloidal_fit = local_geometry.get_b_poloidal(R_fit, theta) 

    # Set up plot if one doesn't exist already
    if axes is None:
        fig, axes = plt.subplots(1, 2)
    else:
        fig = axes[0].get_figure()
    
    # Create plots
    ax1, ax2 = axes
    
    ax1.plot(R, Z, label="Equilibrium")
    ax1.plot(R_fit, Z_fit, "--", label=f"LocalGeometry{local_geometry_type}")
    ax1.set_title(f"Fit to flux surface for {local_geometry_type}")
    ax1.legend()
    ax1.grid()

    ax2.plot(theta, b_poloidal, label="Equilibrium")
    ax2.plot(theta, b_poloidal_fit, "--", label=f"LocalGeometry{local_geometry_type}")
    ax2.set_title(f"Fit to poloidal field for {local_geometry_type}")
    ax2.legend()
    ax2.grid()

    # return fig, axes instead of plt.show/savefig, let the user decide what to do with them
    return fig, axes

# Call as so:

from pyrokinetics.equilibrium import Equilibrium
from pyrokinetics.local_geometry.LocalGeometry import plot_equilibrium_to_local_geometry

eq = Equilibrium("pyrokinetics/templates/test.geqdsk")
fig, axes = plot_equilibrium_to_local_geometry(eq, "Miller", psi_n=0.5)
plt.show()
```
Having an external plotting function also keeps the API cleaner, as we don't need to propagate `show_fit` all the way down the stack. It also means we can return `fig, axes` and let the user handle them instead of forcing them into `plt.show()`. However, I'm not as familiar with any of the Fourier methods, so haven't got this to work for those yet. Again, maybe something to save for a later PR.

pyrokinetics/local_geometry/LocalGeometryFourier.py
```diff
@@ -0,0 +1,593 @@
...
+    shift: Scalar,
+    Rmaj: Scalar,
+    Z0: Scalar,
+) -> np.ndarray:
+    """
+    fourier definition of grad r from
+    fourier, R. L., et al. "Noncircular, finite aspect ratio, local equilibrium model."
+    Physics of Plasmas 5.4 (1998): 973-978.
+
+    Parameters
```
LiamPattinson: The parameter list in the docstring doesn't match the argument list. This also applies to other docstrings in the file.

pyrokinetics/local_geometry/LocalGeometryFourier.py
```diff
@@ -0,0 +1,593 @@
...
+
+    Returns
+    -------
+    grad_r : Array
+        grad_r(theta)
+    """
+
+    n_moments = len(cN)
+    n = np.linspace(0, n_moments - 1, n_moments)
+    ntheta = n[:, None] * theta[None, :]
```
LiamPattinson: Writing `[:, np.newaxis]` instead of `[:, None]` makes the intent clearer (`np.outer(n, theta)` would be even clearer in this particular instance).

You can avoid a lot of extra `[:,None]` later by using the transpose of `ntheta` and summing over axis 1:
```python
ntheta = np.outer(theta, n)
aN = np.sum(cN * np.cos(ntheta) + sN * np.sin(ntheta), axis=1)
```

pyrokinetics/local_geometry/LocalGeometryFourier.py
```diff
@@ -0,0 +1,593 @@
...
+
+    cN[0] *= 0.5
```
LiamPattinson: Better to avoid modifying the input arrays like this, even if only temporarily. If something went wrong between now and the `*= 2.0` lines, such as a numpy exception being raised, this could leave a `LocalGeometryFourier` in an unstable state. I'd recommend weighting the sum:

```python
weights = np.ones(n.shape)
weights[0] = 0.5
aN = np.sum(weights*(cN * np.cos(ntheta) + sN * np.sin(ntheta)), axis=1)
# or maybe...
aN = np.dot(weights, (cN * np.cos(ntheta) + sN * np.sin(ntheta)).T)
```

pyrokinetics/local_geometry/LocalGeometryFourier.py
```diff
@@ -0,0 +1,593 @@
...
+        n = np.linspace(0, self.n_moments - 1, self.n_moments)
+        ntheta = n[:, None] * theta[None, :]
+        cN = simpson(aN[None, :] * np.cos(ntheta), theta, axis=1) / np.pi
+        sN = simpson(aN[None, :] * np.sin(ntheta), theta, axis=1) / np.pi
+
+        self.cN = cN
+        self.sN = sN
+
+        self.theta = theta
+        self.R = R
```
LiamPattinson: The `self.R`, `self.Z` etc in `LocalGeometryMiller` are taken directly from the `Equilibrium` object, but here they've been modified (just interpolated onto an evenly spacing theta grid, as far as I can tell?). Using `self.R` and `self.Z` to represent the 'training data' in `plot_fits` could be a little misleading.

pyrokinetics/local_geometry/LocalGeometryFourier.py
```diff
@@ -0,0 +1,593 @@
...
+            dsNdr=dsNdr,
+            Rmaj=self.Rmaj,
+            Z0=self.Z0,
+            R=self.R,
+            shift=shift,
+            dZ0dr=dZ0dr,
+            dpsidr=dpsidr,
+        )
+
+    def test_safety_factor(self):
```
LiamPattinson: This function is copy pasted across all classes, and isn't used anywhere. Could it be moved to the base class, or removed? It looks like the line `b_poloidal = self.get_b_poloidal` will cause it to fail anyway.

pyrokinetics/local_geometry/LocalGeometryMiller.py
```diff
@@ -268,44 +265,24 @@ def load_from_eq(self, eq: Equilibrium, psi_n: float, verbose=False):
...
-        R, Z = eq.get_flux_surface(psi_n=psi_n)
-
-        b_poloidal = eq.get_b_poloidal(R, Z)
-
-        R_major = eq.R_major(psi_n)
+        drho_dpsi = eq.rho.derivative()(psi_n)
+        shift = eq.R_major.derivative()(psi_n) / drho_dpsi / eq.a_minor
 
-        rho = eq.rho(psi_n)
+        super().load_from_eq(eq=eq, psi_n=psi_n, verbose=verbose, shift=shift, **kwargs)
```
LiamPattinson: Here we pass `shift` to the super function, but as far as I can tell it isn't used for anything. Since these functions take `**kwargs`, it doesn't raise an 'unexpected keyword' error, and instead just silently ignores it. Might it be worth removing `**kwargs` from the functions that aren't currently using it? It would make it easier to spot problems like this, or issues such as typos when writing keyword arguments.

pyrokinetics/local_geometry/LocalGeometryFourier.py
```diff
@@ -0,0 +1,593 @@
...
+            )
+
+        if verbose:
+            print(f"Fourier :: Fit to Bpoloidal obtained with residual {fits.cost}")
+
+        if fits.cost > 0.1:
+            import warnings
+
+            warnings.warn(
+                f"Warning Fit to Fourier in Miller::load_from_eq is poor with residual of {fits.cost}"
```
LiamPattinson: Warning string says 'Miller' instead of 'Fourier'

pyrokinetics/local_geometry/LocalGeometryFourierCGYRO.py
```diff
@@ -0,0 +1,633 @@
...
+        theta = theta_new
+
+        self.R = R
+        self.Z = Z
+        self.b_poloidal = b_poloidal
+        self.theta = theta
+
+        n = np.linspace(0, self.n_moments - 1, self.n_moments)
+        ntheta = n[:, None] * self.theta[None, :]
+        aR = (
```
LiamPattinson, line 238: Black appears to have had fun with these lines. I'm not sure why, as it's within 88 chars. Maybe it's interpreting `[None, :]` as a list instead of a slice? Just a minor readability issue, nothing serious.

LiamPattinson (commented):
I'm currently doing some work for @rahulgaur104 which I'm hoping will make it easier to get VMEC equilibria incorporated into the project, and I was thinking of taking a look over `Equilibrium` and related classes while I'm at it. I think it would be good to get this PR merged first to make sure there aren't any conflicts. I've had a look over everything, and it all looks decent to me. I've left some comments for ways to improve some sections, but there's nothing serious, and it could all wait until a later refactor.

If you're still having trouble with the tests, you could always push a work-in-progress and I'd be happy to help if I'm able to.

pyrokinetics/local_geometry/LocalGeometry.py
bpatel2107: Agreed, forcing them to specify `psi_n` explicitly is best

pyrokinetics/local_geometry/LocalGeometry.py
bpatel2107: So I have just changed this such that all the different versions of `LocalGeometry` have `R`, `Z`, and `b_poloidal`.  This is because this should completely specify an local equilibrium and thus is used when switching between them. I think the best practice is to store these the first time a `LocalGeometry` object is created and then ensure they can't be changed so regardless of which geometry type we used we always fit to the same `R`, `Z` and `b_poloidal`. 

We could also store something like `R_fit`, `Z_fit` and `b_poloidal_fit` which are the values calculated from the `Equilibrium` object. But when we run a GK code, it calculates the `R`  `Z` and `b_poloidal` from the inputs so we should make sure pyro does the same. If its a poor fit then that's a separate issue. This means `R`, `Z`, and `b_poloidal` will always be set from the fitted value so it won't matter where `pyro` gets those values.

You can also calculate `f_psi` by doing an integral so that is probably another thing we should add in

pyrokinetics/local_geometry/LocalGeometryMiller.py
bpatel2107: They should probably be an abstract function as all of them will need these, but the actual function will be different due to the different forms of fits.

For example `Miller` uses `kappa`, `delta` etc. when calculating the `flux_surface`, but `Fourier` uses the fourier coefficients `cN` and `sN`

I agree about the normalised option

pyrokinetics/local_geometry/LocalGeometryMiller.py
bpatel2107: Happy to separate this out! May be slightly difficult given the different geometry types though

pyrokinetics/local_geometry/LocalGeometryFourier.py
bpatel2107: I was just lazy here sorry

pyrokinetics/local_geometry/LocalGeometryFourier.py
bpatel2107: Agreed

pyrokinetics/local_geometry/LocalGeometryFourier.py
bpatel2107: I think I just added in the 0.5 upstream and removed this part

pyrokinetics/local_geometry/LocalGeometryFourier.py
bpatel2107: Agreed, see earlier comment

bpatel2107:
Sorry about the delay on this, haven't had much time to work on this (also on leave for the next 2 weeks which doesn't help!) I've pushed my latest version of this. There are a lot of Copy-Paste jobs so doc-strings may not be updated.

I've left my comments so far and in general agree with many of your points. I'm happy for you to work on this going ahead. The test are very tricky and there are some major issues when calculating the shaping derivatives. Essentially I may need another term on top of `R` `Z` and `b_poloidal` as I can't fully constrain my fits. Essentially I can work out the ratio of `dpsidr` and  `dcNdr` in Fourier but not the actual values so I think I'm not fully constraining the system.

We are planning an even more extensive geometry calculation with the help of Harry Dudding but that may be after we do this 

I'm back on the 8th Nov. Might be worth meeting after that to discuss this as I'm curious about Rahul's work!

bpatel2107:
@LiamPattinson anything been done on this?

LiamPattinson:
I'm afraid I haven't had a chance yet, since I've been working on Equilibrium stuff in preparation for getting VMEC equilibria in (PR https://github.com/pyro-kinetics/pyrokinetics/pull/128). It's probably a good idea for me to leave that work where it is for now, and get to solving these problems.

To solve the fitting problem, do you think it might help to introduce dRdr and dZdr into the calculations? I think it should be possible to get those using things like `s_kappa`/`s_delta` in Miller, though I'm not familiar with the other systems yet. Or do you think we'll need some derivatives of the B field?

bpatel2107:
Yes I think we need to store dRdr and dZdr, but may also need other B field derivatives. 

I am meeting with Harry Dudding today to discuss the data that we need as he has done detailed derivations of all these terms. I am expecting a pretty significant overhaul of the LocalGeometry class which will hopefully be much more detailed and contain many metric coefficients

bpatel2107:
@LiamPattinson @ZedThree 

This is finally ready to be reviewed! 
Re-structured much of the LocalGeometry using the base class method where possible. 
Fixed all the physics issues I was facing so we can now convert between codes easily. 
Docstrings have be updated
Added several tests with some golden answer checks.
Tidied up a lot of code

bpatel2107:
I had to add in the squareness to the code to get all the tests to work along with updating the list of attrs to ignore when comparing LocalGeometries.

The biggest point of concern is that the squareness does change the Miller fit so if someone runs pyro now, the results will be different (but better!). 

GS2 doesn't have squareness in yet so I have forced it to be 0 and raised a warning (@d7919 lmk if this is fine for now). This means cross-code comparisons with GS2 are somewhat limited until this is added in.

bpatel2107:
@hgd511 FYI

pyrokinetics/gk_code/GKInputCGYRO.py
```diff
@@ -310,22 +363,34 @@ def set(
             self.read(template_file)
 
         # Geometry data
-        if not isinstance(local_geometry, LocalGeometryMiller):
+        if isinstance(local_geometry, LocalGeometryMiller):
+            eq_model = 2
+        elif isinstance(local_geometry, LocalGeometryFourierGENE):
```
LiamPattinson: Is it correct to check for GENE Fourier in CGYRO?

pyrokinetics/local_geometry/LocalGeometry.py
```diff
@@ -26,14 +109,425 @@ def __init__(self, *args, **kwargs):
             super(LocalGeometry, self).__init__(_data_dict)
 
     # TODO replace this with an abstract classmethod
-    @not_implemented
-    def load_from_eq(self, eq, psi_n=None):
+    def from_global_eq(
```
LiamPattinson, line 112: Not something that needs to be done for this PR, but ideally these `from_...` methods should be implemented as alternative constructors using `@classmethod`, as in [this tutorial](https://realpython.com/python-multiple-constructors/#providing-multiple-constructors-with-classmethod-in-python). It could also be argued that they should be moved out of the `LocalGeometry` classes altogether:

```python
eq = Equilibrium(...)
# Currently, build empty and then populate:
local_geometry = LocalGeometryMiller().from_global_eq(eq, psi_n=0.5)
# classmethod used as alternative to __init__:
local_geometry = LocalGeometryMiller.from_global_eq(eq, psi_n=0.5)
# Maybe this should instead be the responsibility of Equilibrium?
local_geometry = eq.local_geometry(psi_n=0.5, fit="Miller")
```

pyrokinetics/local_geometry/LocalGeometry.py
```diff
@@ -26,14 +109,425 @@ def __init__(self, *args, **kwargs):
...
+        self.pressure = pressure
+        self.dpressure_drho = dpressure_drho
+        self.dpsidr = dpsidr
+
+        self.R_eq = R
+        self.Z_eq = Z
+        self.b_poloidal_eq = b_poloidal
+
+        # Calculate shaping coefficients
+        self.get_shape_coefficients(self.R_eq, self.Z_eq, self.b_poloidal_eq, **kwargs)
```
LiamPattinson: Could this be `_set_shape_coefficients` instead? I'd expect a `get_` function to return something instead of changing internal state.

pyrokinetics/local_geometry/LocalGeometry.py
```diff
@@ -2,14 +2,97 @@
 from copy import deepcopy
 from ..decorators import not_implemented
 from ..factory import Factory
+from ..constants import pi
+import numpy as np
+from typing import Tuple, Dict, Any, Optional
+from ..typing import ArrayLike
+from ..equilibrium import Equilibrium
+from matplotlib.pyplot import Axes, show
```
LiamPattinson: We import matplotlib twice in this file, first here, and then later in `plot_equilibrium...`. Could we just do the usual `import matplotlib.pyplot as plt` here and just stick `plt.` in front of `Axes` and `show`?

pyrokinetics/local_geometry/LocalGeometryFourierCGYRO.py
```diff
@@ -0,0 +1,503 @@
...
+        self.daZdr = fits.x[self.n_moments : 2 * self.n_moments]
+        self.dbRdr = fits.x[2 * self.n_moments : 3 * self.n_moments]
+        self.dbZdr = fits.x[3 * self.n_moments :]
+
+    @property
+    def n(self):
+        pass
+
+    @n.getter
+    def n(self):
```
LiamPattinson: Is there a reason to define the property with just `pass` and then set the `getter` explicitly like this?

LiamPattinson (commented):
Very nice, looking much tidier. Looks really well documented and tested.

I'm also a little concerned about adding a new parameter to Miller, partly because it's not in the original paper, and partly because it might lead to misleading results for GS2 users if they're able to get a really good fit in Pyrokinetics but their target code isn't able to use it properly.

On the other hand, squareness _is_ recognised by CGYRO, so we can't simply ignore it. It'd be a fairly messy refactor to have `LocalGeometryMiller` behave differently depending on whether the user wants it to contain `zeta` or not, while splitting it into `LocalGeometryMillerLikeInThePaper` and `LocalGeometryMillerButWithSquareness` would lead to the code getting a bit bloated.

I'm honestly not sure how to solve this problem right now, so I think warning the users is the best option at this stage.

pyrokinetics/gk_code/GKInputCGYRO.py
bpatel2107: That's a mistake, good spot!

pyrokinetics/local_geometry/LocalGeometryFourierCGYRO.py
bpatel2107: Nope, changed that now to just calculate it directly

pyrokinetics/local_geometry/LocalGeometry.py
bpatel2107: I agree that these probably should be a  `classmethod` though I'm not 100% sure about how to implement it. I tried but did not succeed :( 

Where would you move them to?

bpatel2107:
I think the best way to move forward is to add squareness into GS2. This should be possible and in theory is not difficult so hopefully this warning won't need to remain in for long. Though this will require the latest version of GS2 to work which isn't necessarily a bad thing.

Squareness is something that should be included as CGYRO, GENE and TGLF all use it so it is quite a common input. We do cite the Turnbull paper where it was officially added but in general people consider this a "Miller" geometry.

ZedThree:
Similarly, it might be nice to add the full Fourier equilibrium to GS2 as well.

I think we're going to want to be able to handle different versions of the same code at this point though -- that will require some thinking about how best to do it.

pyrokinetics/local_geometry/LocalGeometry.py
LiamPattinson: I wouldn't worry about it for this PR, since it would be most easily achieved in a more serious refactor and I think what you have here is good. In a future update, I'm imagining something like this:

```python
class LocalGeometryMiller(LocalGeometry):
    # Have __init__ take just Miller parameters and a few optional args
    def __init__(self, rho, Rmaj, zmid, shat, q, kappa, delta, zeta, s_kappa, s_delta, s_zeta, ntheta=64):
        self.rho = rho
        self.Rmaj = Rmaj
        ...
        self.s_zeta = s_zeta
        # Calculate self.R, self.Z, self.b_poloidal, self.bunit_over_b0, etc using Miller params
        # (This will use a bunch of overridden functions such as get_flux_surface, get_b_poloidal, etc)
        super().__init__(ntheta)


    # Provide alternative constructors
    @classmethod
    def from_global_eq(cls, eq, psi_n, ntheta=64):
        rho = eq.rho(psi_n)
        ...
        s_zeta = calculate_s_zeta_somehow()
        return cls(rho, Rmaj, zmid, shat, q, kappa, delta, zeta, s_kappa, s_delta, s_zeta, ntheta=ntheta)
```

We'd have to write a specialised `__init__` for each type of `LocalGeometry`, and then each `from_global_eq` would have to determine every parameter using `Equilibrium` data only and without touching any attributes of `LocalGeometry` itself.

The point about maybe moving this responsibility to `Equilibrium` is more of a philosophical issue than a practical one. Currently, each local geometry must have the ability to parse and understand the entire global equilibrium. I feel like it would be 'cleaner' for an `Equilibrium` to know how to produce subsets of itself than for the subsets to produce themselves from the whole, if that makes any sense. I made some steps towards it in PR https://github.com/pyro-kinetics/pyrokinetics/pull/128, where `Equilibrium` can produce a generic `FluxSurface` object, with the intention of `LocalGeometry` objects building themselves from that instead. From there, you could have something like:

```python
class Equilibrium:
    ...
    def local_geometry(self, psi_n, fit):
        # This would probably be defined elsewhere:
        fittings = {"Miller": LocalGeometryMiller, "FourierCGYRO": LocalGeometryFourierCGYRO, ...}
        fs = self.flux_surface(psi_n)
        return fittings[fit].from_flux_surface(fs)
```

The only real difference is that `LocalGeometry` no longer has any idea that `psi_n` or any wider equilibrium exists, regardless of whether it's created from an `Equilibrium`, a `GKInput`, or just user inputs. Practically speaking, it doesn't matter much either way.

bpatel2107: Hmm I see what you mean. 

A generic `FluxSurface` Object is essentially a `LocalGeometry` Object but without the shaping coefficients. Something could be done on this though we need to think about whether there is enough of a difference here. But philosophically I agree with you.

bpatel2107:
The version control requires some additional metadata which we don't handle but once that is there hopefully this will be easier. To check the version pyro would need to know where the binary file is which is another thing to consider as you want to know about zeta before the run is submitted.

bpatel2107:
Will merge now and leave the refactoring plus FourierGS2 for another PR

PR #113, opened 2022-08-18
Add method to convert `Pyro` object to IMAS dict

ZedThree:
Return a JSON-compatible structure that conforms to the GKDB/IMAS/OMAS gyrokinetics schema as described in:

- https://gitlab.com/gkdb/gkdb/raw/master/doc/general/IOGKDB.pdf
- https://gafusion.github.io/omas/schema/schema_gyrokinetics.html

Currently only a partial implementation needed for the TDoTP database.

Requires species and geometry data to already exist.

@bpatel2107 I know you had made a start on this already, but it turns out there's some non-trivial mappings, such as the `wavevector` structure. This is because GKDB just has a simple list of all the `(kx, ky)` pairs. That means it's flexible enough to handle any random assortment of wavenumbers (the cost is that it's not terrible efficient storage for dense arrays).

This is WIP -- there's lots more of the GKDB schema to add, but more importantly the normalisations need checking. So this is perhaps somewhat dependent on #97


bpatel2107:
We can probably hard code the conversion from pyro normalisation to GKDB for now to get it working and then update this properly #97 as that has some nuanced behaviours that we need to careful with

ZedThree:
This currently includes #115 so that I can use it in the TDotDat demo server

fcasson:
I still think this is valuable.  Maybe @hgd511 can help you check the normalisations.

bpatel2107:
Now that we have set up the geometry needed for IMAS (`MXH`) and have normalisations I think we can start moving forward with this again.

@ZedThree do you want to crack on with this now or wait until the outputs are normalised too?

ZedThree:
When I find some time, I'll see how far I can push this without the normalised outputs

bpatel2107:
I've added the base structure of what I think the mapping from the pyro **inputs** to the gyrokinetics DD are. I will need to check the outputs but I think we have the basic dictionary down.

We can probably use this along with IDSpy_toolkit to actually create the IDS.

Anything pyro doensn't handle yet I kept as a `None`. We can look into adding these things

bpatel2107:
We can now read/write to IMAS IDS and we get a good agreement between the linear eigenfunctions and values when comparing the original gk data and data from the IDS, example from GS2 below

image


We have the option to write the IDS to HDF5 and as `idspy_toolkit` develops other methods, we will have those options too.

Major points to consider

- Do we want the options to downsample in time when writing IDS? Also options to choose what we actually store?
- Need to determine which frame each code (lab or rotating) and (particle or guiding centre)
- IDS doesn't store fluxes as a function of time or ky. Currently we take a mean but I think adding this functionality into the IMAS IDS should be done. Would be as simple as replacing the time averaged to a 1d array in time.
- Pyro currently doesn't handle case multiple linear (ky, kx) scans are stored in a single IDS. We could likely use a PyroScan object to handle this
- We don't read in the growth rate/mode frequency from the IDS, rather we calculate it from the fields. Probably should have an option to read that data.
- Due to the nested nature of the IDS this is absolutely full of for loops. Any suggestions on how to optimise this would be useful. Not really an issue for linear runs, but nonlinear can take a while
- Might be worth rescaling field using the same amplitude/phase that IMAS uses so we can do like for like comparisons more easily. 
- Need a way to set the normalisations from IDS

ZedThree:
Nice, thanks @bpatel2107 !

For storing the `(kx, ky)`, maybe we can use masked arrays? Might be a terrible idea, but it could make things a bit easier to handle.

I might have missed it, but is there a test for this? Just checking the round-trip like in your graphs is probably enough for now.

bpatel2107:
I'm not sure how the masked arrays would help with the `(kx, ky)` issue. This is a bit nuanced as each run can have a different input file so we need to be careful

No tests yet, will wait for @ycamenen to double check what I've done!

bpatel2107:
@LiamPattinson This is basically ready to go. The major issue is that `idspy-toolkit` doesn't support python 3.8. If this a version gate is the best solution and if we are dropping support for python 3.8 soon then there's not a lot of merit in changing the library.

Tests/Docs are failing due to this

LiamPattinson:
PR https://github.com/pyro-kinetics/pyrokinetics/pull/222 adds some version restrictions to keep things running in Python 3.8, although with a reduced feature set. I imagine we'll need to upgrade to 3.9 towards the end of this year anyway (we're already missing out on the latest NumPy versions).

bpatel2107 (approved):
I approve this which is kind of a given

bpatel2107:
Will merge for now before the docs refactor. Everything is working but IMAS DD may change their structure going ahead. No point waiting until theyre done though

PR #115, opened 2022-08-22
Minor fixes and tests for missing/default keys in GS2 inputs

ZedThree:
`jtwist` and `ngauss` might not be present, but will have default values.

Adds a bunch of tests for various variations of GS2 input files.

The test helper function `modified_gs2_input` enables deleting keys/namelists from an existing input file by setting them to `None` in the dict of replacements -- this could be useful for `GKInput.add_flags`? It might get a bit complicated with how it interacts with keys that will get set from other objects via `GKInput.set`. Maybe we should remove `add_flags` and instead add a `code_specific_parameters` input to `set`? Then it wouldn't be possible to modify things in a potentially inconsistent way

### DIFF
--- a/pyrokinetics/gk_code/GKInputGS2.py
+++ b/pyrokinetics/gk_code/GKInputGS2.py
@@ -304,12 +304,14 @@ class GKInputGS2(GKInput):
         }
 
         try:
-            return GRID_READERS[grid_option]()
+            reader = GRID_READERS[grid_option]
         except KeyError:
             valid_options = ", ".join(f"'{option}'" for option in GRID_READERS)
             raise ValueError(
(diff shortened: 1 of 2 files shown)


bpatel2107:
The main reason I get `add_flags` general was to allow for users to change any parameter they want, even if its not in pyrokinetics somewhere.. As long as this behaviour is maintained it should be fine.

Would this basically be an kwarg like

`pyro.write_gk_file(code_specific_parameters = {foo: bar})`

My only concern would be in the implementation in PyroScan objects but we can probably work around that.

ZedThree:
Yeah, something like that. I'll do it in a separate PR though, this one is just some easy bugfixes

PR #117, opened 2022-10-17
Fix GENE reader for missing `geometry::minor_r` input

ZedThree:
We need `minor_r` to be present in the `geometry` namelist, but it seems it's not required.

Please could a GENE expert check the default value is correct?


mgiacomin:
According to GENE documentation, `minor_r` should be required for circular and miller geometries. However, I found that only `major_R` and `trpeps` are required in circular geometry and `major_R` is usually set to 1. So `minor_r` is ignored in circular geometry. There is no default value for `minor_r` in GENE. In `pyrokinetics` the reference length is equal to the minor radius, so the default value `minor_r` should be 1, as you did.

PR #118, opened 2022-10-17
Drop support for Python 3.7

ZedThree:
Python 3.7 is end of life in June 2023, most OSes that still have it as the system python3 are either EOL or will be shortly. Several of our dependencies (xarray, numpy) have also started dropping support for Python 3.7.


bpatel2107:
Seems appropriate, gotta move with the times. I think we should aim to keep support for the latest 3 versions of python if possible, so have support for 3.8 - 3.10 from now

PR #119, opened 2022-10-20
Fix/cgyro sign frequency

mgiacomin:
The frequency sign needs to be flipped when read from the output file. Before, this was only done in the fields, if available.

### DIFF
--- a/pyrokinetics/gk_code/GKOutputReaderCGYRO.py
+++ b/pyrokinetics/gk_code/GKOutputReaderCGYRO.py
@@ -356,7 +356,7 @@ class GKOutputReaderCGYRO(GKOutputReader):
                 "'mode_frequency' and 'eigenvalue'."
             )
 
-        mode_frequency = eigenvalue_over_time[0, :, :]
+        mode_frequency = - eigenvalue_over_time[0, :, :]
         growth_rate = eigenvalue_over_time[1, :, :]
         eigenvalue = mode_frequency + 1j * growth_rate
         # Add kx axis for compatibility with GS2 eigenvalues
--- a/pyrokinetics/gk_code/GKOutputReaderCGYRO.py
+++ b/pyrokinetics/gk_code/GKOutputReaderCGYRO.py
@@ -356,7 +356,7 @@ class GKOutputReaderCGYRO(GKOutputReader):
                 "'mode_frequency' and 'eigenvalue'."
             )
 
-        mode_frequency = - eigenvalue_over_time[0, :, :]
+        mode_frequency = -eigenvalue_over_time[0, :, :]
         growth_rate = eigenvalue_over_time[1, :, :]
         eigenvalue = mode_frequency + 1j * growth_rate
         # Add kx axis for compatibility with GS2 eigenvalues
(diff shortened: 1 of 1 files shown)


bpatel2107:
I think we need to be careful here as this reminded me that CGYRO sets the frequency by looking at the first species defined in the input file. For example I have an input file where the first species is an electron and the sign is correct with the current implementation in pyro. What's the first species in your input file where the sign is wrong

mgiacomin:
Currently when the fields are read, the sign is always changed regardless of the first species, see here https://github.com/pyro-kinetics/pyrokinetics/blob/99307c169f4026a5b5b755b9f6ada1f926ae648c/pyrokinetics/gk_code/GKOutputReaderCGYRO.py#L248
So, shouldn't we also change here then?

bpatel2107:
Yes I agree, I just wanted to confirm that your sign issue was from the first species being an ion in your input file. Is that the case?

mgiacomin:
No, it is an electron species. Everything is correct if the fields are available. In this case, the function `_set_ eigenvalues` returns after calling `GKOutputReader._set_eigenvalues(data, raw_data, gk_input)`. However, if fields are not available (there are nans for instance), the function `_set_ eigenvalues` reads the mode frequency from `out.cgyro.freq` and the sign needs to be changed in order to be consistent with pyro convention. 

What happens if the first species is an ion species?

bpatel2107:
Oh I see the issue, my bad! What you have done is fine.

What I'm talking about is a slightly different issue then. CGYRO chooses the sign of the ion direction by looking at the first species. Whichever species is first, CGYRO will treat as the positive mode frequency. In pyro, we set the first species as an electron, meaning the output of CGYRO treat electron direction modes as positive, which is why we have the sign flip.

If the first species is an ion then it treats the ion direction as positive and then spits out the output accordingly.

What we really should be doing is instead having these minus signs is something like

```python
        mode_frequency = sign(cgyro_input['Z_1']) * eigenvalue_over_time[0, :, :]
```
 both where you suggest and where it was originally in the code

mgiacomin:
It should be okay now.

bpatel2107:
Can you tweak that to 

`np.sign(gk_input.data["Z_1"])`

in the very small chance that they use something with a charge not equal to 1 (like a helium plasma)? Or if they put the impurity first for some reason

bpatel2107:
Ah you beat me to it!

bpatel2107:
Not sure why multiplying by 1.0 broke the tests...

Compared to the golden answers, the outputs changed at the final decimal point so seems like a floating point error

mgiacomin:
Yes, I was trying to understand.

bpatel2107:
Multiplying by an integer rather than a float seemed to work!

mgiacomin:
Nice!

bpatel2107:
Oh wait no it didn't :(

mgiacomin:
It fails because `gk_input.data["Z_1"] = 1` and not -1.

ZedThree:
If this is fixing a bug, the failing tests are all golden answers -- so they might need to be updated.

bpatel2107:
Yeah I just saw this too. I'm just double checking the sign convention now to see if that should be negative or not

bpatel2107:
Ah okay so the behaviour is different to what I thought and a bit more complicated!

image
image

which is basically set by the sign of q and and btccw (which is negative by default.

bpatel2107:
@ZedThree python3.7 is failing now for a different reason? Is this due to us dropping support? Should we remove it from automated actions?

ZedThree:
Ah, that's just because this was branched off an earlier commit. The `(pull_request)` checks are all correct, so this ok to merge

PR #120, opened 2022-10-20
Fix reading `nonlinear` input for GENE

ZedThree:
Was missing namelist name

### DIFF
--- a/pyrokinetics/gk_code/GKInputGENE.py
+++ b/pyrokinetics/gk_code/GKInputGENE.py
@@ -270,7 +270,7 @@ class GKInputGENE(GKInput):
         numerics_data["nenergy"] = 0.5 * self.data["box"].get("nv0", 16)
         numerics_data["npitch"] = self.data["box"].get("nw0", 16)
 
-        numerics_data["nonlinear"] = bool(self.data["general"].get("nonlinear", 0))
+        numerics_data["nonlinear"] = self.data["general"].get("nonlinear", False)
 
         if numerics_data["nonlinear"]:
             numerics_data["nkx"] = self.data["box"]["nx0"]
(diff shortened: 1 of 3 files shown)


bpatel2107:
I added in the set up for `kx`, `ky` and the fluxes too

pyrokinetics/gk_code/GKOutputReaderGENE.py
```diff
@@ -309,12 +320,12 @@ def _set_fluxes(
         """
         # TODO This was changed to include a ky coordinate to match GS2 and CGYRO.
         #     Should this be reverted?
```
ZedThree: This comment can probably go now!

pyrokinetics/gk_code/GKInputGENE.py
```diff
@@ -78,7 +78,7 @@ def write(self, filename: PathLike, float_format: str = ""):
         super().write(filename, float_format=float_format)
 
     def is_nonlinear(self) -> bool:
-        return bool(self.data.get("nonlinear", 0))
+        return bool(self.data["general"].get("nonlinear", 0))
```
ZedThree:
```suggestion
        return self.data["general"].get("nonlinear", False)
```

bpatel2107: Should this been a boolean or a float?

ZedThree: Oh, I see you've changed the other bit to be like this -- does GENE set nonlinear to 1/0 rather than T/F?

bpatel2107: GENE sets it to a boolean so what you have suggested is right

ZedThree: It's supposed to return a bool. I only have one example GENE input file and it has `nonlinear = F`, but I forgot 1/0 are valid inputs for bools in Fortran. So I would have:

```suggestion
        return bool(self.data["general"].get("nonlinear", False))
```

bpatel2107: Oh god I keep changing it. This seems best

ZedThree:
The test is wrong :facepalm:

bpatel2107:
LGTM

PR #123, opened 2022-10-24
Improved get_flux_surface

LiamPattinson:
Instead of using matplotlib's contour function with a non-graphical backend, uses scikit-image's [`find_contours`](https://scikit-image.org/docs/stable/api/skimage.measure.html#skimage.measure.find_contours). This prevents pyrokinetics from interfering with any of the user's matplotlib code, and leads to a significant speed up. On my machine, unit tests have gone from taking around 30s to under 10s!

Almost all unit tests are continuing to pass with no changes needed. The only exception is `test_miller.py::test_load_from_eq`, which still passes if the relative error tolerance is set to 0.5%. Since this is a golden answer test, I'm hoping this isn't a major problem.

Resolves Issue https://github.com/pyro-kinetics/pyrokinetics/issues/60.

(Also found a way to stop some annoying command line IO whenever freegs reads a GEQDSK file.)

### DIFF
--- a/pyrokinetics/equilibrium/EquilibriumReaderGEQDSK.py
+++ b/pyrokinetics/equilibrium/EquilibriumReaderGEQDSK.py
@@ -1,3 +1,5 @@
+import os
+import sys
 from typing import Dict, Any
 from ..typing import PathLike
 from .EquilibriumReader import EquilibriumReader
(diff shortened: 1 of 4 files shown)


bpatel2107:
Very nice! Will merge now but I vaguely remember having some issues with scikit for particular edge cases but can't remember what they were. May need to keep an eye on this.

The print out comes from the dependency on FreeGS being used to read in the GEQDSK file (it's hard coded in that library). If we switch over to FreeQDSK then this would not be needed

Issue #125, opened 2022-10-31
Missing GENE output files


ZedThree:
There are a few GENE output files that we don't currently read:

- `nrg` if it's an HDF5 file
- `omega`, either text or HDF5 -- has growth rate/frequency, could be useful if `fields` isn't written or present?
- `all_parameters` -- HDF5 with the input namelist stored as text blob

Issue #124, opened 2022-10-31
Frequency sign convention and COCOS


ZedThree:
I noticed that GENE's `omega` is the opposite sign to what pyrokinetics calculates it as from the fields.

I'm not sure whether this comes from GENE choosing the electron diamagnetic drift direction as positive, rather than the ion direction, or if it has a different COCOS convention than we're expecting.

We need to make sure we deal with this correctly for every code (GK and equilibrium!). OMAS has a nice [document](https://docs.google.com/document/d/1-efimTbI55SjxL_yE_GKSmV4GEvdzai7mAj5UYLLUXw) on COCOS, and a [function](https://gafusion.github.io/omas/code/omas.cocos_transform.html#omas.cocos_transform) for converting one convention to another.

PR #126, opened 2022-10-31
Enable reading GENE `omega` text file

ZedThree:
Partial fix for #125 

The sample `omega_0001` output file was manually created from the existing sample output. Could a GENE expert check this works with other outputs too?


bpatel2107:
LGTM. Does this account for #124?

ZedThree:
Nope, that probably needs a GENE expert to check things

bpatel2107:
@dake0795 ?

dake0795:
I believe GENE default Is COCOS=2/12.

dake0795:
Default is that both the plasma current and the toroidal magnetic field are clockwise in a right-handed (ρ, θpol, φtor) coordinate system. Ion diamagnetic direction is positive in the omega_000X file.

ZedThree:
What COCOS do we use in pyrokinetics? ITER/IMAS is 11, right handed $(R, \phi, Z)$ and $(\rho, \theta, \phi)$

bpatel2107:
I _think_ the default pyrokinetics COCOS is 1/11

PR #127, opened 2022-11-06
Feature/gene improvements

mgiacomin:
This PR adds the following improvements to GENE output reader.

- Allow the user to down sample GENE data while reading, setting `downsize > 1`. This may be useful for large GENE run.
- Fix the kx grid. In fact, GENE data are saved starting with `kx=0`, like in GS2.
- Remove the ky dependence in GENE fluxes when it reads from the `nrg` file.
- Fix an issue with the last time entry in output files. Indeed, the last time step is not always written to the output files, but it depends on whatever condition is met first between `simtimelim` and `timelim`.

### DIFF
--- a/pyrokinetics/gk_code/GKOutputReaderGENE.py
+++ b/pyrokinetics/gk_code/GKOutputReaderGENE.py
@@ -79,7 +79,7 @@ class GKOutputReaderGENE(GKOutputReader):
         # search for 'parameters_####' in the run directory. If not, simply return
         # the directory.
         num_part_regex = re.compile(r"(\d{4})")
-        num_part_match = num_part_regex.search(str(filename))
+        num_part_match = num_part_regex.search(str(Path(filename).name))
         if num_part_match is None:
             return Path(filename).parent
         else:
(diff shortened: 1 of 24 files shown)


pyrokinetics/gk_code/GKOutputReaderGENE.py
```diff
@@ -80,7 +83,8 @@ def infer_path_from_input_file(filename: PathLike) -> Path:
         # the directory.
         filename = Path(filename)
         num_part_regex = re.compile(r"(\d{4})")
-        num_part_match = num_part_regex.search(filename.name)
+        num_part_match = num_part_regex.search(str(Path(filename).name))
```
ZedThree: Were you getting an error with some input here? `filename` should already be a `Path` from L84 above, and I thought that `Path.name` was already `str`

ZedThree (approved):
Might be nice to have a test case that shows off how `downsize` works

pyrokinetics/gk_code/GKOutputReaderGENE.py
mgiacomin: I can no longer reproduce the error. Maybe when I originally forked from `unstable`, L84 was not there and appeared after a later merge with `unstable`. I will remove this change. Thank you for noticing it!

bpatel2107:
Can we make this general to all codes? We load everything into an xarray dataset so it should be trivial to do this. The only difference is when we do this in the workflow. I'm not sure if this way saves memory or not but I think it still reads in the whole output file so it may not be much different?

We could also add a test in comparing the output from using downsample here compare to downsampling the output using xarray `.sel` method?

mgiacomin:
The entire output file is not read when `downsize>1`. This reduces the memory usage and makes data loading faster.

This can be extended to CGYRO and GS2. I started with GENE because I have quite large output files.

PR #128, opened 2022-11-09
Proposal for Equilibrium refactor

LiamPattinson:
I've sketched out a refactor for `Equilibrium` and related pieces that aims to provide a consistent interface for developers/users, hides some of the messier implementation details (factories, spline objects, etc), introduces well-defined units, and expands the documentation. This clashes with PR https://github.com/pyro-kinetics/pyrokinetics/pull/112, so I think the best way forward would be to get that sorted and then return to this later -- for now, I was just hoping to get some feedback on what I have so far.

## DatasetWrapper

I've introduced a generic class for creating thin wrappers around xarray Datasets. As per the [xarray docs](https://docs.xarray.dev/en/stable/internals/extending-xarray.html), it isn't recommended to subclass Dataset directly, but it would still be useful for us to have a Dataset-like class for managing grids/coordinates/netCDF IO, but with the additional benefit that we can define our own `__init__` functions to constrain what information they should contain. Xarray provides 'Accessors' (see link above) for extending DataArray/Dataset functionality, but (as far as I can tell) they would still require us to build raw Datasets, and the syntax for accessing our own features (`dataset.pyro.my_function()`, or something similar) is a bit clunky.

- Subclasses should define their own `__init__`, taking arbitrary arguments, and within should build coords/data_vars/attrs and pass them to `super().__init__()`.
- The underlying dataset can be accessed via `self.data`, and there are properties defined to provide direct access to `data_vars`, `coords`, and `attrs`.
- Indexing redirects to the underlying dataset, so `equilibrium["psi"]` will get the `psi` coord/data_var
- Additional metadata is added for free, such as date created, Pyrokinetics version, UUID for the current session, etc.
- Uses [pint-xarray](https://pint-xarray.readthedocs.io/en/stable/) to add units to each variable/coordinate.
- Classes which subclass DatasetWrapper can be written to/read from netCDF using `to_netcdf`/`from_netcdf`.

__Issues__

- NetCDF attributes can't have units. It may be possible to come up with a workaround, such as a global attribute `units` which is a string naming the units that each attribute should take (`"r_axis,meter:psi_lcfs,weber/radian"`, etc).
- To use pint-xarray, I had to set `force_ndarray` in the unit registry. This unfortunately means that any quantities with units are converted to a Numpy array, so floats become 0D arrays with dtype float. They usually behave quite well, so are unlikely to cause many problems, but they can become annoying for any code that directly checks for types. One instance of this is in f90nml, so I had to fix the version we're using and patch an internal function to keep things working. It may be possible for me to chase up this issue with the developers of f90nml and/or pint-xarray.

## Equilibrium

I've rewritten Equilibrium so that it can be constructed directly instead of `__init__` taking an input file. Creation via a file is now possible using a function `read_equilibrium` (similar to how Pandas delegates this to `read_csv`, Xarray uses `open_dataset`, etc). This makes it easier to document what the class contains and how it behaves, which in turn makes it easier to define new equilibrium file readers. It also allows us to test the class more easily by instantiating it directly. 

- Spline objects are created internally, and functions are defined to access them and their derivatives. Users no longer need to create or manipulate these directly.
- Uses `DatasetWrapper` as a base class. This means variables such as `ff_prime` are explicitly defined on the `psi` grid, and you can easily write an `Equilibrium` to file using the `to_netcdf` function. You can also easily view an equilibrium using things like:
    - `plt.contourf(eq["r"], eq["z"], eq["psi_rz"].T)`
    - `plt.plot(eq["psi"], eq["q"])`
    - etc... we could also define our own plotting functions on the class.
- Uses well-defined units throughout. If the user wishes to define their equilibrium in terms of inches, it'll automatically convert.
- All 'prime' variables are derivatives with respect to regular `psi`, in units of Webers per radian, while previously some were with respect to `psi` and others were with respect to `psi_n`.
- `get_flux_surface` renamed to just `flux_surface`, and returns a much more than just the RZ coordinates of a flux surface (see FluxSurface below).
- Greatly expanded the docs to explain what each variable is, what they're used for, and what units they use. Descriptions are also associated with each variable via "long_name" attributes in the underlying netCDF.

__Issues__

- Some of the links in the docs don't work as expected, such as class names under the 'See Also' sections.
- `psi` is defined in Webers per radian, following COCOS 1-8, but I'm not sure how best to convert this if the user instead uses COCOS 11-18, where it's just in Webers. I _think_ we're following COCOS 1 (or maybe 2?), but there's no way to handle a decreasing `psi` or a negative safety factor.

## equilibrium_reader

I've introduced a class decorator which handles registration with the `equilibrium_readers` factory. This means registration occurs in the same place a class is defined, instead of assigning a key to a class using a dict-like interface in the `__init__.py` file. Overall I think this is a much neater solution to the problem. If I push similar changes throughout Pyrokinetics, I may also be able to remove some of the confusing bloat associated with `Reader`, `Factory`, etc.

## GEQDSK/TRANSP readers

I updated the GEQDSK and TRANSP equilibrium readers to be compatible with the new system. I was able to simplify the code in some places, and I added a lot more comments to explain what's going on in the case of TRANSP.

__Issues__

- The GEQDSK reader assumes COCOS 1, but according to [this](https://crppwww.epfl.ch/~sauter/cocos/Sauter_COORD_CONVENTIONS_COCOS_2012_updated_after_reprint_for_Appendices_and_refs.pdf), EFIT uses COCOS 3. I think the COCOS used by a GEQDSK file depends on the software that generated it, but the file itself doesn't contain this metadata. We may need to expand this code to handle a few different GEQDSK sources.
- I can't access TRANSP docs, so I've made some assumptions regarding what each variable in their netCDF files refers to. I tried to explain this in comments wherever possible. I'm still following the same algorithm as before, but in some cases I'm drawing data from different grids.

## FluxSurface

In order to create `LocalGeometry` objects from the previous `Equilibrium` object, the first step is to get the RZ coordinates of the flux surface using `get_flux_surface`, and then these coordinates are passed back into `Equilibrium` to get `b_poloidal` along that path. Instead of doing this, `Equilibrium.flux_surface` now returns a new object which already contains all of the information the user might want on that FluxSurface. This is intended to be an intermediate object generated by `Equilibrium` from which we can create any `LocalGeometry`.

- Subclasses DatasetWrapper
- All quantities have well defined units.
- Contains `r`, `z`, `br`, `bz`, `bt`, `bp` on the theta grid.
- Contains `p`, `q`, `f`, `r_major`, `r_minor`, `z_mid`, and their `psi`-derivatives as attributes.
- The functions `psi_derivative`, `psin_derivative`, `r_minor_derivative` and `rho_derivative` can be used to get derivatives of any non-prime attribute, and it's clear if they contain any normalising factors.

__Issues__

- I wasn't sure what data FluxSurface should contain, so I just gave it everything! There's potential to narrow the list of data_vars/attributes.
- In PR https://github.com/pyro-kinetics/pyrokinetics/pull/112, an issue was brought up regarding difficulties fitting shape coefficients using just `R`, `Z` and `b_poloidal`. It may be necessary to add additional info to this object (Maybe dBp_dr and dBp_dz? Or maybe dR_drho/dZ_drho?).
- The IMAS/OMAS data schema uses 'gyrokinetics.flux_surface' to represent similar information to what we're calling 'local geometry'. There's some potential for naming confusion here.

## General changes/Issues

- I've been trying to stick more closely to some of the standard Python style recommendations:
    - Using snakecase for filenames instead of just naming them after the class they contain (something I picked up from writing Java about 12 years ago, along with rigid rules for class names. It's probably about time I broke those habits.)
    - Using snakecase for variable names. My worry is that this may cause confusion for some conventions, such as 'R' meaning major radius and 'r' meaning minor radius, though I've tried to make the distinction clear via docstrings etc.
- I made some changes to `LocalGeometryMiller` to keep things working for now. It's currently quite ugly, as units are all stripped at this stage. I'm hoping that, over time, units will proliferate throughout the code, and the messy interfaces between unit-aware code and non-unit-aware code will disappear.
- Instead of `LocalGeometry` having a function `load_from_eq(eq, psi_n)`, might it be better to give `Equilibrium` a function `local_geometry(psi_n, geometry_type)`? Not something urgent, but it may be cleaner for a global equilibrium to know what a local geometry is and provide methods to generate one than for a local geometry to know what a global equilibrium is and build itself from one.


bpatel2107:
Is this worth re-visiting now? 

Given that #112 is complete, it would be worth handling the merge conflicts (which might be messy sadly). I'm not sure how much this will change given the recent updates to `LocalGeometry` but I think it's not a big difference. The only change would be that the `LocalGeometry` is loaded using the `FluxSurface` Object instead.

Which netCDF attributes have units? 

The patch fix for  the `pint-xarray` float issue  seems fine but I think we should go to the `pint-xarray` team as its more an issue on their side if I've understood this correctly.

I believe there is a set way of converting between COCOS conventions but I'll need to look it up but I know it has been done before. This should handle any issues with decreasing `psi` or negative `q`. I think we should convert any input file to a pyrokinetics COCOS which we can set in the normalisations. Currently I think we use COCOS=2 (definitely 1 or 2 though)? I also think you can figure out the COCOS by looking at the signs of things details  in Tables 2/3 [here](https://www.sciencedirect.com/science/article/pii/S0010465512002962)

The TRANSP documentation is pretty light but I can go through and confirm everything.

I feel quite strongly about keeping the major radius as capital `R` as I think it will be far too confusing otherwise, even with the documentation. I think `R` (and maybe `Z`) should be the only variables that break the snakecase convention.

I can either go through everything now or after the merge conflicts have been handled. Up to you @LiamPattinson

LiamPattinson:
> Is this worth re-visiting now?

Incredible timing, I just finished work on another project and was planning on returning to this today!

> Given that https://github.com/pyro-kinetics/pyrokinetics/pull/112 is complete, it would be worth handling the merge conflicts (which might be messy sadly). I'm not sure how much this will change given the recent updates to LocalGeometry but I think it's not a big difference. The only change would be that the LocalGeometry is loaded using the FluxSurface Object instead.

As far as I can tell, this would still only need minor changes to `LocalGeometry` in order to work properly. The merge/rebase conflicts are likely to be a serious headache, so I was planning on starting again with a fresh branch/PR and closing this one here.

> Which netCDF attributes have units?

Sorry, not sure if I explained myself well in my original post. Usually, variables/coordinates are assigned an attribute `units` in NetCDF files, but Xarray `DataArray`'s don't actually carry units, and are just plain Numpy arrays under-the-hood. `pint-xarray` allows you to combine the underlying Numpy array and `units` attribute  to Pint `Quantity` objects, so the units are properly carried with the data, and it converts back to a text-based `units` attribute when you write to disk. The problem I ran into is that global attributes can't have units assigned (as a NetCDF attribute can't carry its own attributes), so if you use a global attribute to store something like the minor radius, it's just a plain number instead of a `Quantity`, and it takes some extra effort to add the units back in. I already have an idea in mind to get around this problem.

> The patch fix for the pint-xarray float issue seems fine but I think we should go to the pint-xarray team as its more an issue on their side if I've understood this correctly.

I chased up this problem with `f90nml`, and the latest commits now handle Numpy arrays much better (0D or otherwise). It's not yet in the latest release, but at some point we may be able to just bump the f90nml version to 1.5 and all shall be well. I'm not sure if it would be possible to fix the `force_ndarray` issue entirely, as I'm not yet sure why `pint-xarray` requires this in the first place.

> I feel quite strongly about keeping the major radius as capital R as I think it will be far too confusing otherwise, even with the documentation. I think R (and maybe Z) should be the only variables that break the snakecase convention.

I'll be sure to change things back in my new branch. I'm planning on also going back to `b_poloidal` and `b_toroidal` instead of `bp` and `bt` etc, since I feel like some of my changes have sacrificed clarity for the sake of reducing verbosity. I'll probably also go back to the original method for naming files/classes for the sake of consistency.

---

I'll close this here and try to get a new PR submitted soon-ish. Can't promise on having it done by the end of the year, but I'll see what I can do. Let me know if you'd prefer to re-open it for any reason.

PR #133, opened 2022-11-16
Better method for converting simulation/physical units

### LINKED ISSUE #132: Density/temperature not normalised correctly
When running the example scripts `example_JETTO.py` and `example_TRANSP.py` the density and temperature are not being normalised correctly.

For example in the output of `example_JETTO.py` in `test_jetto.gs2` the electron density is written as 

```
    z = -1
    dens = 6.959310030892369e+20
    temp = 0.49999999999999994
```

Where both `dens` and `temp` should be 1.0

@ZedThree thoughts?

ZedThree:
Instead of converting to base units, we construct a new unit,
replacing base units with the ones in the destination convention, and
then we can convert the quantity directly.
This avoids problems with the derived reference units being converted
to base units, e.g. nref to lref**-3

We now also handle combined units more flexibly, by trying to
transform each individual unit one at a time

Fixes #132

Uses the new "facet" system in pint 0.20 which was released very recently, so users will need to make sure to update that

### DIFF
--- a/pyrokinetics/normalisation.py
+++ b/pyrokinetics/normalisation.py
@@ -154,8 +154,85 @@ class PyroNormalisationError(Exception):
         )
 
 
-def _create_unit_registry() -> pint.UnitRegistry:
-    """Create a default pint.UnitRegistry with some common features we need"""
+class PyroQuantity(pint.Quantity):
+    def to_base_units(self, system: Optional[str] = None):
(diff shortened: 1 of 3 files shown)


bpatel2107:
LGTM

bpatel2107:
I _think_ I understand what you've done and it seems like a much better solution. Changing the individual units seems like a safer way.

PR #134, opened 2022-11-17
Bugfix/remove template beta

bpatel2107:
Currently when loading in a template, the `numerics` are set from the template including `beta`. This `beta then carries through the whole calculation and would override the actual `beta` as shown here

https://github.com/pyro-kinetics/pyrokinetics/blob/0cd4b168e85ba152de93dbd80367d43ab6fa3412/pyrokinetics/gk_code/GKInputCGYRO.py#L342-L343

This forces `beta` back to a `None`, meaning it will get set appropriately.

@ZedThree do you think there is a better place for this? As far as I can tell, `beta` is the only term in `numerics` which would be impacted physical parameters so we won't have to do this anywhere else

### DIFF
--- a/pyrokinetics/local_species.py
+++ b/pyrokinetics/local_species.py
@@ -143,7 +143,9 @@ class LocalSpecies(CleverDict):
             species = self[name]
             # Total pressure
             pressure += species["temp"] * species["dens"]
-            a_lp += species["temp"] * species["dens"] * (species["a_lt"] + species["a_ln"])
+            a_lp += (
+                species["temp"] * species["dens"] * (species["a_lt"] + species["a_ln"])
+            )
 
         self["pressure"] = pressure
 
--- a/pyrokinetics/local_species.py
+++ b/pyrokinetics/local_species.py
@@ -143,7 +143,7 @@ class LocalSpecies(CleverDict):
             species = self[name]
             # Total pressure
             pressure += species["temp"] * species["dens"]
-            a_lp += pressure * (species["a_lt"] + species["a_ln"])
+            a_lp += species["temp"] * species["dens"] * (species["a_lt"] + species["a_ln"])
 
         self["pressure"] = pressure
 
--- a/pyrokinetics/numerics.py
+++ b/pyrokinetics/numerics.py
@@ -38,7 +38,7 @@ class Numerics(CleverDict):
             "bpar": False,
             "max_time": 500.0,
             "delta_time": 0.001,
-            "beta": 0.0,
+            "beta": None,
         }
 
         super().__init__(data_dict)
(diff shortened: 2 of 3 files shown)


bpatel2107:
Turns out the local_species wasn't correctly calculating `a_lp`. It was multiplying the gradients by the cumulative pressure rather than the species pressure

d7919:
> Turns out the local_species wasn't correctly calculating `a_lp`. It was multiplying the gradients by the cumulative pressure rather than the species pressure

What are the consequences of this? It looks like this means that beta_prime in CGYRO instances was wrong, did this ever get into input files?

bpatel2107:
Yes it would have impacted any CGYRO input file but this is the only place it would have had an impact. And it would only really matter when loading data in from a kinetic profile. Loading in a standalone gk file would not be affected.

This seemed to have been an issue since September 23rd (d846691c24dde460692f63d6093742b01a5ebd04) so it may have impacted a few people's calculations

@mgiacomin does this impact your work.

d7919:
> Yes it would have impacted any CGYRO input file but this is the only place it would have had an impact. And it would only really matter when loading data in from a kinetic profile. Loading in a standalone gk file would not be affected.
> 
> This seemed to have been an issue since September 23rd ([d846691](https://github.com/pyro-kinetics/pyrokinetics/commit/d846691c24dde460692f63d6093742b01a5ebd04)) so it may have impacted a few people's calculations
> 
> @mgiacomin does this impact your work.

So if we construct a cgyro input direct from equilibrium it's an issue but if we got it by converting from another code we would be ok?

bpatel2107:
Converting from another code would also cause the issue. Only standalone runs where no conversion is done would not face this issue

d7919:
I'm probably being a bit slow, but also is this unlikely to impact resulting CGYRO runs because this usually calculates beta_prime consistently from the inputs? So it's most likely to be a problem from runs with other codes which have come from a converted CGYRO input?

(also the commit was only merged 27 days ago so actually just a problem for runs since ~21st October?)

bpatel2107:
Ah good point about the merge.

Essentially the calculation of `a_lp` was incorrect. This means that the conversion from `BETA_STAR_SCALE` in CGYRO to the `local_geometry.beta_prime` was incorrect and vice versa. 

It was however the same factor going both ways, so if you convert from `BETA_STAR_SCALE` (correct) -> `local_geometry.beta_prime` (incorrect) -> `BETA_STAR_SCALE` (correct) then you would end up with the same number for `BETA_STAR_SCALE`. This is essentially what happens when you load in a CGYRO gk file.

But if `local_geometry.beta_prime` was calculated somewhere else then when mapping to `BETA_STAR_SCALE` would give you an incorrect value. So you'd be doing `local_geometry.beta_prime` (correct) -> `BETA_STAR_SCALE` (incorrect)

So in the case where you load the equilibrium from TRANSP/JETTO, then `local_geometry.beta_prime` would be the physical value but would get incorrectly mapped to `BETA_STAR_SCALE`

mgiacomin:
@bpatel2107 @d7919 I think I am not affected by this issue, since the CGYRO input files I am using or used in the past were generated with a rather old version of pyrokinetics. Anyway, it would be good to merge this fix as soon as possible.

PR #137, opened 2022-11-23
fix sign for poloidal field fit

hgd511:
Poloidal field fitting was failing due to poloidal field magnitude being calculated as negative, due to sign of dpsidr. Fixed to absolute value


bpatel2107:
LGTM

PR #139, opened 2022-11-24
Bugfix/normalisation gene

### LINKED ISSUE #138: Collisional frequency not converting in GENE
Without reference values the collision frequency is not converting correctly.

Running `example_gene.py`  results in the following error. Surprised the tests didn't catch this but the other template files happens to have `nu=0` so that may be why.

I thought the individual transformations were handled but it looks like it gets a NaN when converting

```
  File "/home/ir-pate3/pyrokinetics/pyrokinetics/normalisation.py", line 862, in convert_dict
    value = value.to(norm).magnitude
  File "/home/ir-pate3/pyrokinetics/pyrokinetics/normalisation.py", line 221, in to
    return self._replace_nan(value, other)
  File "/home/ir-pate3/pyrokinetics/pyrokinetics/normalisation.py", line 167, in _replace_nan
    raise PyroNormalisationError(system, self.units)
pyrokinetics.normalisation.PyroNormalisationError: Cannot convert 'vref_nrl / lref_major_radius' to 'gs2_input0000' normalisation. Possibly 'gs2_input0000' is missing physical reference values. You may need to load a kinetics or equilibrium file

The above exception was the direct cause of the following exception:

  File "/home/ir-pate3/pyro_testing/example_gene.py", line 9, in <module>
    pyro.gk_code = "GS2"
  File "/home/ir-pate3/pyrokinetics/pyrokinetics/pyro.py", line 302, in gk_code
    self._switch_gk_context(value, force_overwrite=False)
  File "/home/ir-pate3/pyrokinetics/pyrokinetics/pyro.py", line 430, in _switch_gk_context
    self.update_gk_code()
  File "/home/ir-pate3/pyrokinetics/pyrokinetics/pyro.py", line 489, in update_gk_code
    self.gk_input.set(
  File "/home/ir-pate3/pyrokinetics/pyrokinetics/gk_code/GKInputGS2.py", line 525, in set
    self.data[name] = convert_dict(namelist, local_norm.gs2)
  File "/home/ir-pate3/pyrokinetics/pyrokinetics/normalisation.py", line 864, in convert_dict
    raise ValueError(
ValueError: Couldn't convert 'vnewk' (0.0002813941179127954 vref_nrl / lref_major_radius) to gs2 normalisation. This is probably because it did not contain physical reference values. To fix this, please add a geometry and/or kinetic file to your Pyro object.
```

bpatel2107:
Ended up going down a rabbit hole but made it out! Fix #138 

For the mapping from one normalisation to another work if the `kinetics` and `equilibrium` object were defined but without them the transformation between the two did not work so I've added a `set_ref_ratios` method to normalisation that only requires either a `LocalGeometry` object that allows for the conversion between `lref_major_radius` <-> `lref_minor_radius` and `bref_B0` <-> `bref_Bunit` . This gets called after the `LocalGeometry` is loaded for a particular `gk_code`

You can also just specify the `aspect_ratio = R/a` which does the `lref` mapping and was needed for GENE when no `Normalisation` is given but a transform is needed.

There were also a few bugs that I happened to fix on the way. It seems that the collisionality was being defined with units of `lref_major_radius` even in cases where the input file had the minor radius as the normalisation. Now we automatically change the parameter from the input file to be normalised to the minor radius and gave it the right units. This does mean that GENE collisionalities have been incorrect since this was merged.

Having the default `lref` for GENE as the major radius has made parts of the code quite cumbersome but I'm not sure we can get around that. For now the files aren't written with different normalisations so it doesn't impact anything yet. When we add units to the `LocalGeometry` hopefully this will become easier.

There was an issue where the aspect_ratio was assumed to be `local_geometry.Rmaj` / `local_geometry.a_minor` when converting between `lref_major_radius` <-> lref_minor_radius` but it should just be `local_geometry.Rmaj`

We also lucked out by having the templates use `beta=0.0` which hid the fact that when we load in a `gk_file`, `kinetics` and `equilibrium`, beta was taken from `gk_file` by default which is not ideal behaviour. We should always take `beta` from the `kinetics`+`equilibrium` when initialising a `Pyro` object


ZedThree:
Please could you add some more tests? There's a few I can immediately see:

- Reading GENE input file and converting to GS2 with no equilibrium/kinetic file
- Getting correct length normalisation from GENE input file
- `beta` set correctly when reading GK + equilibrium + kinetics

Basically, fixing a bug implies adding a new test :)

pyrokinetics/normalisation.py
```diff
@@ -695,6 +697,38 @@ def set_lref(
             lambda ureg, x: x.to(ureg.lref_minor_radius).m * self.pyrokinetics.lref,
         )
 
+    def set_ref_ratios(
+        self,
+        local_geometry: Optional[LocalGeometry] = None,
+        aspect_ratio: Optional[float] = None,
```
ZedThree, line 703: We should `raise ValueError` if neither are given

pyrokinetics/normalisation.py
```diff
@@ -672,13 +672,15 @@ def set_lref(
 
         if local_geometry:
             minor_radius = local_geometry.a_minor
-            major_radius = local_geometry.Rmaj
-
-        # FIXME: We could also set this from minor_radius * aspect_ratio
-        major_radius = major_radius or 0.0
+            aspect_ratio = local_geometry.Rmaj
```
ZedThree, line 675: I'm really not sure about this -- surely this is only true if `a_minor == 1.0`? In which case, it would still be correct to have `aspect_ratio = major_radius / minor_radius`

bpatel2107: At the moment `Rmaj` is defined as `major_radius/minor_radius`, so it's the normalised major radius.

Eventually we should have the normalising length scale be an option but we haven't set up units in the `LocalGeometry` and that's probably a separate PR. But currently this is always the aspect ratio.

https://github.com/pyro-kinetics/pyrokinetics/blob/fb278e4ab7ee28a9899beca0757013368086fb4e/pyrokinetics/local_geometry/LocalGeometry.py#L159

bpatel2107:
Added in some tests now. Let me know if anything else is needed!

PR #140, opened 2022-11-25
Enhancements/add zeff

bpatel2107:
Added in Zeff calculation to `LocalSpecies` and set value for the various different `GKCodes`


d7919:
Looks good to me. Are all the Zeff definitions the same in each code?

bpatel2107:
Yes they all seem to be `sum_{ions} n_i Z_i**2 / n_e Z_e`

PR #141, opened 2022-12-01
Fix incorrect collisionality for ions when Z!=1

bpatel2107:
When setting the collisionality of different ion species from CGYRO/GENE/TGLF, a factor of `Z**4` was missing.

This only impacts results where CGYRO/GENE/TGLF is converted to GS2.  Loading from profile data or a GS2 input file loads in the collisionality correctly. 

Converting between CGYRO/GENE/TGLF would also be fine as this gets the electron collision freq correct and that is what is needed.

### DIFF
--- a/pyrokinetics/examples/example_JETTO.py
+++ b/pyrokinetics/examples/example_JETTO.py
@@ -16,7 +16,9 @@ def main(base_path: Union[os.PathLike, str] = "."):
 
     pyro.load_global_eq(eq_file=eq_file, eq_type="GEQDSK", psi_n_lcfs=1.0)
 
-    pyro.load_global_kinetics(kinetics_file=kinetics_file, kinetics_type="JETTO", time=550)
+    pyro.load_global_kinetics(
+        kinetics_file=kinetics_file, kinetics_type="JETTO", time=550
+    )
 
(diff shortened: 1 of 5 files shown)


ZedThree:
Please could you add tests? :slightly_smiling_face: 

I wonder if it's worth bringing in plasmapy, which has a [bunch of collision formulas](https://docs.plasmapy.org/en/stable/formulary/collisions.html)

bpatel2107:
Yes I always forget the tests... One day you might not need to remind me

LiamPattinson:
> I wonder if it's worth bringing in plasmapy, which has a [bunch of collision formulas](https://docs.plasmapy.org/en/stable/formulary/collisions.html)

It would be nice to use this so that we didn't have to reinvent the wheel, but since plasmapy uses astropy units, would it not cause issues with our units/normalisations?

ZedThree:
It should be possible to convert between the two libraries -- we'd only be passing physical units, so this should be fine

bpatel2107:
Added tests

bpatel2107:
When we have reference values plasmapy should work but when we don't we'll need to work out how to convert the `gk_code` `nu` to the pyro `nu`

Plasmapy would come in in the `LocalSpecies` class

PR #142, opened 2022-12-07
Set squareness to zero when not specifeid in CGYRO input

mgiacomin:
The squareness is not always specified in CGYRO input file, especially if generated with a previous version of `pyrokinetics`. In this case, the squareness should be set to zero.

### DIFF
--- a/pyrokinetics/gk_code/GKInputGENE.py
+++ b/pyrokinetics/gk_code/GKInputGENE.py
@@ -165,7 +165,7 @@ class GKInputGENE(GKInput):
 
         if miller.B0 is not None:
             miller.beta_prime = -self.data["geometry"].get("amhd", 0.0) / (
-                miller.q ** 2 * miller.Rmaj
+                miller.q**2 * miller.Rmaj
             )
 
         return miller
@@ -266,8 +266,8 @@ class GKInputGENE(GKInput):
             # Not exact at log(Lambda) does change but pretty close...
             local_species[key]["nu"] = (
                 nu_ee
-                * (zion ** 4 * nion / tion ** 1.5 / mion ** 0.5)
-                / (ne / te ** 1.5 / me ** 0.5)
+                * (zion**4 * nion / tion**1.5 / mion**0.5)
+                / (ne / te**1.5 / me**0.5)
             ).m * nu_ee.units
 
         local_species.zeff = (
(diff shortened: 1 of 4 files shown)


bpatel2107:
I agree this is an issue but this currently sets all Miller parameters to 0 by if not defined. We should make a new list or tweak the list which has the default for each parameter

bpatel2107:
@mgiacomin what do you think of this? I've defined a separate default value dict that is used. 

Should probably add this in for the other code too

mgiacomin:
@bpatel2107 Yes, I think adding a separate list is a good idea. 
Yes, I think this should be extended to the other codes as well.

bpatel2107:
@dake0795  Can you confirm the GENE defaults?

pyrokinetics/gk_code/GKInputGENE.py
```diff
@@ -123,13 +140,17 @@ def get_local_geometry_miller(self) -> LocalGeometryMiller:
...
-            miller_data[pyro_key] = self.data[gene_param][gene_key]
+        for (pyro_key, (gene_param, gene_key)), gene_default in zip(
+            self.pyro_gene_miller.items(), self.pyro_gene_miller_default.values()
+        ):
+            miller_data[pyro_key] = self.data[gene_param].get(gene_key, gene_default)
 
-        miller_data["Rmaj"] = self.data["geometry"]["major_r"] / self.data[
+        miller_data["Rmaj"] = self.data["geometry"].get("major_r", 1.0) / self.data[
             "geometry"
         ].get("minor_r", 1.0)
```
bpatel2107, line 151: @dake0795 Does `minor_r` have a default? From the manual I have it seems not so not sure what to do if it isn't given

dake0795: No `minor_r` doesn't have a default. Its default should probably be 1.

dake0795:
@bpatel2107 GENE default is 0.0 for all Miller parameters.

pyrokinetics/gk_code/GKInputGENE.py
bpatel2107: This could potentially lead to bad behaviour here so if it is not defined then this won't work. Will add a todo here

PR #145, opened 2022-12-15
Equilibrium refactor and enhancements

LiamPattinson:
This PR is a follow-up from PR https://github.com/pyro-kinetics/pyrokinetics/pull/128. It's mostly a rebase of that PR on top of the latest Pyrokinetics changes, so please see the wall-of-text I wrote there for more info. For info on how to use these new features, I'd recommend checking out the tutorial I've written in `docs/tutorials/equilibrium.rst`.

Summary of work from PR https://github.com/pyro-kinetics/pyrokinetics/pull/128:

- `Equilibrium` now takes a fixed set of parameters to its `__init__` function. All parameters provided are scalars or arrays.
- `Equilibrium` can be initialised without reading a file. Building an `Equilibrium` from a file is now delegated to the function `read_equilibrium`.
- `Equilibrium` acts like an Xarray dataset, with some variables defined on the $\psi$ grid and others on the $(R,Z)$ grid.
- Using [pint-xarray](https://pint-xarray.readthedocs.io/en/stable/), all variables have units. Units are converted when passing arrays into `Equilibrium.__init__`.
- The function `flux_surface` returns a new class instead of just the $(R,Z)$ path of the flux surface. The `FluxSurface` class contains all of the information needed to produce a `LocalGeometry` from a global `Equilibrium` (arguably it contains too much information, and could be stripped down a bit).
- Refactor of GEQDSK and TRANSP readers to use the new system.
- Cleaner method to register `Reader` classes.
- Added a lot to the docs.
- Units are discarded from `Equilibrium`/`FluxSurface` on conversion to `LocalGeometry`, as that section of the code will need its own refactor to add units.

Changes from PR https://github.com/pyro-kinetics/pyrokinetics/pull/128:

- Fixed `DatasetWrapper` so that global attributes can have units.
- `r` and `z` coordinates are now `R` and `Z` throughout. Similarly, `br` is back to `b_radial`, `bp` is `b_poloidal`, etc.
- Can save `Equilibrium` objects using `eq.to_netcdf("myfile.nc")`, and read them back in using `eq = read_equilibrium("myfile.nc")`, so Pyrokinetics itself is recognised as an equilibrium file type.
- `FluxSurface` derivatives are generated on-the-fly using attribute lookup, i.e `fs.dq_drho` gives the derivative of the safety factor with respect to $\rho$, `fs.dR_major_dr_minor` gives the Shafranov shift.
- Split `normalisations.py` into two files, as the dependency on `LocalGeometry` was causing circular import errors (normalisations depends on LocalGeometry, which depends on Equilibrium, which depends on normalisations in order to use `ureg`). There's probably a cleaner fix for this out there, such as removing the explicit dependency on `LocalGeometry` and just passing `a_minor` instead. For now, I moved everything up to `PyroUnitRegistry` into `units.py`.
- Added a tutorial section to the docs, explaining how to use `Equilibrium` and `FluxSurface` objects.
- Added plotting functions to `Equilibrium` and `FluxSurface`.

The plotting functions might not produce anything of publication quality, but you can pull of some cool things pretty easily:

```python
import pyrokinetics as pk
eq = pk.read_equilibrium(pk.eq_templates["GEQDSK"])
ax = eq.contour(levels=50)
ax = eq.flux_surface(0.9).plot_path(ax=ax)
```
testplot

Let me know if you'd like to see any further changes, or if there's some aspect of the design you think should be changed.


pyrokinetics/equilibrium/geqdsk.py
```diff
@@ -0,0 +1,128 @@
+from contextlib import redirect_stdout
+
+from .equilibrium import Equilibrium, equilibrium_reader
+from .flux_surface import _flux_surface_contour
+from ..readers import Reader
+from ..typing import PathLike
+from ..units import ureg as units
+
+import numpy as np
+from freegs import _geqdsk
```
bpatel2107, line 12: Might be worth switching over to reading via FreeQDSK which is a bespoke GEQDSK file reader rather than using a GEQDSK reader in a equilibrium solver. I think all the variable names are the same so hopefully its a simple switch


https://github.com/ZedThree/FreeQDSK

pyrokinetics/examples/example_JETTO.py
```diff
@@ -14,7 +14,7 @@ def main(base_path: Union[os.PathLike, str] = "."):
     # Load up pyro object
     pyro = Pyro()
 
-    pyro.load_global_eq(eq_file=eq_file, eq_type="GEQDSK", psi_n_lcfs=1.0)
```
bpatel2107, line 17: The point of `psi_n_lcfs` was to artificially change where the actual LCFS is. The reason for this was that sometimes, when doing the contour fittings, if the actual value of `psi` was used to specify the LCFS, a connected contour wouldn't be found. So I gave the option of slightly shifting the value of `psi` for the LCFS.

Admittedly this is more of a issue with the contour fitting rather than the equilibrium but it was a simple workaround. I see that you can now you specify `psi_lcfs` directly. Can we add a multiplier to slightly change the value if needed. If you are loading in an equilibrium you likely won't know what `psi_lcfs` is but you could easily specify `psi_n_lcfs` which defines the `psi` used

bpatel2107:
Overall this looks good to me. 

The plot of the poloidal field in the docs is a bit concerning. so we should figure out why it is that oscillatory. Something has likely gone wrong here

bpatel2107:
So having a quick look it seems like `psi_n_lcfs` is actually in here but implementation needs a subtle tweak.

The only slight issue is that in the following lines the interpolation assumes the final element in the GEQDSK arrays matches `psi_n_lcfs` but that is always at `psi_n = 1.0` so if we were to use `psi_n_lcfs=0.9` then it would be slightly off.

https://github.com/PlasmaFAIR/pyrokinetics/blob/unstable/pyrokinetics/equilibrium/EquilibriumReaderGEQDSK.py#L49-L52

What we should do is to fit that using the full range and then redefine the `psi_n` array using `psi_n_lcfs` after `q`, `f_psi`, `ff'` and `pressure` are set.

Are you able to look at the merge conflicts here?

LiamPattinson:
@bpatel2107 I'll be taking a look at this for a few hours today, and I'll probably continue working on it tomorrow and on Friday. I'll try to get `psi_n_lcfs` working again, and I'll see what's going wrong with the merge conflicts. I got some further feedback in person from @ZedThree, so I'll be making those changes too.

LiamPattinson:
Sorry, disregard the last few commits, I don't think I understand how `psi_n_lcfs` is supposed to be used.

In the original code, this changes the grid on which each quantity is defined, and therefore adjusts the derivatives:

https://github.com/pyro-kinetics/pyrokinetics/blob/95176763ea819eb7abf265c064961a47cf9067fa/pyrokinetics/equilibrium/EquilibriumReaderGEQDSK.py#L42
https://github.com/pyro-kinetics/pyrokinetics/blob/95176763ea819eb7abf265c064961a47cf9067fa/pyrokinetics/equilibrium/EquilibriumReaderGEQDSK.py#L49-L53

This also means that if we set `psi_n_lcfs` to 0.9, and if the user tries to get data at `psi_n=0.95` later, it'll be determined by extrapolating from the splines. However, `a_minor` is set by determining the maximum width of the last flux surface:

https://github.com/pyro-kinetics/pyrokinetics/blob/95176763ea819eb7abf265c064961a47cf9067fa/pyrokinetics/equilibrium/EquilibriumReaderGEQDSK.py#L69-L75

https://github.com/pyro-kinetics/pyrokinetics/blob/95176763ea819eb7abf265c064961a47cf9067fa/pyrokinetics/equilibrium/EquilibriumReaderGEQDSK.py#L82

This isn't affected by `psi_n_lcfs` as far as I can tell, so does this imply that `psi_n` and `rho` are normalised with respect to different flux surfaces? My reading is that `psi_n` is normalised with respect to a flux surface outside the bounds of the GEQDSK file, while `rho` is normalised with respect to the GEQDSK LCFS.

Is this the expected behaviour? My current implementation does the opposite, with everything calculated on the `psi` grid as usual (not `psi_n`), and if the user sets `psi_n_lcfs=0.9`, it sets `psi_lcfs` in the Equilibrium object to `0.9 * psi_lcfs + 0.1 * psi_axis`, and calculates `a_minor` on the same flux surface:

```python
        # Adjust normalising factors if psi_n_lcfs is not 1.0
        psi_lcfs = psi_n_lcfs * psi_lcfs + (1.0 - psi_n_lcfs) * psi_axis
        a_minor = UnitSpline(psi_grid, r_minor)(psi_lcfs)
```

bpatel2107:
To clarify I really don't like `psi_n_lcfs` but it was the only hacky fix I could think of.

Essentially the point of it was to get a fit for the LCFS when matplotlib failed to get a nice contour. For some GEQDSK files, when we try to find the LCFS by looking at a contour of `psi_lcfs`, it would not find a good fit and I never figured out why.

Here's an example and it still happens using scikit-image
image

I found that a way around this was to slightly look inside of the LCFS by looking at `0.9999 * psi_lcfs` which would seemed to behave better. The overall change was meant to be quite small so it didn't have a significant impact on the other quantities.
It really wasn't made for a number like 0.9

Essentially if we can work around it then we can completely ignore this. 

If we want to keep this then things like `f_psi` should be initially interpolated from 0 to 1 as that is how they are defined in the GEQDSK. If we do it from 0 to `psi_n_lcfs` to begin with then when we map across to flux surfaces, they will be slightly shifted.

So we should start off with 

```python
psi_n = np.linspace(0.0, 1.0, nr) 
```

And for this example assuming `psi_bdry` is actually 10 from the GEQDSK file and we want `psi_n_lcfs = 0.9`

Do an initial interpolations for `f_psi` `q` etc. and get them from 0 up to `psi_n_lcfs * psi_lcfs`, so 0->9

We interpolate those values onto a new `psi_n` which does go from 0 to 1. We also change the value of `psi_bdry`

Then when we fit our flux surfaces we are always using data normalised to the same point.

This does mean that when a user tries to implement `psi_n=0.5` its not actually 0.5 from the GEQDSK.



The alternative is that we scrap this entirely and find a way to fit to the LCFS, which is probably the better solution

LiamPattinson:
I've made some changes to the PR, but it's nothing major:

- A bunch of things that were at global scope are now class-level attributes/classmethods on `Equilibrium`. Plain function alternatives are still available, but now these simply forward args to their corresponding classmethods. Now we can build an `Equilibrium` from a file using either `Equilibrium.from_file` or `read_equilibrium`.
- `FluxSurface` has been cleaned up, and now its attributes are just the things needed by `LocalGeometry` classes, though named following the [OMAS schema `gyrokinetics.flux_surface`](https://gafusion.github.io/omas/schema/schema_gyrokinetics.html). I think it may be possible to merge these classes with some further work, as they're both describing the same thing -- the only fundamental difference is that `LocalGeometry` is parameterised while `FluxSurface` isn't. That's something for a future PR though.
- Minor updates throughout

I think I'm following RE `psi_n_lcfs`, and if I understand the intent correctly, then I think my current implementation does what's required. Since all of the `psi` grid quantities are determined on `psi` splines instead of `psi_n` splines, all we're doing is adjusting `psi_lcfs` and `a_minor` slightly, and nothing else is changed in `Equilibrium`. The only practical difference is that if the user sets `psi_n_lcfs=0.9`, and then looks up something at `psi_n=0.5`, it'll actually end up looking at `psi_n=0.45` in the GEQDSK file. Let me know if it looks like this won't achieve what's needed or if I'm still misunderstanding.

I'll see if I can get a switch to FreeQDSK working tomorrow, though that might require some quick work on FreeQDSK itself.

bpatel2107:
I believe this will still fail for certain GEQDSK files as the error occurs when finding all of the flux surfaces which happens before your tweak here.

If `psi_n_lcfs` is not specified then you are still trying to fit to the actual LCFS here as `psi_grid` has `psi_lcfs` in it.


Essentially we could  do this line first

https://github.com/pyro-kinetics/pyrokinetics/blob/89a47c50c6d7a0c762a01a9724ca0d83978d880f/pyrokinetics/equilibrium/geqdsk.py#L113

Then adjust the final value in `psi_grid` and then loop through the flux surface fitting here

https://github.com/pyro-kinetics/pyrokinetics/blob/89a47c50c6d7a0c762a01a9724ca0d83978d880f/pyrokinetics/equilibrium/geqdsk.py#L99-L100

Then we could set `a_minor`

https://github.com/pyro-kinetics/pyrokinetics/blob/89a47c50c6d7a0c762a01a9724ca0d83978d880f/pyrokinetics/equilibrium/geqdsk.py#L114

It might be a good idea to set a lower limit on `psi_n_lcfs` to something like 0.99 until we handle this contour issue.

Alternatively we could sidestep this issue and handle this case inside the flux surface fitting. We could have a check if the flux surface fails when `psi_n=1.0` and try again with `psi_n_lcfs=0.999` and raise a warning. Not sure if that is better practice or not.

bpatel2107:
I like the idea of merging `LocalGeometry` and `FluxSurface` as they are describing the same thing

LiamPattinson:
> I believe this will still fail for certain GEQDSK files as the error occurs when finding all of the flux surfaces which happens before your tweak here.

Good point, I hadn't considered that. It would be useful to have a GEQDSK file that fails to fit `psi_n_lcfs = 1.0` in the tests, but I think I've got it this time. I tried to make sure that we're not changing any data beyond that near the boundary:

- Work out new `psi_lcfs`
- Discard any elements of `psi_grid` that are beyond this point (`f`, `p`, `q`, `ff_prime`, `p_prime`)
- Interpolate quantities from the original `psi_grid` to the new one
- Use the new `psi_grid` to perform contour fitting and calculate `R_major`, `r_minor`, `Z_mid`, `a_minor`

I think this results in everything being consistent, but as there's a linear interpolation step, there may be a loss of accuracy due to the steep gradients near the plasma boundary (although I'm not sure how to avoid this).

bpatel2107:
I think what you have done is probably the best we can do. Happy to merge if this is ready

LiamPattinson:
I could try to get FreeQDSK in this PR, since the latest FreeGS release isn't compatible with the latest NumPy, but I suppose that could also be a separate PR. Otherwise I don't think I have anything else to add.

bpatel2107:
Can add in FreeQDSK in another PR.

We should be careful about the reversing of the psi. It's likely that the GEQDSK there have a different COCOs (coordinate convention https://www.sciencedirect.com/science/article/pii/S0010465512002962) that we need to handle with care. 

For now this is good!

PR #146, opened 2023-01-05
Set Jinja version limit in requirements.txt

bpatel2107:
Set `jinja2 < 3.1` in `requirements.txt` due to import error when setting up readthedocs

https://readthedocs.org/projects/pyrokinetics/builds/19084114/

Issue outlined here

https://github.com/readthedocs/readthedocs.org/issues/9038

Do we need both `requirements.txt` and the `setup.cfg`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,6 +9,7 @@ sphinx==3.5.4
 myst_parser==0.14.0
 cleverdict>=1.9.1
 xarray>=0.10
+jinja2 < 3.1.0
 # Not required for Python>=3.8, but included because conda does not support
 # conditional expressions in --file arguments.
 # This file should only be used by developers, so the odd extra package is not


LiamPattinson:
Nowadays we should be able to get away with just `pyproject.toml`, and could remove `requirements.txt`, `setup.cfg`, and `setup.py`. `setup.py` is still needed if you use `setuptools` for your build backend and you want to do editable installs, but even that is likely to be fixed soon (I think we could even get around this if we used `poetry` as the build backend in place of `setuptools`).

I've seen it argued that it's a good idea to keep `requirements.txt`, as it's easier for a human to read that than it is for them to find the relevant part of `setup.cfg`/`pyproject.toml`. I'd argue instead that it's too easy for the developers to make a mistake and allow the two documents to get out of sync, and that it's preferable to have a 'single source of truth'. Either method has its merits.

I'm currently working on some notes/teaching materials on Python packaging, so if you'd like I could take a look over the current setup and see if I can clean things up.

bpatel2107:
Yeah I think having one place where all the requirements are specified is best. I know there have been issues where the different sources become out of sync. If you can tidy that up then that would be awesome!

bpatel2107:
Handled in #147

PR #147, opened 2023-01-09
Upgrade to latest build system (PEP 517/518/621)

LiamPattinson:
I've updated the build system to use only `pyproject.toml`, since this seems to be the way the Python community is heading ([PEP 517](https://peps.python.org/pep-0517/), [518](https://peps.python.org/pep-0518/), [621](https://peps.python.org/pep-0621/)). Support for `pyproject.toml`-only builds has been limited with `pip`/`setuptools`, but it's getting pretty good nowadays and it's unlikely that behaviour will change significantly in later versions.

I also moved the tests into the top-level directory, [which seems to be the standard](https://github.com/pypa/sampleproject). I'd quite like to move `examples/` there too, but since they're used for testing they need to be left in the source folder for now.

Advantages:

- All build information is in one file, which acts as the 'single source of truth' regarding dependencies etc.
- Newer system is simplified and mostly uses standard keys (PEP 621). 

Known issues:

- Pip versions before 22 (or maybe 21?) require `setup.py` to do an editable install (`pip install -e .`). You should be able to get around this by upgrading pip (`pip install --upgrade pip`). Alternatively, I could put a `setup.py` back in there.
- flake8 doesn't use `pyproject.toml` yet (which has proven to be quite controversial). Added `.flake8` file to get around this.
- I'm not 100% sure if readthedocs will still work.

Other issues I found while doing this, which apply regardless of whether this PR is accepted:

- cleverdict doesn't use `pyproject.toml` at all, which means `pip` has to fall back on deprecated legacy features to build a wheel. This isn't a problem right now, but it might become a problem once `pip` gets to version 23.
- Numpy dropped support for alias types such as `np.bool` in 1.24. This isn't a problem for us, but FreeGS still uses them in the latest published version (0.6.1). I think this was fixed in 0.7, but that version is still unpublished. We might be switching to [FreeQDSK](https://github.com/ZedThree/FreeQDSK) soon anyway.


bpatel2107:
LGTM

- I think requiring recent versions of pip isn't a bad thing. Most people who are developing should probably use the latest version
- I can't build the readthedocs from your forked branch so once merged I can check
- Happy if we switch to FreeQDSK
- I think we can probably drop the use of cleverdict and just define the getitem and setitem like below. I don't think cleverdict is adding much else (not for this PR)
https://github.com/pyro-kinetics/pyrokinetics/blob/0bd28f932d58e2e8c10828b09d13cd22e27fac55/pyrokinetics/local_geometry/LocalGeometry.py#L114-L118

ZedThree (approved):

ZedThree:
Docs and tests work locally for me, thanks Liam!

PR #149, opened 2023-01-13
Fix/gene fields ballooning

mgiacomin:
I have added the phase factor in from of each connected domain when converting from flux-tube to ballooning space coordinates in GENE. Before, this was set to `-1` , which works only if `n0_global` is not defined in input file.


dake0795 (approved):
@mgiacomin, this looks great! Nice work catching this.

bpatel2107:
LGTM

bpatel2107 (approved):

Issue #152, opened 2023-01-24
CGYRO heat flux with flow shear


mgiacomin:
https://github.com/pyro-kinetics/pyrokinetics/blob/3d083304a212d2db1f873f261cd4ee585b0764dc/pyrokinetics/gk_code/GKOutputReaderCGYRO.py#L318-L321
Why is `cflux` loaded in place of `gflux` in CGYRO simulations with flow shear? According to the CGYRO documentation
```
gflux(0,ky) = Domain-averaged flux [this is the standard flux-tube flux]
cflux(ky) = Interior-averaged flux [this is the "central" flux]
```
we should always consider gflux, which is the standard flux-tube flux.  `cflux` includes only the average flux in the "positive-shear region".
@bpatel2107 What do you think?

bpatel2107:
This is what is done in the GACODE plotting scripts so I matched accordingly.

https://github.com/gafusion/gacode/blob/4971325b1e6ce786356cfca95d9da5848377e48a/f2py/pygacode/cgyro/data.py#L135-L145

There's also this note
```
! NOTE:
!  cflux is only of interest when profile or ExB shear is active
!  cflux(ky) is the average flux in the "positive-shear region"
!  gflux(ky,0)-cflux is the flux in the "negative-shear region"
```

So I guess cflux is the average flux on the outward side of the flux tube. I agree that we may want to use the same definition throughout. Need to see what GENE/GS2 do when there is flow shear.

d7919:
I'm not aware of GS2 doing anything differently in the diagnostics for runs with flow shear.
My preference would be for us to plot the same quantity in runs with and without flow shear, whilst perhaps flagging to the user that there's another option for CGYRO.

mgiacomin:
I don't think that a different definition of heat flux is used in GENE in presence of flow shear (at least from the documentation), but I haven't look into the source code. 
@dake0795 Are you aware of a different heat flux in GENE when `ExBrate` $\ne 0$?

dake0795:
There is definitely no other definition of the heat flux in GENE. This seems very strange!

mgiacomin:
[...]

PR #153, opened 2023-01-24
Fix/cgyro flux flow shear

mgiacomin:
This is related to issue #152 . We should use `gflux` instead of `cflux` for cross-code consistency, although `cflux` might be more appropriate in CGYRO simulations with flow shear and `SHEAR_METHOD = 2`. I left the lines commented out in case we would like to revisit this in future.


bpatel2107 (approved):
LGTM

PR #154, opened 2023-01-26
Switched out freegs._geqdsk for freeqdsk.geqdsk

LiamPattinson:
We just released [FreeQDSK 0.1](https://pypi.org/project/freeqdsk/), so this PR opts to use that to read GEQDSK files instead of FreeGS. It's the same functions from FreeGS spun off into a small utility library, so nothing in Pyrokinetics needs to be fundamentally changed to accommodate this. An immediate advantage of this is that it allows us to use the latest NumPy release (1.24), which is incompatible with the latest FreeGS release on PyPI (0.6). In the long term, it'll also be easier to develop FreeQDSK if we want to add new features.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,7 @@ classifiers = [
 requires-python = ">=3.8"
 dependencies = [
     "numpy >= 1.20.3",
+    "matplotlib >= 3.6",
     "f90nml >= 1.4.2",
     "scipy >= 1.9.3",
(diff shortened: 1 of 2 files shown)


LiamPattinson:
Lesson learned, always try installing to a fresh environment before pushing changes to `pyproject.toml`. Turns out we were using implicit dependencies via FreeGS.

bpatel2107:
LGTM

Issue #144, opened 2022-12-15
Alternative candidate for flux surface contour fitting


LiamPattinson:
I just came across [ContourPy](https://contourpy.readthedocs.io/en/v1.0.6/), which may be a better candidate than ScikitImage for fitting contours to flux surfaces. It's what Matplotlib uses as its back-end to the algorithm that we were using previously, so we could get the same results but without the associated performance hit of unnecessarily building a Matplotlib figure. Sorry for not catching this in my previous search!

LiamPattinson:
Resolved in https://github.com/pyro-kinetics/pyrokinetics/pull/155

PR #155, opened 2023-02-02
Use contourpy instead of scikit-image

LiamPattinson:
This PR removes scikit-image, which was used only for its contour fitting function, and replaces it with a dedicated contour fitting library, [contourpy](https://contourpy.readthedocs.io/en/v1.0.7/index.html). This was spun off from Matplotlib, so will be very similar to the algorithm we were using before scikit-image. Resolves Issue https://github.com/pyro-kinetics/pyrokinetics/issues/144.

## Issues

- Some of the local geometry golden answer tests had to be updated. Most changes seem fine, but both `shift` and `dZ0dr` were changed drastically for `FourierGENE`, as were `aZ` and `bR` for `FourierCGYRO`. I'm not sure if this is something to be concerned about.


bpatel2107:
This looks good! 

As for the differences, I output `pyro.local_geometry.R_eq` and `Z_eq` to a file from the original version and new version and found that the contourpy has an additional point on the flux surface

First line is R, second is Z
image

If you remove that point in the contour fitting then the original golden answers are correct. 

That additional point should be there as there is an equivalent point  (`Z=0.05`) at the beginning of the contour.

I am surprised that the `shift` and `dZ0dr` change that much but it should be fine. Fourier methods can be tricky if things are symmetric or no symmetric.

`aZ` and `bR` aren't a huge concern. It makes sense that `aZ` is smaller as the flux surface is more symmetric now and that is the asymmetric term. `bR` would also change a bit with an additional point.

I think I'm happy to merge

PR #156, opened 2023-02-06
Handle case when electron is last species

bpatel2107:
If electron is last species in `LocalSpecies` and there are more species than in the template namelist an error occurs when we try to convert to GS2. A KeyError is raised because the species namelist hasn't been added and this was only checked for ion species, not electron species.

This just checks to see if the species namelist exists and if not creates it for all species.


bpatel2107:
@jfparisi Thanks for spotting this!

PR #160, opened 2023-02-16
Enhancements/poincare map

mgiacomin:
This PR adds a new class `Diagnostics` that contains (or better will contain) all the diagnostics. Currently, it only contains the function `poincare` that can be used to generate a Poincare map. 
This PR does not modify anything else but `LocalGeometry`, where the jacobian has need added. This will change in future after merging `enhancement/metric_terms`. 
I have also added the output files of a minimal nonlinear CGYRO simulation, which is used for testing and in the example.

The poincare map is available for CGYRO, GENE and GS2 nonlinear simulations. There is a simple example (`example_poincare.py`) in the `examples` folder.

Please note that the default option is to use an interpolation of Bx(x, y) and By(x, y) while tracing magnetic field lines. I found that in some cases this is not accurate enough and one would need to take an inverse Fourier transform at each (x, y) field line coordinates, which is much slower. This option can be used by setting `use_invfft=True` when calling the poincare function.


pyrokinetics/gk_code/GKInput.py
```diff
@@ -175,6 +175,13 @@ def get_numerics(self) -> Numerics:
         """
         pass
 
+    @abstractmethod
+    def get_poincare_fac(self, dpsidr: float, rhostar: float, Lx: float) -> tuple:
```
bpatel2107: May as well call this get_poincare_factors

pyrokinetics/gk_code/GKInputCGYRO.py
```diff
@@ -491,3 +491,14 @@ def set(
             return
 
         self.data = convert_dict(self.data, local_norm.cgyro)
+
+    def get_poincare_fac(self, dpsidr: float, rhostar: float, Lx: float) -> tuple:
+        """get geometrical factors to generate Poincare map"""
+        qinp = self.data["Q"]
+        shat = self.data["S"]
+        rho = self.data["RMIN"]
```
bpatel2107: Why not take these from a `local_geometry` object. You could take all of this out of the `GKInput` and put it directly in `GKOutputReader`. You can get `Lx` from `numerics too. 

This also avoid any potential normalisation issues and `local_geometry` has a normalisation conventions along with the units

pyrokinetics/local_geometry/LocalGeometry.py
```diff
@@ -175,6 +175,7 @@ def from_global_eq(
         self.dRdtheta, self.dRdr, self.dZdtheta, self.dZdr = self.get_RZ_derivatives(
             self.theta
         )
+        self.jacob = self.R * (self.dRdr * self.dZdtheta - self.dZdr * self.dRdtheta)
```
bpatel2107, line 178: Fine for now but we will eventually be getting all the metric terms in so will update this when that get calculated

pyrokinetics/gk_code/GKOutputReader.py
```diff
@@ -100,6 +124,101 @@ def read(
...
+        # Main loop
+        j = 0
+        for x0 in xarray:
+            for y0 in yarray:
+                x = x0
+                y = y0
+                for iturn in range(nturns):
+                    for ith in range(0, ntheta - 1, 2):
+                        dby = _invfft(ikxapar[:, :, ith], x, y, Kx, Ky)
+                        dbx = _invfft(ikyapar[:, :, ith], x, y, Kx, Ky)
```
bpatel2107: xarray objects have an inverse FFT method built in so we may be able to simplify this significantly not require bespoke method here

See example below
https://github.com/pyro-kinetics/pyrokinetics/blob/7338a0f95e0e278433e55b7b99cd3140fad6c925/pyrokinetics/examples/example_real_space_moment.py#L31

pyrokinetics/gk_code/GKInput.py
mgiacomin: Done!

pyrokinetics/gk_code/GKInputCGYRO.py
```diff
@@ -491,3 +491,14 @@ def set(
...
+        dq = rhostar * Lx * shat / dpsidr
+        qmin = qinp - dq
+        fac1 = 2 * np.pi * dpsidr / rhostar
+        fac2 = dpsidr * qinp / rho
```
bpatel2107: Isn't this bunit?

bpatel2107:
Overall this looks awesome!

A couple of points

- I think you can get all the poincare factors from a `LocalGeometry` object and `Lx` from the numerics object. You can just get these directly inside the poincare method without having to modify any `GKInput` file. You can check for these object, along with if the outputs have been loaded, inside the function like here

https://github.com/pyro-kinetics/pyrokinetics/blob/072e633775f97498100a0461fc63701a8d28e653/pyrokinetics/pyro.py#L1477-L1484

- Jacobian can eventually be taken from `MetricTerms` when this gets merged which will hopefully be soon

https://github.com/pyro-kinetics/pyrokinetics/blob/enhancement/metric_terms/pyrokinetics/local_geometry/MetricTerms.py

- Using xarrays ifft would simplify this quite a bit

- In general, I don't feel like this belongs in the pyro object itself. I am thinking we should make a diagnostics suite and move it in there. This isn't going to be the only diagnostic of this style so we should try and set that up now. @ZedThree @LiamPattinson what do you think?

pyrokinetics/gk_code/GKInputCGYRO.py
mgiacomin: Yes, you are right. I did this because at the beginning I was using `bunit` for CGYRO, which is not defined in GS2 and GENE. Then I changed to make it equal for the three code. So this function is not necessary.

pyrokinetics/local_geometry/LocalGeometry.py
mgiacomin: Yes, I agree.

pyrokinetics/gk_code/GKOutputReader.py
mgiacomin: If you do the inverse transform with xarray, you need to interpolate to get the value at x and y coordinates of the magnetic field line. The interpolation might be faster, but I would need to test it.

pyrokinetics/gk_code/GKInputCGYRO.py
mgiacomin: Yes, see reply above.

mgiacomin:
Yes, as I wrote in my first comment, this is a diagnostic. I attached it as a  `pyro` object because there is nothing in place yet. I am more than happy to move everything in a new diagnostic class.

bpatel2107:
Can you add an example script? Once we put it in a bespoke diagnostics folder then we can add a plot method to generate nice figures!

mgiacomin:
I have created a new class `Diagnostics` that will contain all the diagnostics. Currently, it only contains `poincare` to generate a Poincare map. 

I still need to add tests and an example.

mgiacomin:
It should be much better now.

pyrokinetics/diagnostics/Diagnostics.py
```diff
@@ -0,0 +1,178 @@
...
+        Raises
+        ------
+        NotImplementedError: if `gk_code` is not `CGYRO`, `GENE` or `GS2`
+        RuntimeError: in case of linear simulation
+        """
+        if (
+            self.pyro.gk_code != "CGYRO"
+            and self.pyro.gk_code != "GS2"
+            and self.pyro.gk_code != "GENE"
+        ):
```
bpatel2107: A bit cleaner
```suggestion
        if (self.pyro.gk_code not in ["CGYRO", "GS2", "GENE"]):
```

pyrokinetics/diagnostics/Diagnostics.py
```diff
@@ -0,0 +1,178 @@
...
+            "kx", "ky", "theta"
+        )
+        ikyapar = ikyapar.roll(kx=shift, roll_coords=True).transpose(
+            "kx", "ky", "theta"
+        )
+        By = []
+        Bx = []
+        # Warning: Interpolation might not be accurate enough. Performing an
+        # inverse Fourier transform at every (x, y) is very accurate, but also
+        # very slow.
```
bpatel2107: Maybe we can have an kwarg that allows for the ifft for each `kx` and `ky` but have the default method use the interpolation

pyrokinetics/diagnostics/Diagnostics.py
```diff
@@ -0,0 +1,178 @@
...
+                            y = ymax - (ymin - y)
+                        if y > ymax:
+                            y = ymin + (y - ymax)
+
+                    y = y + np.mod(fac1 * ((x - xmin) / Lx * dq + qmin), Ly)
+                    if y > ymax:
+                        y = ymin + (y - ymax)
+                    points[0, j] = x
+                    points[1, j] = y
+                    j = j + 1
```
bpatel2107: Can vectorise this loop over `x` and `y` as below

```suggestion
        x = xarray[np.newaxis, :]
        y = yarray[:, np.newaxis]
        points = np.empty((2, nturns, len(yarray), len(xarray)))
        loop_length = len(xarray) * len(yarray)

        for iturn in range(nturns):
            for ith in range(0, ntheta - 1, 2):
                dby = By[ith](x, y, grid=False) * bmag[ith] * fac2
                dbx = Bx[ith](x, y, grid=False) * bmag[ith] * fac2

                xmid = x + 2 * np.pi / ntheta * dbx * jacob[ith]
                ymid = y + 2 * np.pi / ntheta * dby * jacob[ith]

                dby = By[ith+1](xmid, ymid, grid=False) * bmag[ith+1] * fac2
                dbx = Bx[ith+1](xmid, ymid, grid=False) * bmag[ith+1] * fac2

                x = x + 4 * np.pi / ntheta * dbx * jacob[ith + 1]
                y = y + 4 * np.pi / ntheta * dby * jacob[ith + 1]

                y = np.where(y < ymin, ymax - (ymin - y), y)

                y = np.where(y > ymax, ymin + (y - ymax), y)

            y = y + np.mod(fac1 * ((x - xmin) / Lx * dq + qmin), Ly)
            y = np.where(y > ymax, ymin + (y - ymax), y)

            points[0, iturn, :, :] = x
            points[1, iturn, :, :] = y

        points = np.reshape(points, (2, npoints), 'F')

```

pyrokinetics/examples/example_poincare.py
```diff
@@ -0,0 +1,26 @@
...
+# Set input for Poincare map
+xarray = np.linspace(6, 8, 5)
+yarray = np.linspace(-10, 10, 3)
+nturns = 1000
+time = 1
+rhostar = 0.036
+
+# Generate Poincare map
+diag = Diagnostics(pyro)
+coords = diag.poincare(xarray, yarray, nturns, time, rhostar)
```
bpatel2107, line 21: Is it worth keeping the starting x value for each coord so it is easy to add a colour scale for field lines starting at the same `x`?

pyrokinetics/diagnostics/Diagnostics.py
```diff
@@ -0,0 +1,178 @@
...
+        # very slow.
+        for th in apar.theta:
+            byfft = np.fft.fftshift(
+                np.fft.irfft2(ikxapar.sel(theta=th), norm="forward"), axes=0
+            )
+            bxfft = np.fft.fftshift(
+                np.fft.irfft2(ikyapar.sel(theta=th), norm="forward"), axes=0
+            )
+            By.append(RectBivariateSpline(xgrid, ygrid, byfft, kx=5, ky=5, s=1))
+            Bx.append(RectBivariateSpline(xgrid, ygrid, bxfft, kx=5, ky=5, s=1))
```
bpatel2107: Could also use xrft to do the fourier transform. I can't figure out how to reshape the data to match your answers though so its up to you if you want to put in new answers or not. Had to add in the multiplier of `nkx * ny` to do the same thing as having `norm="forward"` in the `np.irfft2`

```suggestion
        # Compute bx and by
        ikxapar = 1j * apar.kx * apar * nkx * ny
        ikyapar = -1j * apar.ky * apar * nkx * ny

        ikxapar = ikxapar.transpose(
            "kx", "ky", "theta"
        )
        ikyapar = ikyapar.transpose(
            "kx", "ky", "theta"
        )

        # Warning: Interpolation might not be accurate enough. Performing an
        # inverse Fourier transform at every (x, y) is very accurate, but also
        # very slow.
        byfft = xrft.ifft(ikxapar, dim=['kx', 'ky'], real_dim='ky', lag=[0,0], true_amplitude=False)        
        bxfft = xrft.ifft(ikyapar, dim=['kx', 'ky'], real_dim='ky', lag=[0,0], true_amplitude=False)

        By = [RectBivariateSpline(xgrid, ygrid, byfft.sel(theta=theta, method="nearest"), kx=5, ky=5, s=1) for theta in byfft.theta]
        Bx = [RectBivariateSpline(xgrid, ygrid, bxfft.sel(theta=theta, method="nearest"), kx=5, ky=5, s=1) for theta in bxfft.theta]

```

bpatel2107:
@mgiacomin left several comments

bpatel2107 (changes requested):
Made some suggestions to tidy code

pyrokinetics/examples/example_poincare.py
mgiacomin: This is very easy to do:
```ntot = nturns * yarray.shape[0]
colorlist = plt.cm.jet(np.linspace(0, 1, xarray.shape[0]))
plt.figure()
for i, color in enumerate(colorlist):
    xp = coords[0, i*ntot:(i+1)*ntot]
    yp = coords[1, i*ntot:(i+1)*ntot]
    plt.plot(xp, yp, '.', color=color)
```
I added this to the existing example.

bpatel2107: Probably cleaner to just return 3D array with dimensions [x0, y0, nturns] Essentially we would just not reshape the output after the vectorised nturns/theta loop

pyrokinetics/diagnostics/Diagnostics.py
mgiacomin: Yes, I think we need this option. There are cases in which I am really not happy with the accuracy from the interpolation. I will add this and look if using `xrft` is better.

pyrokinetics/examples/example_poincare.py
mgiacomin: Done!

pyrokinetics/diagnostics/Diagnostics.py
```diff
@@ -0,0 +1,169 @@
...
+        jacob = geo.jacob * geo.dpsidr * geo.get("bunit_over_b0", 1)
+        jacob = np.roll(jacob[::nskip], ntheta // 2)
+        dq = rhostar * Lx * geo.shat / geo.dpsidr
+        qmin = geo.q - dq / 2
+        fac1 = 2 * np.pi * geo.dpsidr / rhostar
+        fac2 = geo.dpsidr * geo.q / geo.rho
+
+        # Compute bx and by
+        ikxapar = 1j * apar.kx * apar * nkx * ny
+        ikyapar = -1j * apar.ky * apar * nkx * ny
```
bpatel2107: Should we move the nkx and ky inside the FFT as this isn't really ikxapar/ikyapar anymore?

mgiacomin: Yes. I am fixing this.

mgiacomin: This is done.

pyrokinetics/diagnostics/Diagnostics.py
mgiacomin: Now this possibility is implemented.

pyrokinetics/diagnostics/Diagnostics.py
```diff
@@ -0,0 +1,265 @@
...
+
+                    y = np.where(y < ymin, ymax - (ymin - y), y)
+
+                    y = np.where(y > ymax, ymin + (y - ymax), y)
+
+                y = y + np.mod(fac1 * ((x - xmin) / Lx * dq + qmin), Ly)
+                y = np.where(y > ymax, ymin + (y - ymax), y)
+
+                points[0, iturn, :, :] = x
+                points[1, iturn, :, :] = y
```
bpatel2107: Can this be one for loop with the if statement around setting `dbx` and `dby`? Prevents repetition of code

bpatel2107: I think they are identical otherwise

bpatel2107 (commented):
Small change suggested for removing repeated code

pyrokinetics/diagnostics/Diagnostics.py
mgiacomin: Yes, although I think the logic should remain outside the main loop for better performance.

d7919: Is it possible to define functions to get dbx/dby outside of the loop and inside the branch and you can then use these functions inside the loop, avoiding the need for repetition?  I suppose there'll be some performance hit, but hopefully not much?

````
if use_invfft:
    def get_dbx(.....)
else 
    def get_dbx(....)

for .....
   for .....
````

You'd probably want the functions to take all possible arguments and just ignore the ones it doesn't need.

bpatel2107: Yeah I see your point. Not sure how much of an impact it would have (would be roughly `2*nturns*ntheta` additional if statements to evaluate, though I'm not sure how much of an impact that has. I don't think we can vectorise it further as `x` and `y` are constantly changing.

Fine to leave as is.

bpatel2107 (approved):
Overall this looks great, happy to merge!

pyrokinetics/diagnostics/Diagnostics.py
mgiacomin: I performed various tests and there is little difference. I will push that in a moment.

PR #161, opened 2023-02-20
Adds automatic COCOS detection and conversion for equilibria

### LINKED ISSUE #159: Flipping psi_RZ grid doesn't flip q
Currently pyro forces `psi` to increase from the magnetic axis to the last closed flux surface. It flips the sign the `psi` along with derivatives w.r.t `psi`

https://github.com/pyro-kinetics/pyrokinetics/blob/ca344686b2ab4bc08d07a79c23ea96f33b7d170e/pyrokinetics/equilibrium/geqdsk.py#L83-L91

Should the sign of `q` also be changed here?

I think this is basically an issue with the COCOs not being handled properly here, referencing #124 here

LiamPattinson:
This PR attempts to resolve Issue https://github.com/pyro-kinetics/pyrokinetics/issues/159. Using the function`identify_cocos` in [OMAS](https://github.com/gafusion/omas), `Equilibrium` will try to determine what COCOS conventions the inputs adhere to. GEQDSK and TRANSP readers have been updated to provide the toroidal B field and plasma current to help determine this. Alternatively, the user may assert the convention followed by their inputs. Following this, `Equilibrium` will use `cocos_transform` to coerce its inputs into COCOS 1.

- In the process of implementing this, the files `transp_eq.cdf` and `transp_eq.geqdsk` were determined to be COCOS 5, according to the `identify_cocos` function.
- To determine the sign of the TRANSP toroidal B field, I'm using the variable `BPHI_MSE`. I have no idea if this is present in all TRANSP files, or whether there's any relationship between its sign and the actual toroidal B field. Let me know if I've misunderstood something crucial here.
- I haven't yet added tests for `Equilibrium` using different COCOS conventions for the inputs, but I can give it a try over the next few days. I recently added a test to OMAS for `identify_cocos`, so hopefully I'll be able to bring in a modified version of that.
- I'm not sure how to test GEQDSK files following different COCOS conventions without simply generating a large number of them artificially.

Let me know if you'd like to see any changes to what I've added here.


LiamPattinson:
Not sure what's happening here, as it installs and runs just fine on my own machine. Looks like the issue is with the installation of pygacode, which is brought in by OMAS. @ZedThree  Any ideas?

ZedThree:
The problem is that their build requires numpy and it's not installed first. Workaround for us is to install numpy in the CI container first. Real fix is for pygacode to list numpy as a build requirement, or to drop pygacode as an install requirement from OMAS (I don't think OMAS even uses pygacode, at least not explicitly)

bpatel2107:
I have write access to pygacode and it looks like their latest version has numpy in the `requirements.txt` file


Looks like @ZedThree beat me to it and proposed a fix.
https://github.com/gafusion/omas/issues/213

I can push these changes to pygacode, but I can't release a new version so will have to wait on the GACODE team to do so

ZedThree:
@bpatel2107 Note that it's not sufficient to have `numpy` in the `requirements.txt`, as that gets installed after the package is built. It doesn't look like OMAS actually use `pygacode` anywhere, so @LiamPattinson is going to submit a PR removing it.

bpatel2107:
There is already a PR trying to fix this, I'm just testing it out now. I think something similar to what you propose has been done

LiamPattinson:
I pushed a fix to `pyproject.toml` a little while back that directs pip towards a branch on the PlasmaFAIR fork of OMAS with updated dependencies, and it seems to work fine there. I'll leave that in place while I'm working on this PR, and will switch it back to the official version once everything is sorted out.

bpatel2107:
Ah okay I think I misunderstood! I'm currently trying to fix `pygacode` (which was the PR I was referring to earlier) but I guess that won't matter soon!

LiamPattinson:
I think the fix to pygacode will still be worthwhile, as this will likely be a problem for every other code that uses it as a dependency. In this case, we may be able to get around it by simply removing pygacode as a dependency from OMAS, but that depends on whether their team is happy to make that change.

LiamPattinson:
I've added some tests for the new COCOS detection/conversion routines, and generally improved `Equilibrium` testing while I was at it. I also moved `Equilibrium` bits to COCOS 11 as standard:

- `psi` is now in Wb instead of Wb/rad, and is therefore larger by a factor of $2\pi$
- `theta` now increases clockwise in `FluxSurface`
- These changes have not been pushed to `LocalGeometry`, so we currently need to convert back to the previous convention.

I'm currently still using a specific git commit of OMAS, so I'll have to ask them to make another release before this will be ready to merge.

Further changes:

- Improved docs for GEQDSK
- Equilibrium tests are better parametrized and use fewer 'magic numbers'
- Adapted the TRANSP CDF/GEQDSK test to compare FluxSurfaces instead of LocalGeometryMiller. I also moved the original with some minor changes to `tests/local_geometry/test_transp_cdf_vs_geqdsk.py`
- Added test for `psi_n_lcfs`
- Changed poloidal current function `f` to `F` throughout, as that seems to be more standard in the publications I've been using, but I can easily change it back if you'd prefer.
- Changed magnetic field functions to `B_radial`, `B_poloidal` etc instead of `b_radial`, `b_poloidal`.
- Reduced sensitivity on that Fourier test that only fails on some systems

pyrokinetics/equilibrium/transp.py
```diff
@@ -191,13 +203,21 @@ def read(
             RZ_coords = np.stack([x.ravel() for x in np.meshgrid(R, Z)], -1)
             psi_RZ = psi_interp(RZ_coords).reshape((nZ, nR)).T
 
+            # Get current and b_axis to determine COCOS
+            # Using 'measured plasma current', PCUR. Should this instead be one of the
+            # following?
+            # - EQ PLASMA CURRENT, PCUREQ (seems to be zero in some files?)
+            # - CALCULATED PLASMA CURRENT, PCURC
+            # - TOTAL PLASMA CURRENT, CUR (units of amps per centimeter^2)
+            I_p = data["PCUR"][time_index] * units.ampere
```
bpatel2107: In TRANSP there is "PCUR" and "PCURC". In an interpretive run these are the same, but in a predictive run when the current is calculated we would use "PCURC". Probably should switch to that to be safe

```suggestion
            I_p = data["PCURC"][time_index] * units.ampere
```

pyrokinetics/local_geometry/LocalGeometry.py
```diff
@@ -113,30 +113,26 @@ def from_global_eq(
         """
         Loads LocalGeometry object from an Equilibrium Object
         """
-        # TODO Currently stripping units from Equilibrium/FluxSurface.
+        # TODO Currently stripping units from Equilibrium/FluxSurface. These should be
+        # added in a later update.
+        # TODO FluxSurface is COCOS 11, this uses something else. Here we switch from
+        # a clockwise theta grid to a counter-clockwise one, and divide any psi
+        # quantities by 2 pi
```
bpatel2107, line 120: Why did you change the units of `psi` here? Do we need `Wb/rad` later on?

bpatel2107: Ah is this to put it in a ITER standard cocos of 11?

bpatel2107 (approved):
Overall this looks really good! Didn't realise that ITER uses COCOS=11! I think the change of `psi` units may have impacts elsewhere in #165 but happy to merge.

pyrokinetics/local_geometry/LocalGeometry.py
LiamPattinson: Yeah, it's just a COCOS 11 thing. I try to make sure that we convert to COCOS 1/2 at the point we switch to `LocalGeometry`, so hopefully it won't break anything for now. If we're aiming to use IMAS as the standard we'll have to push similar changes there too at some point.

pyrokinetics/equilibrium/transp.py
LiamPattinson: Good to know, I've made the change and edited the preceding comment.

Issue #143, opened 2022-12-07
CGYRO doesn't actually use conventional Miller


bpatel2107:
@d7919 @mgiacomin @dake0795 @hgd511 

I have just realised that CGYRO doesn't actually use conventional Miller geometry at all. 
i.e this is **not** the parameterisation
![image](https://user-images.githubusercontent.com/15210802/206260908-7ed168a8-c59a-4cd1-beb0-43d31164dbd0.png)

It **only** uses Miller extended harmonic, so the squareness `zeta` in CGYRO (below s2) is not the same as the squareness `zeta` in GENE (above).
![image](https://user-images.githubusercontent.com/15210802/206261006-9a1468cf-baa0-426f-9a31-ed75c22a5c83.png)

They happen to be the same squareness (in either code) is set to 0 and only the triangularity is included, but otherwise they are different and can't be compared.

Having only recently added squareness this is not a huge deal but we need to think how to handle this going ahead as Miller now means two things. This may requires having `MillerCGYRO` and `MillerGENE`

I have double checked both codes and they are definitely different when defining `R` and `Z

mgiacomin:
If I understand correctly, the Miller parametrisation is identical with squareness set to zero. Adding a squareness different from zero would require to treat CGYRO and GENE geometries differently, which adds another degree of complexity. I am just wondering of how much a fit can be improved by adding the squareness and if it is worth adding this extra degree of complexity just to treat the different definition of `zeta`.

bpatel2107:
The issue lies when there is an existing CGYRO/GENE file that has `zeta` already defined. If we're not careful then when converting we won't have a good cross-code comparison as the physical flux surfaces will be different. 

Also MAST-U plasmas have been quite square which has potentially impacted the pedestal stability so I think from a physics perspective it is important to capture this. 

I think we can potentially handle this under the hood by making `MillerCGYRO` and `MillerGENE` inherit from a general `Miller` object so when the user sets the code to `Miller` pyro will calculate the coefficients for you when writing the code. (could also add in a `MillerGS2` that doesn't have squareness). It's not ideal but there's not much we can do without significantly changing the GK codes themselves.

LiamPattinson:
[Link to the CGYRO geometry page, for reference.](https://gafusion.github.io/doc/geometry.html) It looks like $\zeta$ maps directly to $s_2$ in the equation above.

[...]

PR #165, opened 2023-03-06
Enhancement/metric terms

hgd511:
Calculates the local Grad-Shafranov solution using local geometry, which is used to calculate the covariant and contravariant metric tensor components for the toroidal system and the field-aligned system.

The additional derivatives in R and Z necessary for these computations have been included in the relevant LocalGeometry files. 

Because the equivalent of ff_prime can be calculated in this formalism independently of its measured value, a comparison between and the measurement and the calculation of ff_prime can provide a measure for the quality of the geometric parameterisation.

Attached is a file containing basic documentation of the different quantities that can be calculated using MetricTerms. Please advise regarding the best place for this.
[PyroMetricTerms.pdf](https://github.com/pyro-kinetics/pyrokinetics/files/10897087/PyroMetricTerms.pdf)


pyrokinetics/local_geometry/LocalGeometry.py
```diff
@@ -525,19 +529,21 @@ def plot_equilibrium_to_local_geometry_fit(
...
         -------
         Copy of LocalGeometry object
-        """
+        ""
         # Create new empty object. Works for derived classes too.
         new_localgeometry = self.__class__()
         for key, value in self.items():
             new_localgeometry[key] = deepcopy(value, memodict)
         return new_localgeometry
+    """
```
bpatel2107: Can probably delete this section

bpatel2107:
Overall this looks good!

- We should add some tests for some basic circular plasmas where the metric terms have a simple geometric definitions.
- @ZedThree @LiamPattinson should we put that PDF in docs or is there a nice way to embed in into the readthedocs?
- The tests are currently failing due to the different definitions of "Miller" that codes have (see #143) . When only including elongation and triangularity, the codes are all the same but when including squareness, the codes diverge so we should find a way to handle this. The first thing I can thing of is to have a `BasicMiller` geometry that doesn't have squareness that we can use for cross-code comparisons/roundtrips, but then have the option to use squareness in something like `FullMiller` geometry

ZedThree:
Well, seeing as @hgd511 wrote the PDF, we can just convert the LaTeX to reStructuredText and include it straight in the docs, keeping all the equations intact :slightly_smiling_face: 

@hgd511 The easiest thing perhaps is to use [pandoc](https://pandoc.org) to convert your source file to rst, and then add it to [docs/rst_docs/local_geometry.rst](https://github.com/pyro-kinetics/pyrokinetics/blob/unstable/docs/rst_docs/local_geometry.rst)

bpatel2107:
@ZedThree @LiamPattinson This PR looks good but the failing tests requires quite a big refactor to fix. Is it worth merging this into unstable and temporarily not run those tests, or open a separate PR to fix that, merge that into this and then merge everything into unstable

ZedThree:
I would definitely get the tests fixed before merging!

bpatel2107:
Big ol refactor coming in then!

LiamPattinson:
As easy(-ish) fix might be to catch those `NotImplementedError`s in a `try...except` block, and if the provided `LocalGeometry` type doesn't match the target GK code, it should raise a warning and instead fit a compatible `LocalGeometry`. So rather than saying "GS2 doesn't support MXH" and failing, it could instead just fit a `LocalGeometryMiller` as best as it can.

bpatel2107 (approved):
LGTM

PR #167, opened 2023-03-09
Handle different types of geometries

### LINKED ISSUE #143: CGYRO doesn't actually use conventional Miller

bpatel2107:
There are different definitions of `Miller` that codes use. What I have done is to re-define the `Miller` class to fit the geometry exactly how is done in the Miller 1998 paper (Phys. Plasmas, Vol. 5, No. 4, April 1998 Miller et al). All GK codes are able to use this fitting and is now the default `LocalGeometry`. This geometry only goes up to triangularity and has no squareness

An expansion to Miller was done in Physics of Plasmas 6, 1113 (1999); Turnbull et al ;  https://doi.org/10.1063/1.873380, which include squareness (which was included in the function for `Z`). This is now `LocalGeometryMillerTurnbull` and the only code that can use this is GENE.

Finally a modified Miller geometry called Miller eXtended Harmonic (MXH) was developed (Plasma Phys. Control. Fusion 63 (2021) 012001 (5pp) https://doi.org/10.1088/1361-6587/abc63b) which also has squareness (in the function for `R`) along higher moments which is available for CGYRO and TGLF.

Any exact cross-code comparison can only be done with `LocalGeometryMiller`. If had a GENE input with `LocalGeometryMillerTurnbull` you would have to re-fit the equilibrium using a different geometry so doing a 1-to-1 comparison is not possible but you can probably get close.

The default behaviour is to load `LocalGeometryMiller` (without squareness) which does change the default fits. The main issue was that squareness as it was before in CGYRO was incorrect so simulation where CGYRO and GENE are compared aren't quite perfect matches. I imagine the effect was near negligible but @dake0795 @mgiacomin be aware. Simulations without squareness are fine.

This fixes #143 and is needed for #165


pyrokinetics/examples/example_geqdsk.py
```diff
@@ -20,6 +20,8 @@
 pyro.load_local_geometry(psi_n=0.5, local_geometry="Miller", show_fit=show_fit)
 
 # Switch to different geometry types and plot fit
+pyro.switch_local_geometry(local_geometry="BasicMiller", show_fit=show_fit)
```
d7919: Reading the PR comment I would expect BasicMiller==Miller (i.e. the default Miller case is the simple one). I think this points out that the PR is currently missing documentation describing this option.

pyrokinetics/gk_code/GKInputCGYRO.py
```diff
@@ -36,23 +38,71 @@ class GKInputCGYRO(GKInput):
...
+        "dsndr3": 0.0,
+    }
+
+    pyro_cgyro_miller_defaults = {
+        "rho": 0.5,
+        "Rmaj": 3.0,
+        "q": 2.0,
+        "kappa": 1.0,
+        "s_kappa": 0.0,
+        "delta": 0.0,
```
d7919: no s_delta?

d7919: Also given the mxh ones expand this set can you remove the duplication of rho, rmaj etc defaults by defining the base first and then making mxh start with a copy of this which just adds the extra values?

pyrokinetics/gk_code/GKInputCGYRO.py
```diff
@@ -184,8 +233,15 @@ def get_local_geometry(self) -> LocalGeometry:
         Returns local geometry. Delegates to more specific functions
         """
         eq_type = self.cgyro_eq_types[self.data["EQUILIBRIUM_MODEL"]]
+
+        is_basic_miller = self._check_basic_miller()
+        if eq_type == "MXH" and is_basic_miller:
+            eq_type = "Miller"
```
d7919, line 209: Warn to the user that the type is being reduced (i.e. information lost)?

pyrokinetics/examples/example_geqdsk.py
bpatel2107: Ah that is the wrong name, should be "MillerTurnbull" here. Originally I had "BasicMiller" as the 1998 paper version but I thought that wasn't a good name.

pyrokinetics/gk_code/GKInputCGYRO.py
```diff
@@ -230,14 +287,52 @@ def get_local_geometry_miller(self) -> LocalGeometryMiller:
...
+
     def get_local_geometry_fourier(self) -> LocalGeometryFourierCGYRO:
         """
         Load Fourier object from CGYRO file
         """
         fourier_data = default_fourier_cgyro_inputs()
 
         for (key, val), val_default in zip(
-            self.pyro_cgyro_fourier.items(), self.pyro_cgyro_miller_defaults.values()
+            self.pyro_cgyro_fourier.items(), self.pyro_cgyro_fourier_defaults.values()
```
d7919, line 308: Is this a separate bug fix?

pyrokinetics/gk_code/GKInputCGYRO.py
```diff
@@ -407,6 +504,10 @@ def set(
 
         eq_type = self.cgyro_eq_types[eq_model]
 
+        is_basic_miller = self._check_basic_miller()
+        if eq_type == "MXH" and is_basic_miller:
+            eq_type = "Miller"
```
d7919: Warn?

pyrokinetics/gk_code/GKInputCGYRO.py
bpatel2107: There is but its not a one to one mapping from the gk_code to the pyro definition. (CGYRO is different to the Miller `s_delta` by a factor `1/sqrt(1-delta^2)`, which is what this list is for so you have to manually set it. see below

https://github.com/pyro-kinetics/pyrokinetics/blob/bcc6843e06fc6cb08c441b907dded6d6e6754986/pyrokinetics/gk_code/GKInputCGYRO.py#L264-L266

pyrokinetics/gk_code/GKInputCGYRO.py
bpatel2107: Its not that the information is being lost, but that the squareness and higher moments are all 0 so it doesn't exist to begin with. Essentially, CGYRO without those values is the 1998 Miller so we re-define it as such.

pyrokinetics/gk_code/GKInputCGYRO.py
bpatel2107: Yes

pyrokinetics/gk_code/GKInputCGYRO.py
bpatel2107: See comment above

d7919:
Looks sensible and can't see any obvious errors/issues (although a few comments raised). I think you might want some more documentation around this and perhaps some more flagging to users when data might be lost (e.g. in conversion from CGYRO->GENE etc.). I guess some more tests around switching/comparing between Miller types for some known cases might help as well.

pyrokinetics/local_geometry/LocalGeometryMXH.py
```diff
@@ -190,6 +189,19 @@ def from_local_geometry(
...
         )
 
+    @classmethod
+    def from_gk_data(cls, params: Dict[str, Any], n_moments: Optional[int] = 4):
+        """
+        Initialise from data gathered from GKCode object, and additionally set
+        bunit_over_b0
+        """
+
+        cls.n_moments = n_moments
```
LiamPattinson: This has the potential to break things. If two `LocalGeometryMXH`'s are created in the same Python session, and they each set `n_moments` to something different, they'll both share the value assigned to whichever was created last, as the attribute is set on the class rather than an instance of it. As `n_moments` needs to passed in before the call to `from_gk_data`, the only solution I can think of is to update `from_gk_data` to take optional `**kwargs` and to pass in `n_moments` via that.

(Alternatively, if we're confident that `n_moments == 4` everywhere, it could just be removed as an optional argument)

bpatel2107: CGYRO/TGLF can only use 4 moments so there's actually no need to have this as an option. We should force `n_moments=4` for all `MXH` geometries

bpatel2107:
@LiamPattinson @ZedThree do you know why the workflows haven't run on the latest commit? Is it to do with the automated black changes not triggering the workflow?

LiamPattinson:
I've run into a similar problem before, although it was for a PR I was making for someone else's library from my own fork, and the branch protections they'd implemented disallowed running tests in these cases. It'd be nice if GitHub could report exactly what "Waiting for status to be reported" means, as it seems there are many different ways to arrive at the same error message.

Unfortunately the best I can offer is a [Stack Overflow answer](https://stackoverflow.com/questions/52200096/github-pull-request-waiting-for-status-to-be-reported), where the general advice is to make a new empty commit or close/reopen the PR and hope for the best. If it becomes a recurring issue, then it's likely something to do with branch protections and secret tokens that I don't fully understand.

bpatel2107:
It looked like the empty commit only ran the pytest tests, and not the black/flake8 tests. I tweaked those two to run on a PR as well and it seems like that triggered the job.

It also seems like changing the rules undid the approval. May be the case that each commit undoes the approval though I can't see that setting in the branch protection

pyrokinetics/gk_code/GKInputCGYRO.py
```diff
@@ -491,3 +577,35 @@ def set(
...
+            "SHAPE_S_COS1",
+            "SHAPE_S_COS2",
+            "SHAPE_S_COS3",
+            "SHAPE_S_SIN3",
+        ]
+
+        is_basic_miller = True
+        for param in mxh_only_parameters:
+            if self.data.get(param, 0.0) != 0:
+                is_basic_miller = False
```
LiamPattinson, line 610: A very minor point: Once `is_basic_miller` is set to false, there's no need to continue in the for loop so we may as well `break`, or instead could just `return false`.

LiamPattinson (commented):
Looking through the code, the only issue I could spot by eye is extremely nitpicky, so overall looks good.

Unfortunately, I ran into other issues when running the code myself, but this might be due to improper usage. I'm getting errors when trying to convert from one `LocalGeometry` type to another:
```python
>>> import pyrokinetics as pk
>>> pyro = pk.Pyro(gk_file=pk.gk_templates["GS2"])
>>> pyro.switch_local_geometry("MXH")
>>> pyro.write_gk_file("input.cgyro", gk_code="CGYRO")
  File "<stdin>", line 1, in <module>
  File "/home/ltp511/workspace/pyrokinetics/pyrokinetics/pyro.py", line 917, in write_gk_file
    self._switch_gk_context(
  File "/home/ltp511/workspace/pyrokinetics/pyrokinetics/pyro.py", line 438, in _switch_gk_context
    self.update_gk_code()
  File "/home/ltp511/workspace/pyrokinetics/pyrokinetics/pyro.py", line 497, in update_gk_code
    self.gk_input.set(
  File "/home/ltp511/workspace/pyrokinetics/pyrokinetics/gk_code/GKInputCGYRO.py", line 490, in set
    self.data[val] = local_geometry[key]
  File "/home/ltp511/.pyvenv/pyrokinetics/lib/python3.8/site-packages/cleverdict/cleverdict.py", line 435, in __getitem__
    name = self.get_key(name)
  File "/home/ltp511/.pyvenv/pyrokinetics/lib/python3.8/site-packages/cleverdict/cleverdict.py", line 543, in get_key
    raise KeyError(name)
KeyError: 'delta'
>>> pyro = pk.Pyro(gk_file=pk.gk_templates["GS2"])
>>> pyro.switch_local_geometry("FourierCGYRO")
>>> pyro.write_gk_file("input.cgyro", gk_code="CGYRO")
  File "<stdin>", line 1, in <module>
  File "/home/ltp511/workspace/pyrokinetics/pyrokinetics/pyro.py", line 917, in write_gk_file
    self._switch_gk_context(
  File "/home/ltp511/workspace/pyrokinetics/pyrokinetics/pyro.py", line 438, in _switch_gk_context
    self.update_gk_code()
  File "/home/ltp511/workspace/pyrokinetics/pyrokinetics/pyro.py", line 497, in update_gk_code
    self.gk_input.set(
  File "/home/ltp511/workspace/pyrokinetics/pyrokinetics/gk_code/GKInputCGYRO.py", line 501, in set
    self.data[val] = local_geometry[key]
  File "/home/ltp511/.pyvenv/pyrokinetics/lib/python3.8/site-packages/cleverdict/cleverdict.py", line 435, in __getitem__
    name = self.get_key(name)
  File "/home/ltp511/.pyvenv/pyrokinetics/lib/python3.8/site-packages/cleverdict/cleverdict.py", line 543, in get_key
    raise KeyError(name)
KeyError: 'kappa'
```
I get similar results if I read in the CGYRO template, convert to MXH, and try to write out again as another CGYRO file. On the other hand, the following works:

```python
>>> pyro = pk.Pyro(gk_file=pk.gk_templates["GS2"])
>>> pyro.switch_local_geometry("MillerTurnbull")
>>> pyro.write_gk_file("input.gene", gk_code="GENE")
```

I can't immediately see what could be causing this. Any ideas?

bpatel2107:
Ah good spot, that is a subtle issue. So when reading a geometry file we need to check the input file to see if it is a `Miller` or `MXH`. But when writing we need to check which type the `LocalGeometry` is object instead.

Also `FourierCGYRO` has not been set up yet so I've added a `NotImplementedError`

LiamPattinson (approved):
Those changes seem to have fixed it, and otherwise all looks good to me.

PR #168, opened 2023-03-09
Add command line interface

LiamPattinson:
In a recent chat, it was mentioned that a Command Line Interface (CLI) for basic operations would be helpful, and that it would serve as a useful first step towards a GUI. I've added a system using [`argparse`](https://docs.python.org/3/library/argparse.html) that allows users to convert GK input files on the command line, and they can optionally pull in equilibrium/kinetics/template data too.

As Pyrokinetics is capable of doing a bunch of different things, I designed this to be extendable using 'subparsers'. Currently, it only has one subcommand `convert`, but I could foresee it having subcommands such as `run`, `scan`, `compare`, `analyse`, etc in future.

Let me know if there's anything here you'd like me to change.

## Usage

You can run the CLI via two methods:

- `$ python3 -m pyrokinetics`, which uses the file `__main__.py`.
- `$ pyro`, which uses script entrypoints in `pyproject.toml` to call `pyrokinetics.cli.entrypoint:entrypoint` directly.

Both point to the same entrypoint function, so there's no practical difference in usage. If you only call `$ pyro`, it'll print a help message:

```
usage: Pyrokinetics [-h] {convert} ...

A tool to run and analyse gyrokinetics simulations.

optional arguments:
  -h, --help  show this help message and exit

subcommands:
  Please provide one of the following subcommands as a positional argument. For information on how each subcommand works, try providing '--help' after the subcommand.

  {convert}
    convert   Convert a gyrokinetics input file to a different code.
```

You can also get this by calling `$ pyro --help` or `$ pyro -h`. The `convert` subcommand also provides help text if you call `$ pyro convert --help`:

```
usage: Pyrokinetics convert [-h] [--input_type INPUT_TYPE] [--equilibrium EQUILIBRIUM] [--equilibrium_type EQUILIBRIUM_TYPE] [--kinetics KINETICS] [--kinetics_type KINETICS_TYPE] [--psi PSI] [--a_minor A_MINOR] [--output OUTPUT] [--template TEMPLATE] target input_file

positional arguments:
  target                The target gyrokinetics code, e.g. 'GS2', 'CGYRO'.
  input_file            The gyrokinetics config file you wish to convert.

optional arguments:
  -h, --help            show this help message and exit
  --input_type INPUT_TYPE
                        The code type of the input file. If not provided, this will be inferred.
  --equilibrium EQUILIBRIUM, --eq EQUILIBRIUM, -e EQUILIBRIUM
                        Path to a plasma equilibrium file, which is used to overwrite the flux surface in 'input_file'. Users should also provide 'psi' to select which flux surface to use from the equilibrium.
  --equilibrium_type EQUILIBRIUM_TYPE, --eq_type EQUILIBRIUM_TYPE
                        The type of equilibrium file. If not provided, this is inferred.
  --kinetics KINETICS, -k KINETICS
                        Path to a plasma kinetics file, which is used to overwrite the local species data in 'input_file'. Users should also provide 'psi' and 'a_minor' to select which flux surface to use, or provide 'psi' and 'equilibrium'.
  --kinetics_type KINETICS_TYPE, --k_type KINETICS_TYPE
                        The type of kinetics file. If not provided, this is inferred.
  --psi PSI, -p PSI     The normalised poloidal flux function, used to index which flux surface to draw equilibrium/kinetics data from. Should be in the range [0,1], with 0 being the magnetic axis, and 1 being the last closed flux surface.
  --a_minor A_MINOR, -a A_MINOR
                        The width of the last closed flux surface, in meters. Used to select a flux surface when providing kinetics data but no equilibrium. Otherwise, this argument is ignored.
  --output OUTPUT, -o OUTPUT
                        Name of the new gyrokinetics config file.
  --template TEMPLATE, -t TEMPLATE
                        Template file to use for the new gyrokinetics config file.
```

The following will convert a GS2 file to CGYRO:

```
$ pyro convert CGYRO my_gs2_file.in --output input.cgyro
```

You can also bring in equilibrium/kinetics stuff, but the command line gets a bit crowded if you do:

```
$ pyro convert CGYRO my_gs2_file.in --equilibrium my_eq.geqdsk --kinetics my_kinetics.cdf --psi=0.95 --output input.cgyro
```

Most optional arguments also have shortened aliases:

```
$ pyro convert CGYRO my_gs2_file.in -e my_eq.geqdsk -k my_kinetics.cdf -p=0.95 -o input.cgyro
```

## Issues

- Haven't written tests yet, and CLI stuff is quite difficult to test anyway. @ZedThree would you recommend just testing the `main()` function I've written and leaving the other pieces untested? Or should I test the whole thing using `subprocess`?
- Currently don't have a way to specify the preferred `LocalGeometry` type, but as that whole system is in flux I think it'll be fine to keep this restricted to Miller for now.

### DIFF
--- a/pyrokinetics/cli/convert.py
+++ b/pyrokinetics/cli/convert.py
@@ -11,7 +11,12 @@ def add_arguments(parser: ArgumentParser) -> None:
     parser.add_argument(
         "target",
         type=str,
-        help="The target gyrokinetics code, e.g. 'GS2', 'CGYRO'.",
+        help=dedent(
+            f"""\
+            The target gyrokinetics code. Options include
+            {', '.join(Pyro().supported_gk_inputs)}.
+            """
+        ),
     )
 
     parser.add_argument(
@@ -32,8 +37,8 @@ def add_arguments(parser: ArgumentParser) -> None:
         type=str,
         help=dedent(
             """\
-            The type of flux surface geometry to convert to. Options include 'Miller' or
-            'Fourier'.
+            The type of flux surface geometry to convert to. Options currently include
+            Miller (all), MillerTurnbull (GENE) and MXH (CGYRO, TGLF).
             """
         ),
     )
@@ -44,10 +49,11 @@ def add_arguments(parser: ArgumentParser) -> None:
         "-e",
         type=Path,
         help=dedent(
-            """\
+            f"""\
             Path to a plasma equilibrium file, which is used to overwrite the flux
             surface in 'input_file'. Users should also provide 'psi' to select which
-            flux surface to use from the equilibrium.
+            flux surface to use from the equilibrium. The supported equilibrium types
+            are {', '.join(Pyro().supported_equilibrium_types)}.
             """
         ),
     )
@@ -64,10 +70,11 @@ def add_arguments(parser: ArgumentParser) -> None:
         "-k",
         type=Path,
         help=dedent(
-            """\
+            f"""\
             Path to a plasma kinetics file, which is used to overwrite the local species
             data in 'input_file'. Users should also provide 'psi' and 'a_minor' to
-            select which flux surface to use, or provide 'psi' and 'equilibrium'.
+            select which flux surface to use, or provide 'psi' and 'equilibrium'. The
+            supported kinetcs types are {', '.join(Pyro().supported_kinetics_types)}.
             """
         ),
     )
--- a/pyrokinetics/cli/convert.py
+++ b/pyrokinetics/cli/convert.py
@@ -1,0 +1,146 @@
+from argparse import ArgumentParser, Namespace
+from pathlib import Path
+from textwrap import dedent
+
+from pyrokinetics import Pyro
+
+description = "Convert a gyrokinetics input file to a different code."
+
+
+def add_arguments(parser: ArgumentParser) -> None:
+    parser.add_argument(
+        "target",
+        type=str,
+        help="The target gyrokinetics code, e.g. 'GS2', 'CGYRO'.",
+    )
+
+    parser.add_argument(
+        "input_file",
+        type=Path,
+        help="The gyrokinetics config file you wish to convert.",
+    )
+
+    parser.add_argument(
+        "--input_type",
+        type=str,
+        help="The code type of the input file. If not provided, this will be inferred.",
+    )
+
+    parser.add_argument(
+        "--equilibrium",
+        "--eq",
+        "-e",
+        type=Path,
+        help=dedent(
+            """\
+            Path to a plasma equilibrium file, which is used to overwrite the flux
+            surface in 'input_file'. Users should also provide 'psi' to select which
+            flux surface to use from the equilibrium.
+            """
+        ),
+    )
+
+    parser.add_argument(
+        "--equilibrium_type",
+        "--eq_type",
+        type=str,
+        help="The type of equilibrium file. If not provided, this is inferred.",
+    )
+
+    parser.add_argument(
+        "--kinetics",
+        "-k",
+        type=Path,
+        help=dedent(
+            """\
+            Path to a plasma kinetics file, which is used to overwrite the local species
+            data in 'input_file'. Users should also provide 'psi' and 'a_minor' to
+            select which flux surface to use, or provide 'psi' and 'equilibrium'.
+            """
+        ),
+    )
+
+    parser.add_argument(
+        "--kinetics_type",
+        "--k_type",
+        type=str,
+        help="The type of kinetics file. If not provided, this is inferred.",
+    )
+
+    parser.add_argument(
+        "--psi",
+        "-p",
+        type=float,
+        help=dedent(
+            """\
+            The normalised poloidal flux function, used to index which flux surface to
+            draw equilibrium/kinetics data from. Should be in the range [0,1], with 0
+            being the magnetic axis, and 1 being the last closed flux surface.
+            """
+        ),
+    )
+
+    parser.add_argument(
+        "--a_minor",
+        "-a",
+        type=float,
+        help=dedent(
+            """\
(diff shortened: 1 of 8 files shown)


ZedThree:
You can monkeypatch `argv` and then run the CLI in the unit tests, without the expense of using `subprocess`. See how we do it in Ford: https://github.com/Fortran-FOSS-Programmers/ford/blob/abe9451997b0920d5338f9129741dae7ccbade7d/test/test_example.py#L24-L27

```py
    with pytest.MonkeyPatch.context() as m:
        os.chdir(tmp_path)
        m.setattr(sys, "argv", ["ford", "example-project-file.md"])
        ford.run()
```

bpatel2107:
I think #167 will mean that specifying the input and output geometries are essential! If the input isn't a standard `LocalGeometryMiller` (what will be the new one without squareness) then it won't be possible to directly convert between codes.

Might be worth waiting until that is merged, (which is ready to go!). The actual conversion is straight forward using `pyro.switch_local_geometry`. You would probably only need to specify the output geometry as this method doesn't need the input one.

LiamPattinson:
@ZedThree Cheers for that, I'll take a look at how Ford handles things.

@bpatel2107 Good to know, I'll be sure to get a geometry option in there before finishing up. It should as simple as adding an argument `--geometry`/`-g` that takes something like 'Miller' or 'MXH' or 'Fourier' etc.

bpatel2107:
Would also be good to have some basic plotting scripts for eigenfunctions, fluxes and growth rates, something like a `pyro_plot` script

LiamPattinson:
I've added a test for the CLI 'convert' function. I first tried testing over all combinations of `gk_input`, `gk_output`, using equilibrium, using kinetics, switching geometry, using long option names, and explicitly listing types, but this resulted in over 500 individual tests and took about 10 minutes to run. I settled on 8 tests instead. Any ideas for how I could rewrite things to test a wider range of inputs/outputs without having the tests take forever?

I also tried rebasing the branch, and to avoid a messy merge conflict I figured I could get away with a force push. This seems to have broken the black CI (lesson learned, let's see how long it sticks this time). Should I close this PR and start a new one?

bpatel2107:
Not sure about black. Maybe mark as ready for review and push an empty commit?

500 tests would be a tad excessive. I think as long as each component is covered by the test we should be fine, so maybe for each different option you can try for all the given options, i.e. try all the different equilibrium types, but don't necessarily run that for each gk code/kinetics type etc. So not an outer product.

LiamPattinson:
I've rewritten the tests to cover all combinations of inputs, and the tests for equilibria, kinetics, etc only cover a single case. I'd like to leave this PR here, as now that the framework is ready it should be relatively easy for others to extend it to include additional subcommands. I agree that a plotting tool would be helpful, but I think it'd be best to leave to somebody who actually does gyrokinetics analysis and knows the best way to present that data.

LiamPattinson:
(Please don't merge until the black thing is sorted, I don't know if it'll break things for the whole project)

LiamPattinson:
Just going to try closing and reopening to see if that helps...

LiamPattinson:
@ZedThree I appear to have broken things quite thoroughly. Any ideas?

ZedThree:
I think it was most likely this commit: d2ebc51e85e45c14994ef087b3fcc160f50fad26
It runs `black` on PRs, which isn't necessary -- here it's trying to pull this branch from `origin`, i.e. this repo.

So `black.yml` should have `on: pull_request` removed. Might also be worth you pushing directly to this repo instead of a branch?

bpatel2107:
I found that without that, black wouldn't run again after doing its automatic fix so I wasn't able to merge as the latest commit needs to pass the black test before merging

LiamPattinson:
Whoops, only just read your last comment. Reverted my change.

bpatel2107:
There may be a better way than what I did so happy to change if something else is preferable

LiamPattinson:
Okay, I've now pushed it to pyrokinetics as well as PlasmaFAIR, and that seems to have fixed it. I guess I'll have to stop working from a fork in future. Thanks @ZedThree, and sorry for the notification spam.

pyrokinetics/cli/convert.py
```diff
@@ -0,0 +1,162 @@
...
+    )
+
+    parser.add_argument(
+        "--geometry",
+        "-g",
+        type=str,
+        help=dedent(
+            """\
+            The type of flux surface geometry to convert to. Options include 'Miller' or
+            'Fourier'.
```
bpatel2107: Technically there are a lot more now though I'm not sure if this is best place to put it but we have the following that are fully working
Miller (all)
MillerTurnbull (gene)
MXH (cgyro, tglf)

The fourier ones haven't been implemented in the `gk_code`, we can do the fit but haven't mapped it to the codes yet.

bpatel2107 (approved):
LGTM.

Only suggestion would be to list all the potential options for each type in the description. Maybe we can use `pyro.supported_gk_inputs` for example to list out the available `gk_code`?

pyrokinetics/cli/convert.py
LiamPattinson: Good catch, I'd previously implemented it to interpret 'Miller' and 'Fourier' differently depending on the target code, but since both CGYRO and GENE can handle plain Miller as well as their specialised variants, I couldn't figure out a clean way to implement it and fell back on just letting the user specify whatever they wanted. Looks like I forgot to rewrite the message to match.

bpatel2107 (approved):

ZedThree:
> I found that without that, black wouldn't run again after doing its automatic fix so I wasn't able to merge as the latest commit needs to pass the black test before merging

Ok, another option might be to run the auto-black CI on `push` to `unstable` only. Then it won't stop PRs, but will always get applied

PR #169, opened 2023-03-09
Relax matplotlib contraint to >=3.3.4 (for OMFIT compatibility)

bpatel2107:
OMFIT requires matplotlib up to 3.3.4 and I see no reason for us to _need_ matplotlib>3.6 so this will let us be compatible with them.

I think the amount of work to get OMFIT compatible with the latest versions of matplotlib would probably be significant so this seems easier for now

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
 requires-python = ">=3.8"
 dependencies = [
     "numpy >= 1.20.3",
-    "matplotlib >= 3.6",
+    "matplotlib >= 3.3.4",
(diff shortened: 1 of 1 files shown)


bpatel2107:
@jfparisi  FYI

LiamPattinson:
I recall having to change the minimum version at some point in the past, but I can't remember why (maybe something to do with flux surface fitting?). I think most of the minimum versions we've chosen are arbitrary, so it should be fine to tinker with them a bit provided the tests still pass while using those minimum versions. (I just checked, and 3.3.4 runs fine)

PR #172, opened 2023-03-10
Bugfix/nspecies gene

bpatel2107:
When writing a GENE file, a template namelist is used with only 2 species. For simulations with more species, we would need to create a new namelist for each species. Currently we are only creating a new namelists for ion species, so if the electron species is the 3rd or higher in order, then a new namelist wouldn't be created and an `IndexError` would occur

Here we move the check for the namelist earlier

### DIFF
--- a/pyrokinetics/gk_code/GKInputGENE.py
+++ b/pyrokinetics/gk_code/GKInputGENE.py
@@ -422,7 +422,7 @@ class GKInputGENE(GKInput):
             for key, val in self.pyro_gene_species.items():
                 single_species[val] = local_species[name][key]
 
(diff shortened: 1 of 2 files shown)


bpatel2107:
Ngl I thought we spotted this earlier and fixed it...

ZedThree:
Can you add a test for this behaviour?

bpatel2107:
Added a test in switching the order of species when loading from a JETTO run, so the electron species is last. This failed previously but doesn't with the fix

ZedThree (approved):

ZedThree:
Nice! LGTM

PR #173, opened 2023-03-14
standardize input scripts under templates

arkabokshi:
Added `gkw` input script and updated inputs for `gs2` and `gene` so that they all agree for a physically relevant standardized Cyclone-like case. 

`cygro` and `tglf` inputs need updating... @mgiacomin @bpatel2107 ?


mgiacomin (commented):
Why do you need `beta_prime=0`?

pyrokinetics/templates/input.gs2
```diff
@@ -63,7 +66,7 @@
 /
 
 &collisions_knobs
-    collision_model = 'default'
+    collision_model = 'none'
```
mgiacomin: Why do you need a collisionless case?

mgiacomin: @arkabokshi in case you don't receive notifications.

mgiacomin:
I was wondering if this PR should be part of a more general PR where GKW is added?

pyrokinetics/templates/input.gs2
bpatel2107: We should try and match the collision operators as close as possible and I think including it is best. pyro doesn't currently change this value yet so we should leave it on by default

d7919: Yes, I'd be inclined to say we should match everything as closely as possible whilst enabling as much physics as possible.

pyrokinetics/templates/input.gs2
```diff
@@ -75,8 +78,8 @@
     mass = 2.724e-4
     dens = 1.0
     temp = 1.0
-    tprim = 3.0
-    fprim = 1.0
+    tprim = 9.0
```
bpatel2107: We need to be careful here and pyro doesn't quite deal with using `Rmaj` as the reference length yet. It should be doable given our normalisation setup.

bpatel2107:
I think it makes sense for all the templates to be the same case so happy for that to be a separate PR. We don't really need the GKW input file until GKW is ready to be merged but I'm fine to keep it in there for now.

pyrokinetics/templates/input.gs2
arkabokshi: OK. If we are inclined to keep as much physics as possible then we should enable collisions and beta_prime.  I think beta_prime makes it trickier to match; let me try again.

bpatel2107: Just use pyro to convert it

ZedThree:
Here's my thoughts: having the default template be cyclone seems sensible, it's good, familiar starting point for most purposes. But it also makes sense to supply some templates for more complex cases, like collisional, EM, nonlinear, etc. Having each of these templates for each code is also sensible.

pyrokinetics/templates/input.gs2
arkabokshi: So in these templates, does it matter what the normalizing length-scale is?   I was thinking that the templates should work as is if we ran them, so while I can re-normalize everything to `a_minor`, `GKW` won't like it (`Rmaj` normalization is hardwired)...

bpatel2107:
Might be worth making a CBC directory for those templates and a separate directory for the complicated full geometry, EM, collisions case.

Would be good because we then have the option of eventually running these simple cases to see if the results are as expected (definitely a different PR and not sure how we would do that)

pyrokinetics/templates/input.gs2
bpatel2107: Sure but `TGLF` and `CGYRO` (and `GFTM` when that gets added in) have `a_minor` hardwired as the normalising length case. Given that more codes have `a_minor` I'm inclined to use that in the templates. I do accept that the `GKW` should be physically correct so to have `Rmaj`, such that if we ran we it would be physically the same as the others. 

This is a slightly fiddily I admit

arkabokshi: 1. Should we keep `GKW` in `Rmaj` and `TGLF` & `CYGRO` in `a_minor` then?  Should be fine as long as the templates can produce the same result?
2. Following on from @ZedThree's suggestion, do you want the templates to be linear for now?  A more comprehensive template (perhaps covering nonlinear settings) could be a separate PR?

d7919:
It's worth remembering that the templates are _not_ examples -- they could be, but their first job is to provide sensible defaults I guess.

bpatel2107:
I've updated the TGLF and CGYRO input files to the CBC. I've set `beta_prime` consistent with the gradients and assumed that the GS2 electron collisionality (`vnewk`) is 0.1

Not sure if there is a default collisionality for CBC

arkabokshi:
Have you tried running the two codes to see that they agree?   I can get pretty good agreement between `GKW` and `GS2`, but `GENE` does not agree with them when `beta_prime` is enabled (without $\beta'$ all 3 codes in excellent agreement).  

Tried updating the `GENE` script by hand and using Pyro and the outcome is no better (both wrong in different ways).  I suspect there is another bug in the `GENE` part of the code -- will open another PR.  Maurizio and I will have a check later today.  Could @dake0795 have a look as well?

mgiacomin:
The tests should also be updated since we changed the template files.

pyrokinetics/templates/input.gene
```diff
@@ -59,7 +59,7 @@ istep_schpt = 0
     shat = 1.0
     major_R = 1.0
     trpeps = 0.166
-    minor_r = 0.0
+    minor_r = 1.0
```
bpatel2107, line 68: In the CBC, I believe `minor_r` should be `1/3` if using `Lref = R_major`

bpatel2107:
So I managed to get a cross code comparison for the CBC using GS2, GENE and CGYRO, with the following results for the growth rate and mode frequency
image


This has 3 fields, sugama collisions (same as GS2 default with only deuterium) and a finite beta_prime using the `enhancement/gene_dat_outputs` branch.

I've pushed these input file to this branch. We can then get GKW on there (I don't know how to run that) and we should be good to go

bpatel2107:
It's likely that the difference in GENE was due to `hyp_v` in the GENE input file being non-zero. I found that they agree when this is set to 0.

dake0795 (approved):
All looks good to me on the GENE side. After looking at this yesterday with @bpatel2107 it looks like some of the discrepancy @arkabokshi was finding when comparing runs was due to a poor choice of defaults in the template (for which I take the blame). This should now be resolved.

arkabokshi:
@bpatel2107  Removing `input.gkw` and will include in the `add_gkw` branch. So this branch should be good to go I think.

PR #177, opened 2023-03-17
Multiple GENE fixes

bpatel2107:
It is now possible to load in data which has the format `*.dat` as well as `_0000` when reading GENE output data.

Also fixed bugs  in GENE:
-  Ensure `numerics.nenergy` is an `int`
- `PyroScan` can now load in GENE fluxes, which are not a function of `ky` so was breaking when loading data
- Switched frequency direction to match pyro convention and update tests
- Check if `dpdx_pm` in GENE is set consistently with `amhd` in the geometry namelist and if not raise a warning
- Set `beta_prime` even if `beta=0`


pyrokinetics/gk_code/GKOutputReaderGENE.py
```diff
@@ -31,7 +32,15 @@ def _get_gene_files(filename: PathLike) -> Dict[str, Path]:
         if filename.is_dir():
             # If given a dir name, looks for dir/parameters_0000
             dirname = filename
-            num_part = "0000"
+            dat_matches = np.all(
+                [os.path.exists(filename / f"{p}.dat") for p in prefixes]
```
LiamPattinson: We can avoid bringing in `os` by instead using `Path(filename / f"{p}.dat").is_file()` instead of `os.path.exists`.

tests/gk_code/test_gk_output_reader_gene.py
```diff
@@ -104,7 +104,7 @@ def test_infer_path_from_input_file_gene(input_path):
 # Update: Commit 9eae331 accounts for last time step (7d551eaa-2nd last step)
 # This data was gathered from templates/outputs/GENE_linear
 
-reference_data_commit_hash = "9eae331"
+reference_data_commit_hash = "3974780"
```
LiamPattinson, line 108: Should update the preceding comment

LiamPattinson (commented):
All looks good to me, besides a couple of very minor points. It'll probably be worth having somebody who knows GENE look over this before merging though.

pyrokinetics/gk_code/GKOutputReaderGENE.py
bpatel2107: Done

dake0795 (commented):
All looks good to me! 

Sorry for being slow @bpatel2107, I thought I'd also reviewed and approved this the other day!

dake0795 (approved):

PR #179, opened 2023-03-17
Feature/p file reader

jfparisi:
Functionality to read in a pFile object. Requires a geqdsk file to be read in concurrently. Can be called as:

```
from pyrokinetics import Pyro
eq_file = "your_path/geqdsk"
kinetics_file = "your_path/pFile"
pyro = Pyro(
    eq_file=eq_file,
    eq_type="GEQDSK",
    kinetics_file=kinetics_file,
    kinetics_type="pFile",
)
```


LiamPattinson:
I don't have time for a full review right now, but one thing I noticed is that the pfile reader is licensed under GPL, which means we'd have to re-license the project to GPL. I'll try to have a closer look after the weekend.

d7919:
> I don't have time for a full review right now, but one thing I noticed is that the pfile reader is licensed under GPL, which means we'd have to re-license the project to GPL. I'll try to have a closer look after the weekend.

If eqtools was installed as a requirement and we simply used/wrapped the pfile reader from there would that bypass the issue as we're not distributing their code directly?

pyrokinetics/kinetics/Kinetics.py
```diff
@@ -48,19 +50,32 @@ def __init__(
...
 
-        self.species_data = CleverDict(reader(kinetics_file, **kwargs))
+        if kinetics_type == "pFile":
+            self.species_data = CleverDict(
+                reader(kinetics_file, eq_file, **kwargs)
+            )  # Use reader __call__
+        else:
+            self.species_data = CleverDict(
+                reader(kinetics_file, **kwargs)
+            )  # Use reader __call__
```
ZedThree: Instead of special-casing the pFile reader, it's probably better to make all the readers take an (optional) equilibrium file. Then the `if else` above at L59 isn't needed either

ZedThree:
> If eqtools was installed as a requirement and we simply used/wrapped the pfile reader from there would that bypass the issue as we're not distributing their code directly?

Yes, this is probably better. We could even have it as an optional dependency, and then do something like:

```py
try:
  import eqtools
  have_eqtools = True
except ModuleNotFoundError:
  have_eqtools = False
```

LiamPattinson:
My understanding is that any Python project that has a GPL dependency must itself also be GPL. It you're directly invoking functions from a dependency, the Free Software Foundation consider your library and the dependency to be a single combined program, so if the dependency is GPL, your library must also be GPL. In the C/C++ world, that applies even if you're dynamically linking against a `.so`/`.dll` binary, and Python goes a step beyond that by requiring you to distribute the source code of your dependencies alongside your own code ([relevant bits](https://www.gnu.org/licenses/gpl-faq.html#NFUseGPLPlugins) of the [GPL FAQ](https://www.gnu.org/licenses/gpl-faq.html#GPLPlugins), and a [Stack Overflow question on the topic of mixing GPL and LGPL](https://softwareengineering.stackexchange.com/questions/342156/can-i-use-some-gpl-code-in-lgpl-project)).

Edit: I'm not sure how things work if we use eq_tools as an optional dependency, and only activate the pfile bits if the user already happens to have it installed.

pyrokinetics/readers.py
```diff
@@ -5,19 +5,31 @@
...
 
 
 class Reader(ABC):
     @abstractmethod
-    def read(self, filename: PathLike, *args, **kwargs) -> Any:
-        """Read and process a file"""
+    def read(
+        self,
+        filename: PathLike,
+        second_filename: Optional[PathLike] = None,
```
LiamPattinson: Rather than adding a `second_filename` arg to the `Reader` class, I'd prefer to make it more generic and take an arbitrary number of filenames:

```python
def read(self, *filenames: PathLike, **kwargs) -> Any:
    ...
```
We ran into a similar problem when reading GENE output data, as that's also spread across multiple different files, and it required a bit of a hack to get things working. I'll try to make a new PR into this one sometime today which will hopefully make the base class generic enough to handle any other cases like this we might run into in the future.

pyrokinetics/kinetics/__init__.py
```diff
@@ -3,11 +3,13 @@
 from .KineticsReaderSCENE import KineticsReaderSCENE
 from .KineticsReaderJETTO import KineticsReaderJETTO
 from .KineticsReaderTRANSP import KineticsReaderTRANSP
+from .KineticsReaderpFile import KineticsReaderpFile
 
 # Register each reader type with factory
 kinetics_readers["SCENE"] = KineticsReaderSCENE
 kinetics_readers["JETTO"] = KineticsReaderJETTO
 kinetics_readers["TRANSP"] = KineticsReaderTRANSP
+kinetics_readers["pFile"] = KineticsReaderpFile
```
bpatel2107, line 12: Do we want to keep everything capitalised?

bpatel2107:
I think the main addition needed is some tests. 

You can add the `pFile` tests here where you give the golden answers
https://github.com/pyro-kinetics/pyrokinetics/blob/unstable/tests/kinetics/test_kinetics.py

And you can make a copy of this for the `pFile` reader using this as a template
https://github.com/pyro-kinetics/pyrokinetics/blob/unstable/tests/kinetics/test_kinetics_reader_transp.py

pyrokinetics/kinetics/__init__.py
LiamPattinson: If I'm right in thinking that `pFile == P-EQDSK`, we could call it `PEQDSK` to keep things consistent with the equilibrium part of the code which calls it `GEQDSK`.

pyrokinetics/kinetics/Kinetics.py
LiamPattinson: I'm working on this now, and I found a cleaner solution is just to pass `eq_file` via `**kwargs`, and raise an error if it isn't present for the pFile reader.

bpatel2107:
@jfparisi you should be able to do

```
git checkout feature/pFile_reader
git pull
```

which will update your local version

LiamPattinson:
I've opened a [PR to FreeQDSK ](https://github.com/freegs-plasma/FreeQDSK/pull/5) that adds pFile read/write methods. Unlike the code from eqtools, this will be compatible with our license. If that PR gets merged, I can advise on how to move over to using those functions instead (or I can try to convert it myself in another PR).

LiamPattinson:
FreeQDSK version 0.3 now has a pfile reader. Details on how to use it are in the docs, but the gist of it is:

```python
from freeqdsk import peqdsk

with open(filename, "r") as f:
    data = peqdsk.read(f)

profiles = data["profiles"] # dict of [profile names, dict of profile data]
species = data["species"] # list of species data

# profiles first...
# get electron density profile
psiN = profiles["ne"]["psinorm"]
ne = profiles["ne"]["data"]
dne_dpsiN = profiles["ne"]["derivative"]
# get units
units = profiles["ne"]["units"] # just given as a string

# similarly, for other quantities:
ion_density_profile = profiles["ni"]
electron_temperature_profile = profiles["te"]

# Species is just a list of the species in the order they were listed in the file
atomic_number = species[0]["N"]
charge = species[0]["Z"]
atomic_mass = species[0]["A"]
```

bpatel2107:
Once #185 is merged I can update this to use the latest version of FreeQDSK and in units.

@jfparisi is pFile the same as PEQDSK? I thought they were the same but do you know if they are different? I think we are planning on calling it PEQDSK here but I hope that they are the same.

We can probably write this in a way such that we can set `kinetics_type=pFile` or `PEQDSK` and this will work

jfparisi:
@bpatel2107 Yes, I believe pFile and PEQDSK are the same.

pyrokinetics/kinetics/KineticsReaderpFile.py
```diff
@@ -0,0 +1,241 @@
...
+                    f"""\
+                    {self.__class__.__name__} must be provided with a G-EQDSK file via
+                    the keyword argument 'eq_file'.
+                    """
+                )
+            )
+
+        # Read pFile, get generic data.
+        pFile = PFileReader(str(filename))
+        psi_n = pFile.__getattribute__("ne").x
```
d7919: Isn't it possible in a pFile for different quantities to have different psinorm grids?
In this method I think we're assuming all quantities live on the save psinorm grid as given by the electron density values?

pyrokinetics/kinetics/KineticsReaderpFile.py
```diff
@@ -0,0 +1,241 @@
...
+        )  # Converting 10^20 m^-3 to m^-3.
+
+        electron_temp_func = InterpolatedUnivariateSpline(
+            psi_n, electron_temp_data
+        )  # Interpolate on psi_n.
+        electron_dens_func = InterpolatedUnivariateSpline(psi_n, electron_dens_data)
+
+        # Read geqdsk file, obtain rho_func.
+        geqdsk_equilibrium = read_equilibrium(str(eq_file))
+        rho_g = geqdsk_equilibrium["r_minor"].values
```
d7919, line 97: I think in other kinetics readers this may be the normalised r_minor (r/a). I'm not sure if that is actually important here or not.

bpatel2107:
This should be good to go now. @jfparisi let me know if you have anything else to add

pyrokinetics/kinetics/KineticsReaderpFile.py
jfparisi: @d7919 @bpatel2107  Yes David, you're completely correct. It's not only possible, but normal for different quantities to have different psinorm grids.

PR #180, opened 2023-03-20
Contributions to 'Feature/p file reader'

LiamPattinson:
Contributions to PR https://github.com/pyro-kinetics/pyrokinetics/pull/179

pFile reader requires an additional G-EQDSK file to determine its $\rho$ grid, so a `second_filename` arg was added to the `Reader` class. As this class is very generic and should be applicable throughout the project, I think it'd be better to keep it simple and make use of the `**kwargs` passed to the `read` function to handle any special cases. This PR reverts the changes to `Kinetics` and `Reader`, and instead passes `eq_file` via standard `**kwargs`. Further changes:

- Updated pFile reader ``verify`` method, automatic file type inference now working:
```python
>>> # No need for kinetics_type:
>>> import pyrokinetics as pk
>>> pyro = pk.Pyro(kinetics_file=pk.kinetics_templates["pFile"], eq_file=pk.eq_templates["GEQDSK"])
```
- ``Pyro`` now takes optional dicts `eq_kwargs` and `kinetics_kwargs` to its constructor, so complicated builds that previously required separate calls to `load_global_eq` and `load_global_kinetics` can now be done in one step.
- If ``Pyro`` has an ``eq_file`` and the user tries to load a pFile kinetics, it will detect if the error is simply a missing ``eq_file`` arg and correct it automatically.
- Updated Reader docs, removed unused `*args` and fixed some type hints

### DIFF
--- a/pyrokinetics/kinetics/KineticsReaderpFile.py
+++ b/pyrokinetics/kinetics/KineticsReaderpFile.py
@@ -3,19 +3,20 @@ Reads in an Osborne pFile: https://omfit.io/_modules/omfit_classes/omfit_osborne
 
 
 """
-from typing import Dict
-from ..typing import PathLike
-from .KineticsReader import KineticsReader
-from ..species import Species
-from ..constants import electron_mass, hydrogen_mass, deuterium_mass
-from .pfile_reader import PFileReader
-from pyrokinetics.equilibrium.equilibrium import read_equilibrium
-import numpy as np
-from scipy.interpolate import InterpolatedUnivariateSpline, interp1d
-import copy
-import csv
+
 import re
-from collections import namedtuple
+from textwrap import dedent
+from typing import Dict, Optional
+
+from .KineticsReader import KineticsReader
+from .pfile_reader import PFileReader
+from ..constants import electron_mass, hydrogen_mass
+from ..equilibrium.equilibrium import read_equilibrium
+from ..species import Species
+from ..typing import PathLike
+
+import numpy as np
+from scipy.interpolate import InterpolatedUnivariateSpline
 
 
 def ion_species_selector(nucleons, charge):
@@ -63,19 +64,30 @@ class KineticsReaderpFile(KineticsReader):
     def read(
         self,
         filename: PathLike,
-        second_filename: PathLike,
+        eq_file: Optional[PathLike] = None,
         time_index: int = -1,
-        time: float = None,
+        time: Optional[float] = None,
     ) -> Dict[str, Species]:
         """
         Reads in Osborne pFile. Your pFile should just be called, pFile.
 
-        Also reads a geqdsk file, which is in the same directory as the pFile, and assumed to be called geqdsk.
-
-        geqdsk file passed through second_filename.
+        Also reads a geqdsk file via eq_file, which is in the same directory as the
+        pFile, and assumed to be called geqdsk.
         """
-        # Read pFile, get generic data.
+        # eq_file must be provided
+        # Note: The 'Optional[T]' type hint just means the arg can be of type T or it
+        # can be 'None' -- it doesn't mean the user can skip providing it!
+        if eq_file is None:
+            raise ValueError(
+                dedent(
+                    f"""\
+                    {self.__class__.__name__} must be provided with a G-EQDSK file via
+                    the keyword argument 'eq_file'.
+                    """
+                )
+            )
 
+        # Read pFile, get generic data.
         pFile = PFileReader(str(filename))
         psi_n = pFile.__getattribute__("ne").x
         electron_temp_data = pFile.__getattribute__("ne").y
@@ -87,7 +99,7 @@ class KineticsReaderpFile(KineticsReader):
         electron_dens_func = InterpolatedUnivariateSpline(psi_n, electron_dens_data)
 
         # Read geqdsk file, obtain rho_func.
-        geqdsk_equilibrium = read_equilibrium(str(second_filename))
+        geqdsk_equilibrium = read_equilibrium(str(eq_file))
         rho_g = geqdsk_equilibrium["r_minor"].values
         psi_n_g = geqdsk_equilibrium["psi_n"].values
         rho_func = InterpolatedUnivariateSpline(psi_n_g, rho_g)
@@ -113,7 +125,7 @@ class KineticsReaderpFile(KineticsReader):
 
         num_ions = pFile.__getattribute__("ions").nions
 
-        ## Check whether fast particles.
+        # Check whether fast particles.
         fast_particle = 0
         if "nb" in pFile._params:
             fast_particle = 1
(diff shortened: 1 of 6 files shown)


bpatel2107:
This looks very nice! Can you add an example with `eq_kwargs` or `kinetic_kwargs`. You could also modify this script here

https://github.com/pyro-kinetics/pyrokinetics/blob/unstable/pyrokinetics/examples/example_JETTO.py

Can you add a test for these kwargs? You could use `time` in TRANSP/JETTO netCDFs and `psi_n_lcfs` for GEQDSK

In #179 we can add some tests for when a `pFile` is loaded without a `eq_file`

pyrokinetics/kinetics/KineticsReaderpFile.py
```diff
@@ -63,19 +64,30 @@ class KineticsReaderpFile(KineticsReader):
...
 
-        Also reads a geqdsk file, which is in the same directory as the pFile, and assumed to be called geqdsk.
-
-        geqdsk file passed through second_filename.
+        Also reads a geqdsk file via eq_file, which is in the same directory as the
+        pFile, and assumed to be called geqdsk.
         """
-        # Read pFile, get generic data.
+        # eq_file must be provided
+        # Note: The 'Optional[T]' type hint just means the arg can be of type T or it
```
bpatel2107: I'm not sure I understand this. Why wouldn't `eq_file: PathLike = None`  work?

LiamPattinson:
Sorry for rerunning CI repeatedly there, kept catching minor errors after pushing.

I changed the JETTO example to use `eq_kwargs` and `kinetics_kwargs`, and I added a couple of simple tests to check that kwargs were being forwarded properly.

As the pFile reader doesn't have tests yet, I think it'd be best to hold off on writing those tests until the PR is closer to being finished.

bpatel2107 (approved):
I think these simple tests are good enough for this. Happy to merge

pyrokinetics/kinetics/KineticsReaderpFile.py
LiamPattinson: It'd work just fine, but a static type checker might complain, as the default value `None` isn't of type `PathLike`. The comment is just pointing out that `Optional[T]` is really an alias for `Union[T, None]` (or maybe `Union[T, type(None)]`?), and so despite the implications of the name 'optional', in this case we do actually require the user to provide something.

In other words, it's me being weirdly pedantic about type hints, so it's probably best to just remove the type hint and the comment since they're actually degrading readability in this case, and that kinda defeats the whole point.

jfparisi:
I can't find pk.kinetics_templates["pFile"]? Is this an error on my side, or do we need to upload one?

LiamPattinson:
It should already be there, as these lines added it:

https://github.com/pyro-kinetics/pyrokinetics/blob/ad3053a7019ca55e8f0b12bd21397f55480ca3de/pyrokinetics/templates.py#L24-L33

Did you definitely do `import pyrokinetics as pk` and not include any typos? If so, please could you post your error message?

Not sure if this is the issue, but to make sure you're running my changes, you'll need to clone a new pyrokinetics from `github.com/PlasmaFAIR/pyrokinetics` instead of `github.com/pyro-kinetics/pyrokinetics`, and then switch to the `feature/pFile_reader` branch. I started working on this project from a fork and still haven't broken the habit, and it only just occurred to me that this just makes it harder for others to test my code.

bpatel2107:
I think you either need to checkout the forked PlasmaFAIR branch, or we'd need to merge this into your branch and then you can pull those changes to your local machine

bpatel2107:
I'll just merge this so @jfparisi can pull the latest version of `feature/pFile_reader`

jfparisi:
@LiamPattinson @bpatel2107  I probably messed up. I was testing on: git fetch origin pull/180/head:feature/pFile_reader. Could you tell me the correct way to checkout these further updates and test from the command line?

LiamPattinson:
I'm honestly not sure if that would work -- I tend to use the same few basic git commands 99% of the time, and I don't think I've ever tried fetching/pulling from a PR. Now it's merged into your branch you should be able to just `git pull` on your regular branch, so that'll make things easier. When dealing with forks in general I'd normally handle it like this:

```bash
$ cd ~/my/workspace
$ # get the fork instead of the main version
$ git clone https://github.com/PlasmaFAIR/pyrokinetics pyrokinetics_fork
$ cd pyrokinetics_fork
$ # new python env (I use venv, conda works just as well)
$ python3 -m venv venv
$ source venv/bin/activate
$ # install, run tests
$ pip install --upgrade pip
$ pip install -e .[tests]
$ pytest
```
If you want to play about in the command line, it's usually best to do so from the top level directory, as relative import paths can break if you try to run things from within the source code itself.

If you're not happy with any of the merged changes, please let me know in the [original PR](https://github.com/pyro-kinetics/pyrokinetics/pull/179). We can always revert if we need to.

jfparisi:
Thanks for your help @LiamPattinson! The merged changes are excellent.

I am finding issues when writing GK input files. For example, with the below code, I get beta ~ 1e-25, vnewk ~ 1e-16 in the GK input files.
```
import pyrokinetics as pk
import pathlib
gs2_template = pk.template_dir / "input.gs2"
kinetics_file_in='/p/gyro/jparisi/equilibria/NSTX/nstx_132543/Oak_Dec2022/1.2_1.2_1.0/pFile'
eq_file_in = '/p/gyro/jparisi/equilibria/NSTX/nstx_132543/Oak_Dec2022/1.2_1.2_1.0/geqdsk'
base_path = '/p/gyro/jparisi/equilibria/NSTX/nstx_132543/Oak_Dec2022/1.2_1.2_1.0/'
pyro = pk.Pyro(kinetics_file=kinetics_file_in, eq_file=eq_file_in)
pyro.load_local(psi_n=0.5, local_geometry="Miller")
pyro.gk_code = "GS2"
base_path = pathlib.Path(base_path)
pyro.write_gk_file(file_name=base_path / "test_pFile.gs2")
pyro.write_gk_file(file_name=base_path / "test_pFile.cgyro", gk_code="CGYRO")
pyro.write_gk_file(file_name=base_path / "test_pFile.gene", gk_code="GENE")
pyro.write_gk_file(file_name=base_path / "test_pFile.tglf", gk_code="TGLF")
```
I have uploaded the geqdsk and pFile to[ this Google Drive folder](https://drive.google.com/drive/u/0/folders/1r6qsgUNy3r-S-NQn3E3n0xlkl4NBfYpY). Is this user error on my side?

bpatel2107:
This sounds like a units issue. Can you check that the density and temper are in units of `/m3` and `eV`?

jfparisi:
In the pFile, density is 10^19/m3 and temperature is in keV. That would do it! Shall we convert to /m3 and ev in the pFile reader itself?

bpatel2107:
Yeah that should work

PR #181, opened 2023-03-21
Bugfix/cocos negative ip bt

bpatel2107:
Currently the COCOS checks in equilibrium require `psi` to be increasing, but that is only true if `I_p` is positive. If `I_p` is negative then COCOS=11 expects `psi to decrease.

image

This made the interpolation a tad messy as that needs `psi` to be increasing. We could have the arrays reversed in `UnitSpline` but I'm not sure that is better as we only want to allow for it to be reversed if `I_p` is negative


pyrokinetics/equilibrium/equilibrium.py
```diff
@@ -321,11 +321,17 @@ def __init__(
         R_major = np.asfarray(R_major) * eq_units["len"]
         r_minor = np.asfarray(r_minor) * eq_units["len"]
         Z_mid = np.asfarray(Z_mid) * eq_units["len"]
+
+        B_0 = B_0 * cocos_factors["BT"] * eq_units["B"]
+        I_p = I_p * cocos_factors["IP"] * eq_units["I"]
```
LiamPattinson: An issue with moving these here is that it'll fail for equilibria where we can't define `B_0` and `I_p`, as they're `None` by default. We could upgrade these to non-optional positional arguments, or we could have something like:

```python
Ip_sign = 1 if I_p is None else int(np.sign(I_p))
```
The bit at the bottom of the function where `B_0` and `I_p` are added to `attrs` would then be the same as it was.

pyrokinetics/equilibrium/equilibrium.py
```diff
@@ -321,11 +321,17 @@ def __init__(
...
+
+        Ip_sign = int(np.sign(I_p))
+
         # Ensure psi is 1D and monotonically increasing
         if len(psi.shape) != 1:
             raise ValueError("The grid psi must be 1D.")
-        if np.any(np.diff(psi) <= 0):
-            raise ValueError("The grid psi must have a positive spacing.")
+        if np.any(np.diff(psi * Ip_sign) <= 0):
+            raise ValueError(f"The grid {Ip_sign} * psi must have a positive spacing.")
```
LiamPattinson: It'd be better to not have an f-string and just say something like `"The grid 'sign(I_p) * psi' must have a positive spacing."`, as otherwise the error message would change between different files and the user might not know why.

pyrokinetics/equilibrium/equilibrium.py
```diff
@@ -347,8 +353,10 @@ def __init__(
 
         # Check that floats are valid
         psi_lcfs = float(psi_lcfs) * cocos_factors["PSI"] * eq_units["psi"]
-        if psi_lcfs < psi[0]:
-            raise ValueError("psi_lcfs should be greater than psi[0].")
+        if Ip_sign * psi_lcfs < Ip_sign * psi[0]:
+            raise ValueError(
+                f"{Ip_sign} * psi_lcfs should be greater than {Ip_sign} * psi[0]."
```
LiamPattinson: Maybe something like `"psi_lcfs should be greater than psi[0] when I_p > 0, and vice versa when I_p < 0"`?

pyrokinetics/equilibrium/equilibrium.py
```diff
@@ -363,15 +371,26 @@ def __init__(
...
-        self._p_psi_spline = UnitSpline(psi, p)
-        self._p_prime_psi_spline = UnitSpline(psi, p_prime)
-        self._q_psi_spline = UnitSpline(psi, q)
-        self._R_major_psi_spline = UnitSpline(psi, R_major)
-        self._r_minor_psi_spline = UnitSpline(psi, r_minor)
-        self._Z_mid_psi_spline = UnitSpline(psi, Z_mid)
+        self._F_psi_spline = UnitSpline(psi[::index_direction], F[::index_direction])
+        self._FF_prime_psi_spline = UnitSpline(
+            psi[::index_direction], FF_prime[::index_direction]
+        )
```
LiamPattinson: We could implement this behaviour in `UnitSpline` to keep this cleaner and avoid having to use `index_direction`. Currently the `__init__` is this:
```python
    def __init__(self, x: ArrayLike, y: ArrayLike):
        self._x_units = x.units
        self._y_units = y.units
        self._spline = InterpolatedUnivariateSpline(x.magnitude, y.magnitude)
```
It could be altered to this:
```python
    def __init__(self, x: ArrayLike, y: ArrayLike):
        self._x_units = x.units
        self._y_units = y.units
        x_mag = x.magnitude
        y_mag = y.magnitude
        # Assume x is monotonically increasing/decreasing
        if x_mag[1] > x_mag[0]:
            self._spline = InterpolatedUnivariateSpline(x_mag, y_mag)
        else:
            self._spline = InterpolatedUnivariateSpline(x_mag[::-1], y_mag[::-1])
```

LiamPattinson (commented):
Good catch!

Testing might be tricky, as currently everything is compared to a single set of expected values, and the inputs to `Equilibrium` are then modified to check it can handle different units and COCOS conventions. We may therefore have to parametrize the expected values to test negative $\psi$/$I_p$, which will double the total number of tests run. I'll try it out this afternoon if I have time.

pyrokinetics/equilibrium/equilibrium.py
bpatel2107: Yeah I thought this would work, wasn't sure if we wanted to be explicit in when we reverse, but this is probably fine

LiamPattinson:
My bad, I think this line:

```python
Ip_sign = 1 if I_p is None else int(np.sign(I_p))
```

Should be:

```python
Ip_sign = 1 if I_p is None else int(np.sign(I_p * cocos_factors["IP"]))
```

bpatel2107:
@LiamPattinson any thoughts on how to parameterise these expected results? I've spent a very frustrating afternoon attempting this with no luck

LiamPattinson:
Sorry at not looking at this for a while, got pulled away working on other features and forgot this was one I was supposed to be working on.

For a minimum-effort test, we could try supplying a G-EQDSK file that has decreasing $\psi$ and negative $I_p$ and just check that an `Equilibrium` can be made and that it contains the expected results. I don't think I'll have time to work on more thorough tests in the near future (at least not without dropping work on another PR).

bpatel2107:
Well that was significantly easier to do! Added in a test so I'm happy for this to be merged

pyrokinetics/equilibrium/equilibrium.py
```diff
@@ -347,8 +350,11 @@ def __init__(
 
         # Check that floats are valid
         psi_lcfs = float(psi_lcfs) * cocos_factors["PSI"] * eq_units["psi"]
-        if psi_lcfs < psi[0]:
-            raise ValueError("psi_lcfs should be greater than psi[0].")
+        print(Ip_sign, psi_lcfs, psi[0])
```
LiamPattinson: Is this needed? Or just a debugging line?

tests/equilibrium/test_equilibrium.py
```diff
@@ -600,6 +600,7 @@ def test_circular_eq_netcdf_round_trip(tmp_path, circular_eq):
         ("transp_eq.cdf", "TRANSP"),
         ("transp_eq.geqdsk", "GEQDSK"),
         ("test.geqdsk", "GEQDSK"),
+        ("reverse_ipbt.geqdsk", "GEQDSK"),
```
LiamPattinson: Rather than putting it in the filetype inference test, could this have a new test? I don't think it'd need to be all that complicated, just something like the following:

```python
def test_reverse_ipbt():
    # Check we can handle files with decreasing psi
    eq = read_equilibrium(template_dir / "reverse_ipbt.geqdsk")
    assert(eq.psi(0.0) > eq.psi(1.0))
    assert(eq.I_p == ...) # golden answer, whatever negative value it takes
    assert(eq.f(0.0) == ...) # check a few arrays, make sure they haven't been accidentally reversed
```

LiamPattinson (commented):
Wow, that was fast. Generally looks good to me, only a couple of simple points:

pyrokinetics/equilibrium/equilibrium.py
bpatel2107: Debugging!

tests/equilibrium/test_equilibrium.py
bpatel2107: Added a more rigorous test

tests/equilibrium/test_equilibrium.py
```diff
@@ -614,3 +614,39 @@ def test_supported_equilibrium_types():
...
+        "a_minor": 1.5000747773827081,
+        "psi_axis": -0.0,
+        "psi_lcfs": -13.842116098951283,
+        "I_p": -21e6,
+        "B_0": -2.4,
+    }
+
+    for key, value in expected_attrs.items():
+        actual = getattr(eq, key)
+        assert actual.m == value
```
LiamPattinson: I'd recommend changing this and the assert line below to `assert np.isclose(actual.m, value)` to avoid floating point errors popping up in future, but otherwise happy to merge.

bpatel2107: Done!

LiamPattinson (approved):

PR #182, opened 2023-03-21
Enhancements/gk output normalisation

### LINKED ISSUE #176: GK Output normalisation
Currently the output from different codes don't have any units and normalisations are manually forced to convene with pyro's normalisation. We should add in units/normalisation to the outputs.

bpatel2107:
To make pyro IMAS compliant it is necessary to have units on the GK outputs. This will allow for re-normalisation and a more direct comparison between the codes.

Fixes #176


bpatel2107:
@ZedThree @LiamPattinson I've implemented units for GS2 for now. Can you have a look and see if I have done this correctly and if so I can copy this over to the other codes.

@d7919 Can you check the normalisations for GS2? The expected units are shown in GKOutputReader docstring

pyrokinetics/gk_code/GKOutputReader.py
```diff
@@ -50,15 +51,20 @@ class GKOutputReader(Reader):
...
         pitch       1D array of floats
         moment      ["particle", "energy", "momentum"]
         field       ["phi", "apar", "bpar"] (the number appearing depends on nfield)
         species     list of species names (e.g. "electron", "ion1", "deuterium", etc)
 
     data_vars
-        fields      (field, theta, kx, ky, time) complex array, may be zeros
-        fluxes      (species, moment, field, ky, time) float array, may be zeros
+        phi                    (theta, kx, ky, time) complex array, units [qref / tref * lref / rhoref], may be zeros
+        apar                   (theta, kx, ky, time) complex array, units [1 / bref * lref / rhoref **2], may be zeros
```
d7919: I think it may be missing a factor 2 on apar. I'd also suggest using brackets to clearly group the numerator and denominator.

pyrokinetics/gk_code/GKOutputReader.py
```diff
@@ -50,15 +51,20 @@ class GKOutputReader(Reader):
...
+        bpar                   (theta, kx, ky, time) complex array, units [1 / bref * lref / rhoref], may be zeros
```
d7919: In GS2 bpar use B(theta) rather than B_ref to normalise Bpar. This might complicate things. Probably the easiest would be to adjust for this when reading in (e.g. scale by Bmag)?

pyrokinetics/gk_code/GKOutputReaderGS2.py
```diff
@@ -166,36 +193,67 @@ def _set_fields(
...
+            field = (
                 field_data[0, :, :, :, :] + 1j * field_data[1, :, :, :, :]
-            )
+            ) * field_unit
+
+            # Account for
+            if "apar" in field_name:
+                field *= 0.5
+            elif "bpar" in field_name:
+                field *= raw_data["bmag"].data[:, np.newaxis, np.newaxis, np.newaxis]
```
bpatel2107: @d7919 I've currently got this which I think is what you were referring to?

d7919: Ah yes that account for the apar and bpar differences I commented on.

LiamPattinson:
The `DatasetWrapper` class isn't intended to be used directly, and it should instead be used as a base for classes that would benefit from a lot of (pint-)Xarray dataset functionality, but should also have a rigorously defined constructor. The issue with returning raw datasets is that we can put just about anything into them, so it's hard to ensure that all codes are producing equivalent objects, especially since each works in different units/normalisations. `DatasetWrapper` is intended to help that problem by allowing us to define a constructor for our dataset-like classes which can perform validity checks and convert units automatically.

What we'll probably need here is a `GKOutput` class which inherits `DatasetWrapper` and does things similarly to `Equilibrium`:

```python
class GKOutput(DatasetWrapper): 

    # The standard units we'll use for each quantity                     
    # I think normalised units like lref etc will work here?                                                                                                   
    _init_units = {
        "self": None,
        "time": time_units, # ureg.seconds?
        "kx": kx_units, # whatever units kx is expressed in
        ...
    }

    # Decorate init function with 'wraps'
    # If passed in units that don't match exactly, but have the same dimensions,
    # this automatically converts the inputs to the units defined in _init_units
    # above. If passed in something that doesn't have units, it assumes those
    # quantities are already in the correct units. Note that it also strips
    # units within the function itself, so we'll need to add them back!
    @ureg.wraps(
        None, # return type
        [*_init_units.values()], # units for each argument
        strict=false, # allow inputs that don't already have units
    )
    def __init__(
        self,
        time: ArrayLike,
        kx: ArrayLike,
        ky: ArrayLike,
        ... # include all of the required coords, data_vars, and attrs
    ):
        # Check that the inputs are valid
        if not test_that_time_array_makes_sense(time):
            raise ValueError(...)

        # Add units
        time = np.asfarray(time) * time_units

        # Assemble coords/data_vars
        # (See Equilibrium for adding descriptions etc)

        coords = {
            "time": time,
            ...
        }

        data_vars = {
            ...
        }

        attrs = {
            ...
        }

        # Pass on to the super().__init__()
        # This performs pint_xarray quantify stuff 
        super().__init__(data_vars=data_vars, coords=coords, attrs=attrs)
```

The job of the `GKOutputReader` classes will then be to gather the necessary data from their output files and return a `GKOutput` object. We don't necessarily need to stick to the current strategy of building an Xarray dataset piece-by-piece through a series of `staticmethod`s, but we could do so just to keep things organised while we construct our arrays. It'd need to be disassembled at the point we return a `GKOutput` though:

```python
return GKOutput(
    time=data["time"] * time_units,
    kx=data["kx"] * kx_units,
    ...
)
``` 

If the units we choose for `GKOutput` are convertible from the GS2 units/normalisations, we should just be able to pass in the GS2 outputs as-is and let the `wraps` decorator handle the conversion for us. However, if the plan is to have normalised units as the standard, I'm not sure if it's possible to have `SimulationNormalisation` stuff interact with `wraps`, and it may be that we need to ignore `wraps` and perform all the conversions manually within the `__init__` function. @ZedThree any ideas?

bpatel2107:
Hmm I had a feeling I was using it wrong.

So at the moment the different GKOutputReaders are generating an Dataset with all the data. So what I would need to do is to then take that Dataset and feed in the individual components into this `GKOutput` object that has a set constructor. And all the units need to be defined there too.

LiamPattinson:
Yeah, that's about it. Since you'll probably need to use `SimulationNormalisation` to handle the units, I think that the `ureg.wraps` bit won't actually work in this case, and instead you'll need to pass a `norm` object as one of the arguments to the constructor and then convert units manually. Looking at how it's handled in `GKInputGS2`, maybe the function `convert_dict` could be used to make it easier.

bpatel2107:
So I've reverted back to have the `GKOutputReader` using an `xr.Dataset` instead of `DatasetWrapper` and I have got it to the point where everything is working as before, but with units now. A few outputs have changed slightly due to the normalisations changing.

### Major changes
- The output dataset now has the fields and fluxes separated into separate components due to their different units. This will break a lot of existing scripts...
- I've tweaked the name of the heat flux from `energy` to `heat` to prevent confusion if the energy moment calculation is added in the future
- Due to the normalisations being done on the fields before the eigenvalue calculations, from now eigenvalues will be slightly different to before

### Issues
- It seems that using `da.pint.to({da.name : local_norm.pyrokinetics.bref)` doesn't work and ends up with a NaN and I'm not sure why. I worked around this by using `da.data.to(local_norm.pyrokinetics)`. This also means mapping all the output in one go in not simple. I think the ratios between units are not being defined when you get down to the Convention class but I'm not sure how to get that set up. @ZedThree any thoughts
- I'm not sure how a generic `GKOutput` object using `DatasetWrapper` would work as each code will likely have slightly different structure and it will change for nonlinear/linear or intial value/eigenvalue codes. Do you think it is still worth setting up?
- I have to use `pint.dequantify` a lot when doing using basic numpy operations, otherwise I get an lot of `UnitStrippedWarnings`. Not sure if there is a way around this but will likely get annoying going forward

bpatel2107:
I should clarify with the `pint.to` method, converting between ratios immediately defined like `vref_nrl` and `vref_most_probable` works fine, but when you have to define the ratios like using `bref_B0` and `bref_Bunit` we get the NaN. So pint doesn't seem to know about the context

Here's a minimal working example

```python
import xarray
import pint
import pint_xarray

pyro = Pyro(gk_file=template_dir / "outputs/CGYRO_nonlinear/input.cgyro")
pyro.load_gk_output()
norms = pyro.norms

velocity_test = xarray.DataArray(10, name="test").pint.quantify(norms.pyrokinetics.vref)
norms.default_convention = "gs2"

print(velocity_test)
print(velocity_test.pint.to({velocity_test.name: norms.vref}))
print(velocity_test.pint.to(norms.vref))
print(velocity_test.data.to(norms.gs2))

print(" ")


bfield_test = xarray.DataArray(10, name="test").pint.quantify(norms.pyrokinetics.bref)
norms.default_convention = "cgyro"

print(bfield_test)
print(bfield_test.pint.to({bfield_test.name: norms.bref}))
print(bfield_test.pint.to(norms.bref, unit_registry=norms.units))
print(bfield_test.data.to(norms.cgyro))

```

bpatel2107:
Given that we are given the field units with a $\rho_* $  term anyway should we remove this somewhat arbitrary `rho_star` when loading the fields. That `data.rho_star` isn't even $\rho_* $

https://github.com/pyro-kinetics/pyrokinetics/blob/7754284c12856f09f01173374d696cd9c70c171b/pyrokinetics/gk_code/GKOutputReaderCGYRO.py#L287-L289


Here are the pyro units for the fields which as the actual $\rho_*$

https://github.com/pyro-kinetics/pyrokinetics/blob/7754284c12856f09f01173374d696cd9c70c171b/pyrokinetics/gk_code/GKOutputReader.py#L384-L388

@hgd511 @mgiacomin  thoughts

mgiacomin:
Let's say that we remove `data.rho_star` from 

https://github.com/pyro-kinetics/pyrokinetics/blob/7754284c12856f09f01173374d696cd9c70c171b/pyrokinetics/gk_code/GKOutputReaderCGYRO.py#L287-L289

then wouldn't the CGYRO normalisation for $\delta A_\parallel$ be $\rho_\mathrm{unit}B_{unit}$ only?

hgd511:
Yes we should be careful here. As Bhavin points out, `data.rho_star` is not $\rho_*$, as this is arbitrarily small in the gyrokinetic framework. The relation is `data.rho_star` $= (RMIN / Q) * KY$, which does seem to be a somewhat arbitrary factor.

bpatel2107:
What they have with `data.rho_star` is the $\rho_*$ if the minimum $k_y$ in the system corresponds to $n=1$. I guess the question is whether or not the code assumes that the `data.rho_star` is the normalised value or not.

If it does then we should divide by `rho_star` as then we will get the truly normalised value and then it doesn't matter how we map back to the un-normalised value. 

Looking at how they normalise the heat fluxes, which we know have units of $Q_{gB}$, they need to divide by `data.rho_star **2` (one factor from the field and one from the distribution function). This is done internally in the code, but no such factor is applied when writing out the fields. This tells me that it needs that `rho_star` term should be there in pyro.

mgiacomin:
Okay so if the fields are divided by `data.rho_star`, this brings in an extra $B_\mathrm{unit}$ dependence, and $\delta A_\parallel$ is normalised to $\rho_\mathrm{unit}^2B_\mathrm{unit}/a$, which is what is done currently in `pyrokinetics`. Would it be good to compare CGYRO and GS2 (in which we are sure about the normalisation) fields output for a case we know there is agreement (with $B_\mathrm{unit}/B_0 \ne 1$)?

bpatel2107:
Yeah I did this a while ago manually. We ideally want them to have the same starting conditions. @d7919 is the CGYRO initial condition still in GS2?

d7919:
> 

Yes it is.

mgiacomin:
As @hgd511 wrote before, 
```math
(\rho_*)_\mathrm{CGYRO} = k_{y, \mathrm{min, CGYRO}} \frac{r_\mathrm{min, CGYRO}}{q}
```
which leads to
```math
(\rho_*)_\mathrm{CGYRO}= \Bigl(\frac{n_0 q}{a} \rho_s \frac{B_\mathrm{unit}}{B_0}\Bigr) \frac{r}{aq} = \frac{B_\mathrm{unit}}{B_0} \frac{\rho_sn_0}{a}\frac{r}{a} = \frac{B_\mathrm{unit}}{B_0} \rho_*\frac{r}{a}
```
where I assume $n_0=1$ in CGYRO. From this we obtain
```math
(\delta A_\parallel)_\mathrm{CGYRO} = \frac{\delta A_\parallel}{(\rho_*)_\mathrm{CGYRO}\rho_\mathrm{unit}B_\mathrm{unit}} = \frac{\delta A_\parallel}{(\rho_*)_\mathrm{CGYRO}\rho_sB_0} = \frac{\delta A_\parallel}{\rho_* \rho_s B_0}\frac{a}{r}\frac{B_0}{B_\mathrm{unit}} = (\delta A_\parallel)_\mathrm{PYRO} \frac{a}{r}\frac{B_0}{B_\mathrm{unit}}
```
Therefore I think there should be an extra `RMIN` factor in the normalisation and this applies to all fields. @bpatel2107  what do you think?

pyrokinetics/gk_code/GKOutputReader.py
```diff
@@ -278,4 +322,90 @@ def _set_growth_rate_tolerance(data: xr.Dataset, time_range: float = 0.8):
...
+    Returns
+    -------
+        field_units [Dict] : dictionary of the flux units
+
+    """
+
+    return {
+        "phi": convention.qref / convention.tref * convention.lref / convention.rhoref,
+        "apar": convention.lref / convention.rhoref**2 / convention.bref,
+        "bpar": convention.lref / convention.rhoref / convention.bref,
```
bpatel2107: These field are inverted! bpar currently has units of `1/Tesla`

bpatel2107:
> Therefore I think there should be an extra `RMIN` factor in the normalisation and this applies to all fields. @bpatel2107 what do you think?

Not quite I think.

$k_{y,min} = \frac{ n_0 q }{r}$ not $\frac{ n_0 q }{a}$ so the $r$ cancel out

mgiacomin:
> > Therefore I think there should be an extra `RMIN` factor in the normalisation and this applies to all fields. @bpatel2107 what do you think?
> 
> Not quite I think.
> 
> ky,min=n0qr not n0qa so the r cancel out

Yes, you are right.

bpatel2107:
Here are the initial conditions from the template case with $\kappa=2$ and `BETAE_UNIT=0.1` for CGYRO and GS2. 

I've used `AMP=0.001`  in CGYRO

I've used `phiinit = 0.001 * B0/Bunit / sqrt(2) = 0.00034864359987170964` and also  `ginit_option='cgyro'` for GS2

Here is the output from pyro as is without any sort of changes. Apar should be zero as the distribution function is initialised as even

image

I had the field units inverted so I've tweaked that, will probably break all the tests as the golden answers I made are wrong now

I feel like the `sqrt(2)` to change initial amplitudes iswrong though...

mgiacomin:
I think a very nice test would be to compare the saturated value of $\delta \phi$ and $\delta A_\parallel$ from GS2 and CGYRO nonlinear simulations. The actual value of the initial condition should not matter much in this case since we expect to converge to the same field amplitude, when the same field normalisation (pyro normalisation) is used.

d7919:
It would be good to understand the sqrt(2) in the phinit value.

It does seem to be in pretty good agreement, which is promising. I would guess that the agreement might get better with more velocity space resolution? It might also be nice to show how to agreement carries on to larger theta/kx or with non-zero theta0 (although I appreciate that's not exactly the motivation of the comparison at this point).

Also worth thinking about how to check A|| if all initial conditions essentially have A||=0.

d7919:
> I think a very nice test would be to compare the saturated value of δϕ and δA∥ from GS2 and CGYRO nonlinear simulations. The actual value of the initial condition should not matter much in this case since we expect to converge to the same field amplitude, when the same field normalisation (pyro normalisation) is used.

This is a nice idea but of course relies a little on the codes capturing/treating the physics consistently which we know isn't always the case!

pyrokinetics/gk_code/GKOutputReaderGENE.py
```diff
@@ -340,19 +375,38 @@ def _set_fields(
...
+        local_norm = data.local_norm
+        pyro_flux_units = flux_units(local_norm.pyrokinetics)
+        if gk_input.data["geometry"].get("minor_r", 0.0) == 1.0:
+            gene_flux_units = flux_units(local_norm.pyrokinetics)
+        elif gk_input.data["geometry"].get("major_R", 1.0) == 1.0:
+            gene_flux_units = flux_units(local_norm.gene)
+        else:
+            raise NotImplementedError(
+                "Pyro does handle cases where neither major_R and minor_r are 1.0"
+            )
```
mgiacomin:
```suggestion
            gene_flux_units = flux_units(local_norm.gene)
        elif gk_input.data["geometry"].get("major_R", 1.0) > 1.0:
            gene_flux_units = flux_units(local_norm.pyrokinetics)
        else:
            raise NotImplementedError(
                "Pyro does not handle cases where both major_R is smaller than 1.0 and minor_r is not 1.0"
            )
```
I would suggest this change to force the normalisation to `minor_r` when `major_R > 1`.

bpatel2107: What is happening here is that we are figuring out what the normalisation is, not setting what pyro stores as the normalisation. 

The normalisation gets forced to a pyro convention at the end here 
https://github.com/pyro-kinetics/pyrokinetics/blob/ae52b411beab0cd205c4a40c2d40adfc28830d16/pyrokinetics/gk_code/GKOutputReaderGENE.py#L475-L478


This is just seeing if we need to re-normalise the output or not. If `minor_r=1` then the run already uses the minor radius as `Lref` so we can leave it as is. If `major_R=1` then the major radius was used and we need to re-normalise. That is what is being done here.

We should probably check for `major_R` first and if that isn't 1 check `minor_r`. If that isn't 1 or isn't defined then we should raise a warning of some kind

mgiacomin: Of course this change should be reported everywhere in `GKOutputReadearGENE.py` and not just in the above lines.

bpatel2107: This will cause the program to raise an error but we can put it in the `init_dataset` method so we catch it earlier

mgiacomin: Yes, I think we can also invert the logic and check `major_R` first, but I think we should check if `major_R>1`. 
Maybe it is better in the `init_database` to avoid multiple repetition.

pyrokinetics/gk_code/GKOutputReader.py
```diff
@@ -223,11 +256,18 @@ def _set_eigenvalues(
         data["growth_rate"] = (("kx", "ky", "time"), growth_rate.data)
         data["mode_frequency"] = (("kx", "ky", "time"), mode_frequency)
         data["eigenvalues"] = (("kx", "ky", "time"), eigenvalue)
+
+        eigenvalue_units = eigenvalues_units(data.local_norm.pyrokinetics)
+
+        data = data.pint.quantify(eigenvalue_units)
```
ZedThree: This is a little confusing reusing the function name, maybe just do:

```suggestion
        data = data.pint.quantify(eigenvalues_units(data.local_norm.pyrokinetics))
```

pyrokinetics/gk_code/GKOutputReader.py
```diff
@@ -236,24 +276,28 @@ def _set_eigenfunctions(
...
-            square_fields.sum(dim="field").integrate(coord="theta") / (2 * pi)
-        )
-
-        eigenfunctions = eigenfunctions / field_amplitude
+        for ifield, field in enumerate(data["field"].data):
+            eigenfunctions[ifield, :] = data[field].pint.dequantify()
+            square_fields += np.abs(data[field].pint.dequantify()) ** 2
 
         data["eigenfunctions"] = (
             ("field", "theta", "kx", "ky", "time"),
```
ZedThree: Reuse `coords` from above?

pyrokinetics/gk_code/GKOutputReader.py
```diff
@@ -236,24 +276,28 @@ def _set_eigenfunctions(
...
-            eigenfunctions.data,
+            eigenfunctions,
         )
+        field_amplitude = np.sqrt(square_fields.integrate(coord="theta") / (2 * pi))
+
+        data["eigenfunctions"] *= 1.0 / field_amplitude
```
ZedThree: Try not to modify variables after creation -- constructing them with the final value in the first place usually leads to clearer code as you only need to look in one place to understand the end result. This bit could then be:

```python
        field_amplitude = np.sqrt(square_fields.integrate(coord="theta") / (2 * pi))
        data["eigenfunctions"] = (coords, eigenfunctions / field_amplitude)
```

pyrokinetics/gk_code/GKOutputReader.py
```diff
@@ -278,4 +322,90 @@ def _set_growth_rate_tolerance(data: xr.Dataset, time_range: float = 0.8):
         return data
 
 
+def eigenvalues_units(convention: Convention):
+    """
+    Generates dictionary of flux output units
+    Parameters
```
ZedThree: You'll want a blank line here to render this properly in the docs:

```suggestion

    Parameters
```

And in the routines below.


Turns out we don't actually build docs for the complete package, so I'll add those in a separate PR

pyrokinetics/gk_code/GKOutputReaderCGYRO.py
```diff
@@ -260,59 +290,69 @@ def _set_fields(
...
-                # Get eigenfunction data
-                raw_eig_data = raw_data.get(f"eigenfunctions_{field_name}", None)
-                if raw_eig_data is None:
-                    logging.warning(
-                        f"When setting fields, eigenfunction data for {field_name} not "
-                        f"found, expected the file bin.cygro.{field_name}b to exist. "
-                        f"Setting the field {field_name} to 0."
+                fields = field_data.swapaxes(0, 1)
+            else:
+                # If theta_plot != theta_grid, we get eigenfunction data and multiply by the
```
ZedThree, line 346: Something to consider: is this block large/complex enough to pull out into a separate helper function?

pyrokinetics/gk_code/GKOutputReader.py
```diff
@@ -278,4 +322,90 @@ def _set_growth_rate_tolerance(data: xr.Dataset, time_range: float = 0.8):
         return data
 
 
+def eigenvalues_units(convention: Convention):
```
ZedThree: Nice! I like these helper functions

pyrokinetics/gk_code/GKOutputReaderCGYRO.py
```diff
@@ -116,6 +125,14 @@ def _init_dataset(
             xr.Dataset: Dataset with coords and attrs set, but not data_vars
         """
 
+        # Coordinate units for pyro and cgyro
+        pyro_coord_units = coord_units(local_norm)
+        cgyro_coord_units = coord_units(local_norm.cgyro)
+
+        bunit_over_b0 = (
+            (1 / cgyro_coord_units["ky"]).to(local_norm.pyrokinetics).magnitude
```
ZedThree: Just to note that `.to(local_norm)` converts to the default convention, i.e. `pyrokinetics`, but I guess it's also less explicit

pyrokinetics/gk_code/GKOutputReaderCGYRO.py
```diff
@@ -260,59 +290,69 @@ def _set_fields(
...
+            fields = fields.to(local_norm.pyrokinetics).magnitude
 
-            fields[ifield, :, :, :, :] = field_data.reshape(
-                [data.ntheta, data.nkx, data.nky, data.ntime]
-            )
+            data[field_name] = (coords, fields)
 
-        data["fields"] = (coords, fields)
+        data = data.pint.quantify(pyro_field_units)
         return data
```
ZedThree: Could just return straight away?

```suggestion
        return data.pint.quantify(pyro_field_units)
```

pyrokinetics/gk_code/GKOutputReaderGENE.py
```diff
@@ -128,6 +137,27 @@ def _init_dataset(
         """
         nml = gk_input.data
 
+        pyro_coord_units = coord_units(local_norm.pyrokinetics)
+        if nml["geometry"].get("major_R", 1.0) == 1.0:
+            local_norm.default_convention = "gene"
+        else:
+            try:
```
ZedThree: This whole bit is probably complex and distracting enough that I would pull it out into a helper function, something like `set_default_convention()`

pyrokinetics/gk_code/GKOutputReaderGENE.py
```diff
@@ -417,40 +470,66 @@ def _set_fluxes(
...
+        local_norm = data.local_norm
+        pyro_eigval_units = eigenvalues_units(local_norm.pyrokinetics)
+        if gk_input.data["geometry"].get("minor_r", 0.0) == 1.0:
+            gene_eigval_units = eigenvalues_units(local_norm.pyrokinetics)
+        elif gk_input.data["geometry"].get("major_R", 1.0) == 1.0:
+            gene_eigval_units = eigenvalues_units(local_norm.gene)
+        else:
+            raise NotImplementedError(
+                "Pyro does handle cases where neither major_R and minor_r are 1.0"
+            )
```
ZedThree: Nice, I love a good error message!

pyrokinetics/gk_code/GKOutputReaderGENE.py
```diff
@@ -417,40 +470,66 @@ def _set_fluxes(
...
+                growth_rate.append(float(growth))
 
         last_timestep = [data.time.isel(time=-1)]
         coords = {"time": last_timestep, "ky": kys, "kx": [0.0]}
+        dims = coords.keys()
+
+        mode_frequency = mode_frequency * gene_eigval_units["mode_frequency"]
+        mode_frequency = mode_frequency.to(local_norm.pyrokinetics).magnitude
+        growth_rate = growth_rate * gene_eigval_units["growth_rate"]
+        growth_rate = growth_rate.to(local_norm.pyrokinetics).magnitude
```
ZedThree: Given that we end up doing this pattern a _lot_, it's probably worth writing something like `convert_dataset()`, like `normalisation.convert_dict`, that iterates over the keys (or a given set of them) in the dataset and does `.to(norm).magnitude` on all of them. Actually, if you also passed in the new units, you'd just have this common pattern of:

```python
# set all the fields in data using units for the local convention
...
return convert_dataset(data, norm, units)
```

And in fact, I guess `units` will contain exactly which keys to convert as well

pyrokinetics/gk_code/GKOutputReaderGS2.py
```diff
@@ -166,36 +188,57 @@ def _set_fields(
...
-                fields[ifield, :, :, :, :] = 0
+                field[:, :, :, :] = 0
                 continue
 
             # raw_field has coords (t,ky,kx,theta,real/imag).
             # We wish to transpose that to (real/imag,theta,kx,ky,t)
             field_data = raw_data[field_name].transpose("ri", "theta", "kx", "ky", "t")
-            fields[ifield, :, :, :, :] = sqrt2 * (
+            field = (
                 field_data[0, :, :, :, :] + 1j * field_data[1, :, :, :, :]
```
ZedThree: Just to note that numpy lets you do `...` for "the rest of the indices":

```suggestion
                field_data[0, ...] + 1j * field_data[1, ...]
```

pyrokinetics/gk_code/gk_output.py
```diff
@@ -0,0 +1,224 @@
...
+    within_time_range = difference["time"].data > time_range * final_time
+    tolerance = np.sum(np.where(within_time_range, difference, 0), axis=-1) / np.sum(
+        within_time_range, axis=-1
+    )
+    return tolerance
+
+
+class GKOutput(DatasetWrapper):
+    """
+    GKOutput contains the output data from gyrokinetics codes, and converts it to
```
ZedThree: Something missing here?

pyrokinetics/gk_code/utils.py
```diff
@@ -0,0 +1,23 @@
+from ..units import ureg as units
+
+# Define basic units
+gk_output_units = {
```
ZedThree: Can this just be in `gk_output.py`?

pyrokinetics/normalisation.py
```diff
@@ -604,9 +663,23 @@ def _update_system(self):
             "nref_electron": {str(self.nref): 1.0},
             "tref_electron": {str(self.tref): 1.0},
             "vref_nrl": {str(self.vref): 1.0},
+            "rhoref_pyro": {str(self.rhoref): 1.0},
             "beta_ref_ee_B0": {str(self.beta_ref): 1.0},
         }
 
+    @property
+    def references(self):
```
ZedThree, line 677: Is this used anywhere?

pyrokinetics/normalisation.py
```diff
@@ -448,6 +466,41 @@ def set_lref(
             lambda ureg, x: x.to(ureg.lref_minor_radius).m * self.pyrokinetics.lref,
         )
 
+    def set_rhoref(
+        self,
+        local_geometry: Optional[LocalGeometry] = None,
```
ZedThree, line 477: Could instead just pass `bunit_over_b0` in? Or maybe in a separate PR, completely remove the dependence of this file on `LocalGeometry` and `Kinetics` ? I think that was a mistake on my part

ZedThree (commented):
LGTM, a few minor comments worth considering, but the big one is I think we need a `convert_dataset` function to simplify some of this stuff.

The common pattern I think we should see basically everywhere in pyrokinetics is we create the `thing` with units natural to that `thing`: so a GS2 `thing` uses GS2 units when we create it. Then we convert the whole `thing` to our units in one go. So it looks like:

1. Raw data from some code, without units
2. Read in data, add code's units
3. Convert to our internal representation, convert to our units
4. Complete PK object

pyrokinetics/gk_code/GKOutputReaderGS2.py
LiamPattinson: First time I've seen this used in a real case. [Relevant stack overflow ](https://stackoverflow.com/questions/772124/what-does-the-ellipsis-object-do) for one of the most obscure bits of Python syntax.

pyrokinetics/gk_code/gk_output.py
LiamPattinson: I think the `GKOutput(DatasetWrapper)` object was originally planned to replace the use of raw `xr.Dataset` in this PR but was discarded as it's adding multiple layers of complexity to an already difficult refactor. I'm currently working on a toy example version of `GKOutput(DatasetWrapper)` that could be used as the basis of a future PR, and will hopefully have something ready in the next couple of days. The `convert_dataset` function you recommended in other comments could be implemented on `DatasetWrapper` once I've figured everything out.

pyrokinetics/gk_code/GKOutputReaderGS2.py
bpatel2107: Ooooh that's good to know!

bpatel2107:
So I've added the GKOutputReader for each code. I think the format we have is pretty good but there are some points I'm concerned about.

For example at the moment we **have** to provide fields and fluxes when we instantiate a `GKOutput` object. We should probably allow for cases where we don't need to load everything in. Probably just need to make them optional and have a default empty dict.

Currently in `GKOutput` we specify the shape of each output. However, there are a few cases where codes have different dimensions/coordinates of specific outputs, for example GENE doesn't output fluxes as a function of `ky` and TGLF has additional modes which require a dimension to capture. Right now its not too bad but this could become quite messy. I think we could maybe add in the option to set these in each code and pass it when instantiating `GKOutput`

Otherwise happy for a review to be done. @LiamPattinson @ZedThree do you have any time soon to review

pyrokinetics/examples/example_read_gene_output.py
```diff
@@ -33,39 +33,22 @@
...
-apar = (
-    data["fields"]
-    .sel(field="apar")
-    .isel(ky=0)
-    .isel(kx=0)
-    .sel(theta=0.0, method="nearest")
-)
-apar = np.abs(apar)
-apar.plot.line(x="time")
+if "apar" in data:
```
LiamPattinson, line 48: This line fails for me, and the error seems to be in the `dataset_wrapper` class. I fixed it by adding the following function:

```python
# class: dataset_wrapper
def __contains__(self, name: str) -> bool:                                          
    return name in self.data
```

Let me know if I should push this change to your branch or raise a new PR etc.

pyrokinetics/examples/example_read_gs2_output.py
```diff
@@ -18,10 +18,9 @@
 mode_freq = data["mode_frequency"]
 
 # Plot growth and mode frequency
-
-growth_rate_tolerance = data["growth_rate_tolerance"].values[0, 0]
+growth_rate_tolerance = data.growth_rate_tolerance()[-1]
```
LiamPattinson: I think this can just be `growth_rate_tolerance = data.growth_rate_tolerance`

pyrokinetics/examples/example_read_tglf_linear.py
LiamPattinson, line 1: This fails for me after the first plot, saying `apar` isn't a field for the eigenfunctions.

pyrokinetics/examples/example_read_tglf_transport.py
```diff
@@ -1,7 +1,9 @@
-from pyrokinetics import Pyro
+from pyrokinetics import Pyro, template_dir
 import matplotlib.pyplot as plt
 
-pyro = Pyro(gk_file="input.tglf", gk_code="TGLF")
+gk_file = template_dir / "../../../pyro_test/TGLF_transport/input.tglf"
```
LiamPattinson: Should this file be included in the repo?

pyrokinetics/gk_code/__init__.py
```diff
@@ -1,19 +1,27 @@
...
+from .gk_output import GKOutput, supported_gk_output_types  # noqa
+from .GKOutputReaderGS2 import GKOutputReaderGS2  # noqa
+from .GKOutputReaderCGYRO import GKOutputReaderCGYRO  # noqa
+from .GKOutputReaderGENE import GKOutputReaderGENE  # noqa
+from .GKOutputReaderTGLF import GKOutputReaderTGLF  # noqa
+
+# from . import GKOutputReaderGS2
+# from . import GKOutputReaderCGYRO
+# from . import GKOutputReaderGENE
+# from . import GKOutputReaderTGLF
```
LiamPattinson: I think these comments can be deleted

pyrokinetics/gk_code/gk_output.py
```diff
@@ -0,0 +1,646 @@
...
+                if np.shape(field) != (len(theta), len(kx), len(ky), len(time)):
+                    raise ValueError(f"field '{name}' has incorrect shape")
+
+        flux_units = get_flux_units(convention)
+        for flux_type, flux in fluxes.items():
+            fluxes[flux_type] = _renormalise(
+                fluxes[flux_type], convention, flux_units[flux_type]
+            )
+            # check dims
+            if gk_code == "GENE":
```
LiamPattinson: I'm not a fan of checking the `gk_code` and hardcoding what the field/flux dims should be, as `GKOutput` should be independent of any particular code, and it should be possible to add a new code without modifying this class. Maybe we should just remove the checks on the dimensions? We could add some lines to the docs explaining which dimensions can be included and in what order, and specify that it's okay to drop dimensions if a particular code doesn't support them.

pyrokinetics/gk_code/GKOutputReaderCGYRO.py
```diff
@@ -407,33 +483,101 @@ def _set_eigenfunctions(
...
+        square_fields = np.sum(np.abs(eigenfunctions) ** 2, axis=0)
+        field_amplitude = np.sqrt(np.trapz(square_fields, coords["theta"], axis=0)) / (
+            2 * np.pi
+        )
+        result = eigenfunctions / field_amplitude
+
+        return result
+
+    @staticmethod
+    def to_netcdf(self, *args, **kwargs) -> None:
```
LiamPattinson: I'm not sure why the `to_netcdf` and `from_netcdf` functions are defined here (similarly for other reader classes). It looks to me like they're just handling complex data for netcdf IO, which is something we should probably be doing in `dataset_wrapper`  by default. Could we just move these lines there? If we only want this behaviour in `GKOutput`, we could define these functions there instead (but also use `super()` to avoid repeat code if possible).

(staticmethods also shouldn't have a `self` arg, as in this context `self` will just be a standard function argument that you can pass in, and it doesn't operate on a particular instance of the class. I think `@staticmethod` just needs to be removed in this case)

LiamPattinson (commented):
Looks good to me besides a couple of points.

I didn't realise the dimensionality of the data would vary as much as it does, so perhaps the tight constraints on units/dims should just be removed from `GKOutput.__init__` and the docstring. A better way to handle this might just be to have the user provide field/flux/eigenvalue/eigenfunction coords as arguments to `GKOutput.__init__` (e.g. `GKOutputReaderTGLF` could pass in `field_coords = ("ky", "mode")`, but `field_coords = ("theta", "kx", "ky", "time")` would be the default). We could still check that the data they've provided has the correct shape according to the coords they pass in.

A problem for a further PR: after seeing the points raised in Issue https://github.com/pyro-kinetics/pyrokinetics/issues/199, I'm thinking that the `DatasetWrapper` approach might not work in the long term. I'd imagined it could be used to add some nice features to `Equilibrium`, `Kinetics` and `GKOutput`, giving us an Xarray interface to those data structures while letting us define our own constraints on the data (dims, units, etc) and functions on the classes. The problem is, the way we're using Xarray/pint causes it to load the entire dataset into memory without any lazy loading, so pyrokinetics can't handle larger-than-memory datasets. We'll probably need to move to the Dask backend for Xarray to get this to work for large datasets, but that may warrant a complete rethink of how `GKOutput` works, and it may need to be decoupled from `DatasetWrapper`.

pyrokinetics/examples/example_read_gene_output.py
bpatel2107: I've added that here!

pyrokinetics/gk_code/GKOutputReaderCGYRO.py
bpatel2107: Not sure why I did that originally but I've moved it to `dataset_wrapper` now

bpatel2107:
So I've got updated this and addressed the comments. I've made some additional changes here too

- The moments of the distribution function have been added to `GKOutput` but has only been set up for CGYRO
- You can choose what data out of `fields`, `fluxes` and `moments` are loaded
- The size of different outputs are now specified directly in the individual code readers rather than `GKOutputReader`
- The linear fluxes/moments are now normalised as is done in GKDB (correct quasilinear flux calculations)
- All examples work except for `example_real_space_moment` as I don't have the moment data for it yet
- Revert back to using `to_netcdf` and `from_netcdf` in `dataclass_wrapper`

Hopefully ready to merge 

Failing due to scipy issues noted in #210

mgiacomin:
@bpatel2107  I have added moment files to the CGYRO output data in template.

LiamPattinson:
Haven't had a chance to look through everything in full, but it looks like some docs/typehints might need updating. I should have time this afternoon to make a small PR.

LiamPattinson:
Update on my own PR: [it's taking a little longer than I'd expected](https://en.wikipedia.org/wiki/Hofstadter%27s_law), hopefully should have it done within a couple of days. Sorry for holding things up again!

LiamPattinson (approved):
I'm happy with where things are, though it might be worth getting a third opinion since this is such a massive update.

bpatel2107:
So overall this is ready to merge in terms of structure.

The only thought we should have is about the fluxes/moments that we store.

Currently we have `particle`, `heat` and `momentum` (parallel) fluxes. Are we expecting to want perpendicular/total momentum flux? Can always add that in later

Similar for the moments we have `density`, `temperature` and `velocity` right now. Do we want to be able to break down the moments as parallel and perpendicular temperature moments? Similarly do we want temperature or energy (standard CGYRO output) fluctuations 

Below are the outputs of GENE for example 

image
image


@d7919 @mgiacomin @dake0795  @hgd511 @fcasson thoughts?

LiamPattinson:
Would it be possible to add the extra fluxes/moments in a separate PR? That way we could also start work updating the docs for this after merging PR https://github.com/pyro-kinetics/pyrokinetics/pull/203, and I think @arkabokshi was waiting for this to be merged before raising a PR for his GKW work.

bpatel2107:
I think we can merge for now. The fluxes aren't likely to change but the moments might. We can cross that bridge when we come to it though. Will raise a separate issue  on this

PR #184, opened 2023-03-28
Refactor Numerics

LiamPattinson:
Some low-hanging fruit, following the discussion we had a few weeks ago. `Numerics` works by taking a dict to its `__init__` and absorbing everything given to it. If there's a typo in the code for a necessary key (say, `n_theta` instead of `ntheta`), it will quietly add a default value of `ntheta=32` and also store `n_theta=whatever`. We could catch this in testing, but it's easier to just enforce a reduced set of possible inputs in the first place.

This PR refactors `Numerics` as a [dataclass](https://docs.python.org/3/library/dataclasses.html), which gives it a nice `__init__` for free which will catch any stray `kwargs` and raise an error.

## Other features:

- Uses `__getitem__`/`__setitem__` for dict-like access.
- Can be written to/from json.
- Has a 'pretty format' method, works when you call `str(numerics)` or `print(numerics)`.
- Contains metadata that allows you to identify which Pyrokinetics session generated each object.
- Improved docs.

## Issues:

- Docs could use some further improvements. @bpatel2107 If you have time, please could you look over the docstrings/annotations (comments written as `#:`) and correct any mistakes or misunderstandings in there? I used the GS2 docs for reference, but it's likely that I've written something incorrect in there.
- Does it make sense for everything to have a default value?
- Needs a couple of tests for the json bits. I'll get on that tomorrow morning.

### DIFF
--- a/pyrokinetics/metadata.py
+++ b/pyrokinetics/metadata.py
@@ -1,0 +1,44 @@
+import uuid
+from datetime import datetime
+from typing import Dict
+
+try:
+    from importlib.metadata import version, PackageNotFoundError
+except ModuleNotFoundError:
+    from importlib_metadata import version, PackageNotFoundError
+try:
+    __version__ = version(__name__)
+except PackageNotFoundError:
+    try:
+        from setuptools_scm import get_version
+
+        __version__ = get_version(root="..", relative_to=__file__)
+    except ImportError:
+        __version__ = "0.0.1"
+
+
+# Define UUID and session start as module-level variables.
+# Determined at the first import, and should be fixed during each session.
+
+__uuid = uuid.uuid4()
+__session_start = datetime.now()
+
+
+def metadata(**kwargs: str) -> Dict[str, str]:
+    """
+    Return a dict of metadata uniquely describing this Pyrokinetics session. Should be
+    written to Pyrokinetics output files to establish their provenance.
+
+    Parameters
+    ----------
+    **kwargs: str
+        Extra keywords to add to the returned dict.
+    """
+    return {
+        "software_name": "Pyrokinetics",
+        "software_version": __version__,
+        "session_started": str(__session_start),
+        "session_uuid": str(__uuid),
+        "date_created": str(datetime.now()),
+        **kwargs,
+    }
--- a/pyrokinetics/metadata.py
+++ b/pyrokinetics/metadata.py
@@ -7,7 +7,7 @@ try:
 except ModuleNotFoundError:
     from importlib_metadata import version, PackageNotFoundError
 try:
-    __version__ = version(__name__)
+    __version__ = version("pyrokinetics")
 except PackageNotFoundError:
     try:
         from setuptools_scm import get_version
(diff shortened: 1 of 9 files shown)


LiamPattinson:
I also had to shuffle around the metadata-generating bits from `DatasetWrapper` so I could reuse them here, but otherwise the code is barely changed.

pyrokinetics/numerics.py
```diff
@@ -1,58 +1,151 @@
...
+    #: Number of elements in the velocity-space :math:`k_x` grid
+    nkx: int = 1
+
+    #: Value of :math:`k_y\rho`
+    ky: float = 0.1
+
+    #: Value of :math:`k_x\rho`
+    kx: float = 0.0
+
+    #: Initial time step, in units of ``tref``
```
bpatel2107: Units of time step would be in ``lref / vref``

bpatel2107:
```suggestion
    #: Initial time step, in units of ``lref / vref``
```

pyrokinetics/numerics.py
```diff
@@ -1,58 +1,151 @@
...
+    delta_time: float = 0.001
+
+    #: Time step, in units of ``tref``
```
bpatel2107:
```suggestion
    #: Time step, in units of ``lref / vref``
```

bpatel2107:
Overall looks good, that `dataclasses` is nifty! I think allowing for non-standard inputs like `n_theta` is a bit risky as someone may think they are changing a value when they aren't. Is there a way for us to at least warn the user? Maybe in the `__setitem__` have a warning for non-standard inputs which read standard ones from a list?

pyrokinetics/metadata.py
```diff
@@ -0,0 +1,44 @@
...
+        __version__ = get_version(root="..", relative_to=__file__)
+    except ImportError:
+        __version__ = "0.0.1"
+
+
+# Define UUID and session start as module-level variables.
+# Determined at the first import, and should be fixed during each session.
+
+__uuid = uuid.uuid4()
+__session_start = datetime.now()
```
ZedThree: I think these probably want to be instance-level variables on `Pyro`, rather than globals. Actually, probably the time wants to be set when we write files too

LiamPattinson:
> Overall looks good, that dataclasses is nifty! I think allowing for non-standard inputs like n_theta is a bit risky as someone may think they are changing a value when they aren't. Is there a way for us to at least warn the user? Maybe in the __setitem__ have a warning for non-standard inputs which read standard ones from a list?

It took some dark magic, but I've implemented a way to prevent assigning to the wrong keys/attributes. This would be much easier if our minimum version were 3.10, as then we could use `@dataclasses.dataclass(slots=True)` which would achieve the same thing.

pyrokinetics/metadata.py
LiamPattinson: My intent with those globals was to use them to identify which Python sessions created each object. I'm imagining a scenario where a bunch of files are generated in a parameter scan of some sort, which may involve multiple `Pyro` objects (or no Pyro objects at all), and these UUIDs/session starts can be used to tie them all together. I think a `pyro_uuid` would also be helpful, but `Pyro` doesn't use `metadata` yet, and I think that'd be best achieved in another PR. I'm currently not sure how a `Pyro` could pass this down to every object it handles, and it might require changing constructors/function signatures throughout the code.

LiamPattinson: Actually, I may need to rethink how this works altogether. Currently the metadata is set when an `Equilibrium` or `Numerics` is created, and the same would be pushed to `GKOutput`/`Kinetics`/`LocalGeometry` if they adopt `DatasetWrapper` as a base class. If it'd be better to only add this when we write to files, and if we'd need it to be handled (exclusively?) by `Pyro`, then the current system just won't work. Could this be handled later in a different Issue/PR?

pyrokinetics/numerics.py
LiamPattinson: Cheers, fixed now

pyrokinetics/metadata.py
ZedThree: Yes, we can put it off for now. Something is better than nothing, but it's probably going to require a fair bit of thought to get this right

bpatel2107 (approved):
LGTM

Issue #10, opened 2021-07-05
TODO: Make pyrokinetics IMAS compliant


bpatel2107:
Make an interface between pyrokinetics and IMAS to further diversify the application of pyrokinetics

ZedThree:
[OMAS](https://gafusion.github.io/omas/index.html) might be useful here

fcasson:
Or GKDB:  https://gitlab.com/gkdb/gkdb/-/wikis/home

bpatel2107:
Just bumping this issue. 

We should discuss whether we want to switch over our naming conventions to match imas

bpatel2107:
Done in #113

PR #185, opened 2023-04-03
Enhancements/kinetics units

bpatel2107:
I've added in units to the `Kinetics` object and then modified the `Species` and `LocalSpecies` to take units from `Kinetics` directly. The object needs to be populated with data in the following format as indicated by the docstriing

```python
"""
    - psi_n       [units] dimensionless
    - r/a         [units] dimensionless
    - Charge      [units] elementary_charge
    - Mass        [units] kg
    - Temperature [units] eV
    - Density     [units] meter**-3
    - Rotation    [units] meter/second
"""
 ```

### CODE REVIEW

File: pyrokinetics/constants.py [hunk 1]
```diff
@@ -1,15 +1,16 @@
...
 import numpy as np
+from .units import ureg as units
 
 bk = constants.k
 pi = constants.pi
 mu0 = constants.mu_0
 eps0 = constants.epsilon_0
 
-electron_charge = constants.elementary_charge
+electron_charge = constants.elementary_charge * units.elementary_charge
```
ZedThree, line 10: I think `units.elementary_charge` is the value + units -- worth double checking at least. I think electron and proton mass are in there too. In fact, we could probably just use those directly?

bpatel2107: It looks like it is just the value. Not sure scipy supports pint units yet

bpatel2107:
So I think that we could potentially add units to the gradients below such that they use `lref_minor_radius` so we have $a/L_T$. This would allow for converting to different normalisations like $R_{maj}/L_T$. My main concern is that if someone were to do `local_species.a_lt.to_base_units` then it would convert this into something with units of length which is not really the case.

Is there a way to add a unit but would keep it dimensionless? Or is the name `a_lt` just not really correct and we should call it is something like `inverse_lt`. @ZedThree thoughts?

https://github.com/pyro-kinetics/pyrokinetics/blob/d0bc035100d3ba145faea757acb510a4075a7adb/pyrokinetics/species.py#L58-L83

bpatel2107:
I've added in units to the kinetic gradients such that it would now be possible to change from $a/L_T$ to $R_{maj}/L_T$ using pint. I thought keeping the variable name as `local_species.electron.a_lt` didn't quite right as this term now has units of `1/lref` so I changed it to `inverse_lt` which will break existing code. We seem to be going through some major changes so now seems like a good time to do this.

@arkabokshi FYI

ZedThree:
If we're changing the names, is it worth changing straight to the IMAS names? They call them `density_log_gradient_norm` etc. Bit of a mouthful, but that's what tab-completion is for :)

bpatel2107:
Well the question then becomes do we just change everything?

bpatel2107:
Happy to discuss this further as it may be useful in the long run! Will probably keep this as a separate PR though (mentioned in #10 as a reminder) so but this is probably a separate issue for now. Given that the code is going to change quite a bit over the next few months I think its best to get these changes in so we don't have lots of open PRs lying around

PR #186, opened 2023-04-06
Example GKOutput Implementation

LiamPattinson:
This PR is a demonstration of how we might implement a `GKOutput` class for PR https://github.com/pyro-kinetics/pyrokinetics/pull/182. If we'd prefer not to change the existing code structure too much at this point, we can hold back these changes for a future PR. I'll be on leave for a week and I'm expecting to be busy with other projects after that, so I'll have very limited time available in the near future to push these updates throughout the code.

`GKOutput` is implemented as a `DatasetWrapper`, which functions similarly to the [`UserDict`](https://docs.python.org/3/library/collections.html#collections.UserDict) Python class. It allows us to implement a custom `__init__` function and other utilities on top of an Xarray Dataset, though without directly subclassing Dataset (something discouraged by the Xarray maintainers). `DatasetWrapper` also handles `pint-xarray` stuff behind the scenes, and provides functions for writing to/reading from netcdf. Within the `__init__` I've implemented:

- Units are changed to pyrokinetics convention. We won't need to do this in each `GKOutputReader`.
- Dimensions of the inputs are checked
- Eigenvalues are created from field data if the user hasn't provided them
- Eigenfunctions are created from field data if the user hasn't provided them

I've also moved the growth rate tolerance function here, and provided `field` and `flux` utility functions. In future we could move output diagnostics to this class and/or add plotting functionality, similarly to `Equilibrium`.

I've only implemented the changes for GS2. The other gk codes have not been updated to match, nor have the tests (hence why they're failing catastrophically). The changes to `GKOutputReaderGS2` are:

- No longer creates a dataset. Just creates Numpy arrays and passes them to `GKOutput`.
- Only applies units at the end, using the GS2 convention.
- Inherits from the base `Reader` class, intermediate abstract class `GKOutputReader` not in use. A similar change was made to equilibrium reader classes, as I found that the intermediate abstract classes impose annoying restrictions on how the concrete classes can be implemented without adding much functionality themselves.

To test it, I've been dropping into the directory `pyrokinetics/templates/outputs/GS2_linear` and calling the following:

```python
>>> import pyrokinetics as pk
>>> pyro = pk.Pyro(gk_file="gs2.in")
>>> pyro.load_gk_output()
>>> pyro.gk_output # access the GKOutput object and manually check that the contents look correct
```

Sorry that I'm not able to achieve more at this stage! I might be able to work on this further after I'm back from leave, but I won't be able to dedicate as much time as I have been recently. Feel free to add stuff to this branch if you want to take it further, but I'd recommend checking my changes very carefully as there's a good change I've broken something somewhere.


bpatel2107:
I'm also away for the next week so will be unable to work on this until I get back. Can look into this further then

bpatel2107:
I've updated each of the codes to use the updates `gk_output` reader. The tests still need updating but it all works in theory.

There are some points I want to think about. Different codes may have different dimension for certain outputs. For example most codes have the flux as a function of `ky`, but GENE only has it summed over `ky`. This means when checking the shape of the array we need to add code specific lines. Maybe we can add the shape of the output as a kwarg to `GKOutput`

On leave for the next week and a half so will get back to this when I can.

bpatel2107:
Will merge this into the other branch and leave big review for there

PR #196, opened 2023-04-26
Enhancements/ion normalsation inputs

bpatel2107:
Re-normalised GK input files such that the electron temp and density is the reference values, in line with the units applied to `LocalSpecies`. 

Particularly relevant if an ion species has been used as the normalising species. Working out which species is being used is a bit longer (would likely require guessing given the charges and mass) but this method basically guarantees that we get the electron as the normalising speices.

Requires re-normalising density, temperature and beta.

### DIFF
--- a/pyrokinetics/gk_code/GKInputCGYRO.py
+++ b/pyrokinetics/gk_code/GKInputCGYRO.py
@@ -633,16 +633,22 @@ class GKInputCGYRO(GKInput):
         return is_basic_miller
 
     def get_ne_te_normalisation(self):
-        # Get electron temp/dens first
-
+        found_electron = False
         if self.data.get("AE_FLAG", 0) == 1:
             ne = self.data["DENS_AE"]
             Te = self.data["TEMP_AE"]
+            found_electron = True
         else:
             for i_sp in range(self.data["N_SPECIES"]):
                 if self.data[f"Z_{i_sp+1}"] == -1:
                     ne = self.data[f"DENS_{i_sp+1}"]
                     Te = self.data[f"TEMP_{i_sp+1}"]
+                    found_electron = True
                     break
+
+        if not found_electron:
+            raise TypeError(
+                "Pyro currently requires an electron species in the input file"
+            )
 
         return ne, Te
--- a/pyrokinetics/gk_code/GKInputCGYRO.py
+++ b/pyrokinetics/gk_code/GKInputCGYRO.py
@@ -243,7 +243,8 @@ class GKInputCGYRO(GKInput):
 
         # Assume pref*8pi*1e-7 = 1.0
         # FIXME Should not be modifying miller after creation
-        beta = self.data["BETAE_UNIT"]
+        ne_norm, Te_norm = self.get_ne_te_normalisation()
+        beta = self.data["BETAE_UNIT"] * ne_norm * Te_norm
         if beta != 0:
             miller.B0 = 1 / (miller.bunit_over_b0 * beta**0.5)
         else:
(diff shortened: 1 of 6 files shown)


pyrokinetics/gk_code/GKInputCGYRO.py
```diff
@@ -623,3 +631,13 @@ def _check_basic_miller(self):
                 break
 
         return is_basic_miller
+
+    def get_ne_te_normalisation(self):
+        # Get electron temp/dens first
+        for i_sp in range(self.data["N_SPECIES"]):
+            if self.data[f"Z_{i_sp+1}"] == -1:
```
d7919: What about adiabatic electron runs?

bpatel2107: Good point, will add a flag!

bpatel2107: @dake0795 in the case where adiabatic electrons are used in GENE, what is the temp/dens they have? I'm guessing density comes from quasi-neutrality but what about the temp?

dake0795: Density is set by quasi-neutrality. I believe the temp is set to be the same as the kinetic species.

bpatel2107: In the scenario where there are multiple ion species with different temperatures and adiabatic ions (not sure how likely this is) then which temp would be taken?

dake0795: The temp of the first kinetic species.

d7919: Also this assumes electrons have Z=-1 I guess this may be true for CGYRO?

pyrokinetics/gk_code/GKInputGS2.py
```diff
@@ -549,3 +559,14 @@ def set(
 
         for name, namelist in self.data.items():
             self.data[name] = convert_dict(namelist, local_norm.gs2)
+
+    def get_ne_te_normalisation(self):
+        # Load each species into a dictionary
+        for i_sp in range(self.data["species_knobs"]["nspec"]):
+            gs2_key = f"species_parameters_{i_sp + 1}"
+            if self.data[gs2_key]["z"] == -1:
```
d7919: In GS2 we are free to set the normalising charge so this won't necessarily always find the electrons.

bpatel2107: Oh god why would someone do that.

I guess we can use the species type instead?

pyrokinetics/gk_code/GKInputCGYRO.py
bpatel2107: Definitely true for CGYRO and TGLF. I think its true in GENE, see excerpt from manual

image

pyrokinetics/gk_code/GKInputGS2.py
d7919: The only reasons I've come across so far are:
1. Testing (check choice of normalisation changes result in expected way).
2. To allow one to change the ion and electron direction convention (I'd guess at least, picking q_electron = 1 might give ion frequencies as negative, matching other codes).

Unfortunately the species type doesn't promise this either really as it only influences the collision operator and is probably really meant to mean "light species". I think it's not unreasonable to use this though. You might be able to do something like, check if there are species with both signs of charge, if so take the one with the smallest mass as the electrons. If only one sign of charge then either fall back to type/adiabatic_option if only one kinetic specie or assume adiabatic electron if >1 kinetic specie? All getting quite complicated so perhaps pyro should just make a clear assumption.

bpatel2107: Ive made a check for species and type and raise an error if the electron isnt found

d7919 (approved):
Looks good to me. Only question I have is if it's possible to add some form of test.

bpatel2107:
I can add a test when there is no electron species and get it to raise an error

bpatel2107:
Added a test, where I write a input file without an electron species and then try to read that file in again

PR #203, opened 2023-06-26
Reorganise and expand documentation

ZedThree:
In preparation for the docathon, rearrange the docs, add stub sections, try to sort out all the minor quibbles, warnings, and errors.

Still need to do:

- fix lots of RST formatting, mostly maths markup
- fix lots of duplicate object description warnings
  - I think this is mostly from explicitly documenting `Attributes` in class docstrings, not quite sure best way to fix this in general


ZedThree:
There's some failing tests for some versions of python -- I've not touched any code, just docstrings and rst files, so this seems spurious.

```
FAILED tests/local_geometry/test_fourier.py::test_load_from_eq - AssertionError: shift difference: 0.04441850862255228
assert False
 +  where False = <function allclose at 0x7f62e64e0b30>(0.22851126700198046, 0.18409275837942818, rtol=0.001)
 +    where <function allclose at 0x7f62e64e0b30> = np.allclose
FAILED tests/local_geometry/test_fourier_cgyro.py::test_load_from_eq - AssertionError: aR difference: [8.46300070e-08 8.31583997e-07 2.86840035e-06 6.25578335e-06
   1.09652848e-05 1.70182132e-05 2.43833169e-05 3.29975940e-05
   4.29233702e-05 5.41189327e-05 6.64960413e-05 8.00597232e-05
   9.47956568e-05 1.10638419e-04 1.27560514e-04 1.45534415e-04]
assert False
 +  where False = <function allclose at 0x7f62e64e0b30>(array([ 2.63000047e+00,  1.13517464e+00,  1.59739427e-01, -1.42496368e-01,\n       -1.40991905e-02,  4.91552812e-02, -4.04506621e-03, -2.14264426e-02,\n        6.39614136e-03,  9.99431632e-03, -5.57916740e-03, -4.57230854e-03,\n        4.31106823e-03,  1.90740097e-03, -3.07430189e-03, -5.48380615e-04]), [2.63000039, 1.13517381, 0.159736559, -0.142502624, -0.0141101558, 0.049138263, ...])
 +    where <function allclose at 0x7f62e64e0b30> = np.allclose
FAILED tests/local_geometry/test_mxh.py::test_load_from_eq - AssertionError: bunit_over_b0 difference: 0.00049174217029746
assert False
 +  where False = <function allclose at 0x7f62e64e0b30>(3.572443446778996, 3.5719517046086984)
 +    where <function allclose at 0x7f62e64e0b30> = np.allclose
```

Are these known failures?

bpatel2107:
Hmm I don't recognise these and the fact that it works for 3.8 but not 3.9 or 3.10 is suspicious. The results look close so maybe some default value changed in one of the fitting algorithms used?

ZedThree:
Most of the warnings now fixed, remaining ones are a little stubborn.

In this PR, there are individual stub files for each module and class, created using `sphinx-apidoc`. This needs to be run manually to add files for new modules/classes. A better, more automated way would be to use [`autosummary`](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html), however this doesn't like that we often name files the same as the class. Some of these are easy fixes (e.g. `kinetics/Kinetics.py` should be moved into `kinetics/__init__.py`), others less so. Worth thinking about!

ZedThree:
Oops, broke the examples tests because I've moved them out of `pyrokinetics/examples` into `docs/examples` -- I think this is probably a better place anyway. Will fix the tests

ZedThree:
API docs currently aren't building correctly because a bug in `omas` means it can't be imported on RTD. It has been fixed (OMAS \#245), but there's been no new release since then.

bpatel2107:
Shall we raise an issue on the OMAS github?

ZedThree:
Yeah, I think they just need to make a new release. I might also offer to tidy up their build system while I'm at it

LiamPattinson:
OMAS is quite a large dependency, but I think we're only using it to check the COCOS of equilibria. Maybe we could reimplement those parts in FreeQDSK and remove it as a dependency?

LiamPattinson:
@ZedThree Is there anything else to do for this PR? If not, I'd be happy to go ahead with it.

ZedThree:
I think there may be still a few warnings I need to squash, but other than that it's basically ready.

ZedThree:
There are a few remaining warnings I've not been able to fix, and they're basically due to a bug in `pint`. Possibly fixed in the latest version, but we'd need to rejig all the normalisation stuff to get it work, so just putting that off for now.

This is ready now

bpatel2107 (approved):

PR #209, opened 2023-07-03
Tests: Run example in /tmp

ZedThree:
`test_jetto_ffprime` in `test_metric_terms` creates a directory, which by default is in the current directory where pytest was run. This uses the `tmp_path` fixture to run it in pytest's `/tmp` directory


bpatel2107 (approved):
LGTM

Issue #210, opened 2023-07-03
Some failing equilibrium tests in SciPy 1.11


ZedThree:
SciPy 1.11 was released recently and this caused a few tests to start failing:

```
FAILED tests/local_geometry/test_fourier.py::test_load_from_eq - AssertionError: shift difference: 0.04441703134647415 (19.437693%)
FAILED tests/local_geometry/test_fourier_cgyro.py::test_load_from_eq - AssertionError: aR difference: [8.46300070e-08 8.31583997e-07 2.86840035e-06 6.25578335e-06
FAILED tests/local_geometry/test_mxh.py::test_load_from_eq - AssertionError: bunit_over_b0 difference: 0.0004917421601211558 (0.013765%)
```

Installing `scipy==1.10.1` makes these tests pass. There was a small internal change to `InterpolateUnivariateSpline` that I suspect has resulted in a small numerical differnce.

Something a bit more concerning though is that these tests all use the same `test.geqdsk` equilibrium file, but they have different expected values for some of the quantities that look like they're passed straight through, e.g. `test_mxh` has `"shift": -0.5766834602024067`, but `test_fourier` has `"shift": 0.18409275837942818`. These even have different signs!

Ideally we'd also reduce the duplication here to make sure they're all in sync

bpatel2107:
Shall we pin the version to `<1.10.1`?

ZedThree:
Good idea, definitely the easiest thing to do for now till we work out a better fix

bpatel2107:
Well it turns out that `shift` shouldn't even be in `FourierGENE` so that's just plain wrong! I can go and fix that. luckily nobody uses that and pyro can't actually write those files yet so it won't have impacted anyone.

PR #211, opened 2023-07-04
Pin scipy to 1.10.1

ZedThree:
Temporary fix for failing tests while we investigate actual issue


LiamPattinson:
Could one solution to this problem be to update the local geometry tests to perform simple sanity checks on the fitted curves instead of golden answer tests on the fitted parameters? The fourier geometry tests have been unstable like this for quite a long time.

bpatel2107 (approved):

bpatel2107:
Will merge for now but we can move this discussion to #210

PR #213, opened 2023-07-06
WIP: Enhancements/gk output normalisation tweaks

LiamPattinson:
I'd intended to do a quick clean up of some docs and type hints, but this turned into a much bigger edit than I'd originally hoped for. I'd noticed that the helper classes `FieldDict`, `FluxType` (later renamed `FluxDict`) and `MomentDict` were non-functional, and the number of `__init__` arguments to `GKOutput` was becoming unreasonable. The docs and type hints also didn't match the code in some places.

I tried to improve matters by defining some helper dataclasses that could be used to organise the inputs to `GKOutput` and also wrap up some of the units handling: `Coords`, `Fields`, `Fluxes`, `Moments`, `Eigenvalues`, and `Eigenfunctions`. These classes:

- Define the names of each entry and whether they're optional
- Define the units for each entry
- The function `.with_units()` automatically adds units to the contents and converts across normalisations.
- The `dims` attribute can be set for GK codes that have output data with non-default dimensions.

Although the definitions of these classes add quite a few lines of code, it reduces the complexity within `GKOutput.__init__` and the `read` functions for each reader class. Very few changes were otherwise needed to the reader classes.

Current minor issues that would be nice to fix:
- A lot of the functions in these classes look very similar, so there's probably an inheritance trick I could use to reduce repeat code.
- There's a bit too much `getattr` and `setattr` going on for my liking. Could do with implementing a dict-like interface.

Current major issues that are causing me to lose my sanity:
- Failing a bunch of golden answer tests, but I have no idea what changes could have led to this. The differences are severe enough that I'll have to fix it before this will be ready to merge.

### DIFF
--- a/pyrokinetics/dataset_wrapper.py
+++ b/pyrokinetics/dataset_wrapper.py
@@ -164,13 +164,11 @@ class DatasetWrapper:
         Writes self.data to disk. Forwards all args to xarray.Dataset.to_netcdf.
         Complex data is expanded out into float arrays of shape ``[dims..., 2]``.
         """
-        # TODO Expand complex numbers selectively
-        #     Currently expands all data, regardless of type, doubling memory
-        #     requirements for non-complex data.
-        data = self.data.expand_dims("ReIm", axis=-1)  # Add ReIm axis at the end
-        data = xr.concat([data.real, data.imag], dim="ReIm")
-
-        data.pint.dequantify().to_netcdf(*args, **kwargs)
+        data = self.data.pint.dequantify()
+        for key, data_var in data.data_vars.items():
+            if data_var.dtype == complex:
+                data[key] = xr.concat([data_var.real, data_var.imag], dim="ReIm")
+        data.to_netcdf(*args, **kwargs)
 
     @classmethod
     def from_netcdf(
@@ -237,12 +235,10 @@ class DatasetWrapper:
         # Set up attr_units
         attr_units_as_str = literal_eval(dataset.attribute_units)
         instance._attr_units = {k: ureg(v).units for k, v in attr_units_as_str.items()}
-        attrs = instance.attrs
 
-        # isel drops attrs so need to add back in
-        instance.data = instance.data.isel(ReIm=0) + 1j * instance.data.isel(ReIm=1)
-        instance.data.attrs = attrs
-        # FIXME Roundtrip converts everything to complex, regardless of the original
-        #       data type.
+        # Recombine ReIm dim
+        for key, data_var in instance.data_vars.items():
+            if "ReIm" in data_var.dims:
+                instance.data[key] = data_var.isel(ReIm=0) + 1j * data_var.isel(ReIm=1)
 
         return instance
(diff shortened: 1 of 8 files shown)


LiamPattinson:
Never mind, caught the bug that was causing tests to fail. Now just need to prettify things and add a couple of tests. Should have that done tomorrow morning.

LiamPattinson:
I've cleaned up most of the repeat code by introducing a base helper class `GKOutputArgs` that each of the concrete helper classes derive from, although the result still isn't as neat as I'd like. I also added a dict-like interface so it's easier to use the helper classes without needing a load of `getattr` and `setattr` calls. Finally, added some simple tests -- let me know if you want anything more rigorous.

Let me know what you think of all this! I'm concerned that this all feels a little overengineered, but I think that these helper classes really simplify matters at point of use, particularly with assigning/renormalising units and shrinking the `GKOutput.__init__` argument list.

pyrokinetics/dataset_wrapper.py
```diff
@@ -155,12 +155,18 @@ def __repr__(self) -> str:
...
+        """Redirect ``x in y`` calls to the inner dataset"""
         return name in self.data
 
     def to_netcdf(self, *args, **kwargs) -> None:
-        """Writes self.data to disk. Forwards all args to xarray.Dataset.to_netcdf."""
+        """
+        Writes self.data to disk. Forwards all args to xarray.Dataset.to_netcdf.
+        Complex data is expanded out into float arrays of shape ``[dims..., 2]``.
+        """
+        # TODO Expand complex numbers selectively
```
bpatel2107: We could probably check if any of the data is complex and only convert to `ReIm` if its there. Would stop things like `Equilibrium` and `Kinetics` having this issue

pyrokinetics/gk_code/gk_output.py
```diff
@@ -14,144 +26,338 @@
...
+    #: TODO document
+    mode: Optional[ArrayLike] = None
+
+    #: 1D grid of the energy in the simulation
+    #: Units of [dimensionless]
+    energy: Optional[ArrayLike] = None
+
+    #: 1D grid of the pitch angle in the simulation
+    #: Units of [dimensionless]
+    pitch: Optional[ArrayLike] = None
```
bpatel2107, line 180: Nothing actually uses energy/pitch angle coords and likely won't unless we add the ability to read distributions functions in. @d7919 Do you think this is likely to happen so worth including?

bpatel2107:
I think this looks great! It makes it very clear what units everything has and simplifies the process for when other code outputs will be added. Less repetition of code is always good and I was a little concerned about it in the original PR.

bpatel2107 (approved):
Overall this looks good. Optionally can check if we need to add a dimension for complex data but happy to merge

pyrokinetics/dataset_wrapper.py
LiamPattinson: Sorted. I thought this would be a really annoying problem to solve, so I thought it best to put it off until a future PR, but it turned out to be much easier than expected.

PR #214, opened 2023-07-12
Enhancements/gamma exb

bpatel2107:
Add in calculation of $\gamma_\mathrm{E\times B}$ from rotation data.

We use the following definition

$\gamma_\mathrm{E\times B} = \frac{\rho}{q}\frac{\partial \omega}{\partial \rho}$ where $\omega$ is the toroidal rotation in radians/second.

Currently we store it in Numerics under `numerics.gamma_exb` as it requires both `local_species` data (`local_species.deuterium.domega_drho`) and `local_geometry` (`local_geometry.rho` and `local_geometry.q`). When loading full equilibrium/kinetics, it is loaded after `load_local` is called. It is also called when the `gk_code` is set in case you load in the species/geometry seperately.

When loading in from GK inputs, it reads in the input files values for $\gamma_\mathrm{E\times B}$ and sets `domega_drho` for all the species, as well as setting it in numerics (from which it is set). 

TODO
- [x] Write some tests


bpatel2107:
Added in test plus ExB shear. Ideally want this merged before the main docs re-structure

bpatel2107:
Merging so docs can be merged

PR #216, opened 2023-07-13
Switch to pyloidal from omas

ZedThree:
We only use the COCOS functions from OMAS, and it's a fairly heavy-weight package. It also includes on-the-fly Cythonisation which slows down imports, and the current latest package breaks on readthedocs.


bpatel2107:
Did you make a library just for the COCOS transform 😆?

bpatel2107 (approved):
Looks good to me

ZedThree:
I asked Liam to do so, yep! It will be useful for some other packages too, like freeqdsk. We might add more things there too that are more generally useful.

PR #217, opened 2023-07-14
Great rename of modules

ZedThree:
This renames modules so that they don't conflict with any class names. Module names are all now lower case.

This also moves the `GKInputReader*` and `GKOutputReader*` classes into the same file -- one per code, e.g. `gs2.py`.

Some classes have also moved into the sub-package `__init__.py` files as well, as there wasn't a need to keep them separate


LiamPattinson:
A lot of this is actually undoing a lot of renaming that I did a little over a year ago, but since I later came to regret most of it, I'm happy to see the changes reverted overall. The only thing I'm unsure about is defining things within `__init__.py`, mostly from a stylistic perspective rather than anything concrete. At some point it's likely that `Kinetics` will have to move back to its own file, as eventually it might end up looking similar to `Equilibrium` and `GKOutput`.

bpatel2107:
I think I agree with @LiamPattinson here. Is there a reason to not have a bespoke kinetics/diagnostics file? Especially for diagnostics which is likely to expand significantly

bpatel2107:
Is this ready to merge btw?

ZedThree:
Yep ready to merge, but note it's into my `docs-structure` PR (#203) not `unstable`.

I suggest we tackle any issues with `kinetics/diagnostics` when/if we run into them.

PR #222, opened 2023-07-26
Restrict IDSPy features to Python version >= 3.9

LiamPattinson:
Disables IMAS data dictionary interoperability for older Python versions. Also restricts the import of IMAS features in `pyrokinetics.databases`.

When this is integrated with the new docs system, I think docs will fail to build for Python 3.8. I've bumped the readthedocs Python version to 3.9 to account for this, but as it stands I don't think users will be able to build their own docs with Python 3.8.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -9,7 +9,7 @@ version: 2
 build:
   os: ubuntu-20.04
   tools:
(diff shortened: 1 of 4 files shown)


bpatel2107:
Looks good to me

PR #223, opened 2023-07-27
Nl documentation

dake0795:
Added documentation explaining how to read in the output from a nonlinear simulations and make simple plots.


bpatel2107:
Looks good so far. I think the following would be good to include

- What the major outputs are and their names/locations in a table
- Link to the `GKOutput` object
- How to convert to different normalisations (see here https://github.com/pyro-kinetics/pyrokinetics/blob/fcf716b19edbc7b83e2932607a2062fd004e5e9f/docs/howtos/normalisations.rst?plain=1#L175-L176)
- Where you have the print statement, can you include what was actually printed out?
-  @ZedThree how do we embed figures as it would be good to see what the plots look like
- I don't think we need to `dequantify()` as the plot should work and have units in the axes

docs/howtos/read_nonlinear_data.rst
```diff
@@ -0,0 +1,46 @@
...
+The gyrokinetic file ``input.cgyro`` is our input file template where we set all the extra flags that have been used in the simulation.  
+
+We can then read the nonlinear simulation into a dataset using ``pyrokinetics``.
+
+.. code-block:: python 
+
+   >>> pyro.load_gk_output(load_moments=True, load_fluxes=True, load_fields=True)
+   >>> data = pyro.gk_output.data
+
+Here, ``pyro`` initialises the utility base dataclass ``GKOutputArgs`` which is used to pass quantities to ``GKOutput``. Derived classes include ``Coords``, ``Fields``, ``Fluxes``, etc. This clas contains features such as automatic unit conversion and a dict-like interface to quantities. Derived classes should define an ``InitVar[Tuple[str, ...]]`` called ``dims``, which sets the dimensionality of each quantity, e.g. ``("kx", "ky", "time")``.
```
bpatel2107: typo `clas` to `class`

docs/howtos/read_nonlinear_data.rst
```diff
@@ -0,0 +1,46 @@
...
+   >>> from pyrokinetics import Pyro, template_dir
+   >>> gk_file = template_dir / "outputs/CGYRO_nonlinear/input.cgyro"
+
+The gyrokinetic file ``input.cgyro`` is our input file template where we set all the extra flags that have been used in the simulation.  
+
+We can then read the nonlinear simulation into a dataset using ``pyrokinetics``.
+
+.. code-block:: python 
+
+   >>> pyro.load_gk_output(load_moments=True, load_fluxes=True, load_fields=True)
```
bpatel2107: This is probably obvious but should we state what these kwargs do? Also which ones are on by default

docs/howtos/read_nonlinear_data.rst
```diff
@@ -0,0 +1,46 @@
...
+
+.. code-block:: python 
+
+   >>> print(pyro.gk_output)
+
+Let's suppose we want to plot the time evolution of the total heat and particle fluxes from our nonlinear simulation. First, we need to extract the relevant quantities from the dataset. After examining the data set, we see that the heat and particle fluxes are given as functions of ``("field", "species", "ky", "time")``. In order to plot the total fluxes as a function of time we sum over the other dimensions. 
+
+.. code-block:: python
+
+   >>> hflux = data['heat'].pint.dequantify().sum(dim='ky').sum(dim='field').sum(dim='species')
```
bpatel2107: You can do
```python
   >>> hflux = data['heat'].sum(dim=('ky', 'field', 'species')) 
```

bpatel2107: Its worth mentioning that we using an xarray `Dataset` as the format so people can use the standard methods here

ZedThree:
For a basic image, use [`.. image:: <filename>`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#images), or to include a caption etc use [`.. figure::`](https://docutils.sourceforge.io/docs/ref/rst/directives.html#figure)

PR #225, opened 2023-07-31
Factories and Readers refactor + documentation

LiamPattinson:
For a while now, I've wanted to revisit how the `Factory` and `Reader` components of Pyrokinetics work under the hood. There are a few issues with them:

- Too much boilerplate: having to manually create and keep track of named `ReaderFactory` objects using `create_reader_factory` and registering readers with them in `__init__.py` files is messy.
- `Kinetics`, `Equilibrium`, `GKInput` and `GKOutput` all work with these tools differently. Need to unify the interface.
- `Equilibrium` and `GKOutput` each hide their reader factories as class attributes and implement the classmethods `from_file`, `supported_types` and `reader`. These functions are almost identical between the two classes, leading to a lot of repeat code.
- `x = factory[key]` creates a new instance of an object rather than referencing something static like in a regular dict, while `factory[key] = X` registers the type `X` with the factory, as you'd expect in a regular dict. The lack of symmetry here is confusing. 
- Naming issues: it's not clear what `verify()` is supposed to do, `read()` could read from anything (file, string, dict, etc), and it isn't obvious that `Reader` is an abstract base class that forces the implementation of these functions.

Since we're trying to expand the docs, now seems like a good time to fix these problems, as it'll be much easier to document a unified and simplified system. This PR fixes these issues and cleans up some related issues with the code. The new interface is as follows:

```python
@readable_from_file
class MyData:
    """Class containing data that can be read from files (GKInput, Equilibrium, etc)"""
    ...

@MyData.reader("csv"):
class MyDataReaderCSV(AbstractFileReader):
    """Reader class that creates MyData from a csv file"""
    def read_from_file(self, path) -> MyData:
        ...

    def verify_file_type(self, path):
        ...

# Factory creation and registration happens behind the scenes
# Automatic file type inference etc still works as usual
my_data = MyData.from_file("foo.csv")
```

## Features 

- Renaming:
    - `readers.py` -> `file_utils.py` 
    - `Reader` -> `AbstractFileReader`
    - `read()` -> `read_from_file()`
    - `verify()` -> `verify_file_type()`
- Updated `Factory` interface:
    - `factory[key] = X` registers `key` with the type `X`
    - `X = factory[key]` returns the type registered with `key` -- no longer instantiates a new object.
    - `x = factory(key)` instantiates a new object
    - Functions `factory.register(key, cls)`, `factory.type(key)`, and `factory.create(key)` are available for those who would rather be explicit
    - Doesn't subclass `UserDict`, but does now explicitly implement a subset of the dict interface. 
- No longer need to create factories explicitly. Instead, the decorator `@readable_from_file` can be used to mark classes as being generated from files, and the decorator `@Readable.reader(key)` then registers.
- Updated `Kinetics` to work similarly to `GKOutput`/`Equilibrium`. Constructor no longer takes a file and uses reader classes to generate its own `kwargs` -- instead the `Kinetics` constructor takes plain data and the reader classes return a fully instantiated `Kinetics`.
- `GKInput`, `GKOutput`, `Equilibrium`, and `Kinetics` now have the same interface for reading from files and registering new readers

## Issues

- Still need to write a tutorial for the docs to explain how this stuff works for other devs
- Docs building raises a bunch of new warnings regarding some decorator-created classmethods. It recommends to use `@functools.wraps`, but I don't think this is actually applicable here, and doing so breaks everything in Python 3.10. @ZedThree Any ideas?

---

I'm keeping this as a draft until I've written up a proper tutorial for this aspect of the code. Let me know if you'd like to see any other changes while I'm at it.


LiamPattinson:
I fixed the sphinx warnings by moving the classmethods introduced by `readable_from_file` into a [mixin](https://en.wikipedia.org/wiki/Mixin) class. Unfortunately, this means you need _both_ a decorator and a new base class:

```python
@readable_from_file
class MyReadable(ReadableFromFileMixin):
    ...
```

This is less clean than I would like, and there are issues with linking to the mixin methods in the docs, but it does resolve the issues.

(I also tried solving this using metaclasses, which allowed inheritance from `ReadableFromFileMixin` to also inject a `_factory` class attr,  but then I had to also make a new metaclass for `GKInput` to resolve a metaclass conflict, and so the solution started looking a little too complex for the maintainability of the project.)

LiamPattinson:
I've added a tutorial for file readers, though I'm not sure if that's the correct place for it or if it should be in 'user guides' or something. Let me know if you'd like me to change anything about the new docs, or about the changes to factory/reader stuff. Otherwise, I think this is ready.

bpatel2107:
That looks great and super detailed! I think tutorial is the best place for that right now so I'm happy to merge.

bpatel2107 (approved):
LGTM

PR #227, opened 2023-08-01
Change theta grid from original geometry to MXH geometry when loading…

bpatel2107:
… in IMAS data dict and vice versa

When writing/reading IMAS data dictionaries we need to convert the geometry to `MXH`. We should have also changed the `theta` in the `gk_output` to the `MXH` equivalent as they don't map 1 to 1. This fixes this


mgiacomin (approved):
It looks good.

PR #231, opened 2023-08-03
Need |q| for balloon transform in CGYRO

bpatel2107:
CGYRO needs to use |q| when doing ballooning transform otherwise phase between connected `2 pi` segments are wrong if `q<0` (depends on field/current directions)

E.g. MTM eigenfunction with negative q before fix:
image

After fix:
image


mgiacomin (approved):
Well spotted!

PR #232, opened 2023-08-03
Jetto files

arkabokshi:
This PR enables reading in the impurity (profiles) directly from the JETTO `.jss` and `.jsp` files.


pyrokinetics/kinetics/jetto.py
```diff
@@ -8,67 +8,59 @@
...
         """
         Reads in JETTO profiles NetCDF file
         """
         # Open data file, get generic data
-        with nc.Dataset(filename) as kinetics_data:
-            time_cdf = kinetics_data["TIME3"][:]
+        try:
+            kinetics_data = read_binary_file(filename)
+            jetto_jss = read_binary_file(str(filename).replace("jsp", "jss"))
+            time_cdf = kinetics_data["TIME"].T[:]
```
bpatel2107: I don't think want to remove the ability to read the `netCDF` file, just add in the ability to read `.jsp` and `.jss`. Ideally we would be able to read both `jsp` and `netCDF` files and then also require there to be the `.jss` file to get the charge information

pyrokinetics/kinetics/jetto.py
```diff
@@ -105,51 +97,42 @@ def read(
...
-                    )
-                    impurity_mass = (
-                        self.impurity_charge_to_mass[int(impurity_charge.m)]
-                        * hydrogen_mass
-                    )
+                # impurity charge can have a profile variation
+                impurity_charge_data = (
+                    kinetics_data[impurity_z][time_index, :] * units.elementary_charge
+                )
+                impurity_charge_func = UnitSpline(psi_n, impurity_charge_data)
```
bpatel2107: The `kinetics` object needs to contain the full profile information, not just one flux surface. What we need to do is to change the `charge` in a `Species` object to be a function that takes in `psi_n` rather than a number. Then we need to change that for all the other readers too so they are all consistent. Finally we'll modify the `local_species.from_kinetics` to be able to read in this new format.

For the lighter elements this will just end up being a constant for all flux surfaces, but for heavier impurities there will be profile variation in the charge.

pyrokinetics/kinetics/jetto.py
arkabokshi: Is there any advantage of keeping the option to read in the `netCDF` file?  Is the latter not derived from the `jetto.{jsp,jss}` files?

LiamPattinson:
If you want to get the automated tests working, you'll need to add `"jetto-tools"` to the dependencies list in `pyproject.toml`. I'd also recommend running `flake8` and fixing any of the issues it reports.

As an example for a way to have NetCDF and `.jss` reading:

```python
def read(self, filename: PathLike, time_index: int = -1, time: float = None) -> Dict[str, Species]:
    if self.verify_jetto_netcdf(filename):
        return self.read_jetto_netcdf(filename, ...other args...)
    elif self.verify_jetto_jss(filename):
        return self.read_jetto_jss(filename, ...other args...)
    else:
        raise RuntimeError("some error msg")

def read_jetto_netcdf(self, filename: PathLike, time_index: int = -1, time: float = None) -> Dict[str, Species]:
    # The original read function
    ...

def read_jetto_jss(self, filename: PathLike, time_index: int = -1, time: float = None) -> Dict[str, Species]:
    # The new read function
    ...

def verify(self, filename: PathLike) -> None:
    try:
        self.verify_jetto_netcdf(filename)
    except Exception:
        try:
            self.verify_jetto_jss(filename)
        except Exception:
            raise RuntimeError("File is neither jss nor netcdf")

def verify_jetto_netcdf(self, filename: PathLike) -> bool:
    # The original verify, but return True at the end
    ...
    return True

def verify_jetto_jss(self, filename: PathLike) -> bool:
    # The new verify, but return True at the end
    ...
    return True
```

You could extend this for `.jsp` pretty easily too.

pyrokinetics/kinetics/jetto.py
bpatel2107: @fcasson can probably comment on this best. I guess my concern is to keep current work flows as is

fcasson: Personally I would prefer it if you just read the jsp files.  It is is closer to the single source of truth.  The NETCDF is not always present and represents an extra post-processing step.

But if you need to preserve existing workflows, sure whatever.  Just check the results are the same...

bpatel2107: Okay happy to make the switch, we'll need to switch out the template files too and find the corresponding `.jsp` and `.jss` files

arkabokshi: @bpatel2107 Could you switch the template file?

docs/examples/example_JETTO.py
```diff
@@ -9,12 +9,13 @@ def main(base_path: Union[os.PathLike, str] = "."):
...
 
     # Kinetics data file
-    kinetics_file = template_dir / "jetto.cdf"
+    kinetics_file = template_dir / "jetto.jsp"
 
     # Load up pyro object
     pyro = Pyro(
         eq_file=eq_file,
         kinetics_file=kinetics_file,
+        kinetics_type="JETTO",
```
bpatel2107, line 18: We'll wanna get the `verify` method working before we merge

bpatel2107:
I've updated the template files and tests so we should be okay. I want to make sure that this generates the same inputs as before. @dake0795 @mgiacomin can you check to see if you get the same input files when you switch from the netCDF to the jsp files? The only thing that should be different is the charge on the `impurity` and `zeff`

If that is the same then I'm happy to merge.

mgiacomin:
@bpatel2107 I confirm I get the same input files if I switch from netCDF to jsp files. This can be tested only on old STEP cases. New STEP cases require to read jsp files.

bpatel2107:
Sounds good to me! @arkabokshi anything else you want to add?

arkabokshi:
Thanks, this looks good to me! Need to fix `TALF`, but can be done in a separate PR.

bpatel2107 (approved):

Issue #233, opened 2023-08-05
coll_FLR in GENE


arkabokshi:
@dake0795  Should we add `coll_FLR =.true.` in GENE?  The code processes this as `.false.` if not specified I believe?

arkabokshi:
![omega_gene](https://github.com/pyro-kinetics/pyrokinetics/assets/30498018/59711a13-9e5d-4cc6-863c-4e0ac55a8194)
![omega_gene](https://github.com/pyro-kinetics/pyrokinetics/assets/30498018/398cb46f-c41a-426a-a518-4771656aa31d)

Impact of `coll_FLR` term appears to be substantial.

bpatel2107:
FLR terms are on by default in CGYRO/GS2. This sort of open up a wider issue of which collisional terms/operators are on by default in different codes. We can probably store this info in `numerics` similar to how it is done in the GKDB doc (section 4.6.4)

arkabokshi:
Aside, would you not want the most "advanced" collisional operator options enabled in each GK code?

bpatel2107:
This is beginning to step outside the remit of pyrokinetics (though not quite yet I think). Different codes have different models implemented so the most advanced for one code maybe be equivalent to a simpler model in another code. 

Part of me wants to keep the codes chosen default for collisions and let the user decide what to do (there may be a reason that FLR effects are off in GENE by default). But given that this leans more towards physics rather than numerical methods I would be fine to include a more sophisticated way to handle this.

d7919:
I think I'd agree with leaning towards leaving switches as code defaults unless we have a very good reason not to do this. I think the role of pyrokinetics is to act as something which understands the approach of different codes but not as something which has opinions about how to use those codes?

One might consider documenting differences to consider and/or providing opinionated templates for different codes but I think the defaults should be decided by the codes themselves.

Of course we could change our mind and let the tool become opinionated, but I'm not sure that's something we'd want at this stage.

mgiacomin:
I think a similar discussion was in #200, where we concluded that `pyrokinetics` should not change the default values in templates unless there is a very good reason to do so.

bpatel2107:
Okay so I think we leave defaults as is, but leave open the option to set these terms in the future probably via something like `pyro.numerics.collisions.flr = True`

arkabokshi:
I agree with the points, but in this case the rationale is slightly different:

- If we are choosing `sugma` operator, enabling this is "more physical" 
[...]

PR #234, opened 2023-08-05
switch on coll_flr

arkabokshi:
See issue #233


bpatel2107:
I think we've settled on leaving code defaults where possible and leaving it to the user to check these sort of things. I think collision physics can blur the lines here but if/when we handle it I think we can use the GKDB method of selecting flags as that covered most of what would be needed. It is also a thing in IMAS so we will likely want to change that too.

PR #236, opened 2023-08-06
Add TGLF SAT123 and saturation parameter calculations

tomneiser:
This adds the ability to calculate saturated fluxes from quasilinear weight, growth rate and ky spectra, and relevant plasma parameters in `input.tglf`. The saturation rules were developed by Gary Staebler _et al_ for SAT12, and @hgd511 and @fcasson _et al_ for SAT3. This code is very fortran-y, which is deliberate since it is easier to update in response to changes to the saturation rules. I've added a test `test_tglf.py` that can also serve as a minimal working example. Over time the test can be modified to use pyro objects (I wasn't able to load the TGLF template output using pyro at the time of this writing).


bpatel2107:
This looks like a good start! I think main issue is that this only works using TGLF inputs, whilst we would want this to work with all inputs. Ideally we'd like to use existing `pyro` object to define the inputs to the saturation rule as you say so I think it makes sense to merge into unstable when that is achieved.

For example we already calculate things like local geometry parameter like `Q_LOC` and `Q_PRIME_LOC` and even derivative of flux surface values like   `$\partial Z / \partial r$` etc. in `local_geometry`. `metric_terms` also calculates things like `grad_r` too. Essentially everything in `get_sat_params` should already exist or be trivial to obtain.

I think we would want something like `intensity_sat` to read in a `PyroScan` object which has been done over `ky` or something equivalent. Then we could take a GS2/CGYRO/GENE scan and calculate the sat rules from it.  May be worth arranging a meeting to discuss how to get everything out of pyro from this

PR #237, opened 2023-08-07
Bugfix/tglf transport eigval

### LINKED ISSUE #235: GKOutputReaderTGLF expecting eigenvalues
When trying to use `GKOutputReaderTGLF()` to read the template outputs in the `templates/outputs/TGLF_transport` directory with the code snippet below
```path = template_dir / "outputs" / "TGLF_transport"
pyro = Pyro(gk_file=path / "input.tglf", name="test_gk_output_tglf")
norm = pyro.norms
tglf = GKOutputReaderTGLF().read(template_path, norm=norm)
```
I get the bug:
```
File pyrokinetics/pyrokinetics/gk_code/tglf.py, line 570, in read
eigenvalues=Eigenvalues(**eigenvalues, dims=eigenvalues_dims).with_units(
TypeError: __init__() got an unexpected keyword argument 'eigenvalues'
```
The code snippet works when trying to access the template data in `templates/outputs/TGLF_linear`, which is also covered by a regression test, so we should also cover `templates/outputs/TGLF_transport` once this bug is fixed. I could try and fix it but someone more experienced with pyro will probably find a better fix faster.

bpatel2107:
Remove `eigenvalues` from TGLF transport `Eigenvalues` object as we only need `growth_rate` and `mode_frequency`. Add a test to try and read output data.

Fixes #235

### DIFF
--- a/pyrokinetics/gk_code/tglf.py
+++ b/pyrokinetics/gk_code/tglf.py
@@ -861,7 +861,6 @@ class GKOutputReaderTGLF(Reader):
             eigenvalues = np.reshape(full_data, (nky, nmode, 2))
             eigenvalues = -eigenvalues[:, :, 1] + 1j * eigenvalues[:, :, 0]
 
-            results["eigenvalues"] = eigenvalues
             results["growth_rate"] = np.imag(eigenvalues)
             results["mode_frequency"] = np.real(eigenvalues)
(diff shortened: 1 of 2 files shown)


tomneiser (approved):
Looks good, thanks!

PR #238, opened 2023-08-08
Add check for newer gs2 netcdf plus fix eigenfunction attr error

bpatel2107:
Read in GS2 netCDF from `8.0.6-1277-g8fb9bcae` 
Also fix attr of `Eigenfunction` in `GKOutput`


dake0795 (approved):
Looks good to me!

PR #244, opened 2023-08-17
Fix alpha temp

arkabokshi:


arkabokshi:
@bpatel2107  @fcasson  Is this broadly what you would expect the fast temperature to be?  Worth pointing out that at the edge  `NALF` rapidly goes towards zero, so the fast temperature is unphysically large.

pyrokinetics/kinetics/jetto.py
```diff
@@ -167,6 +167,37 @@ def read_from_file(
...
+        fast_temp_data = (
+            (ion_total_pressure - slow_pressure) / kinetics_data["NALF"][time_index, :]
+        ) * units.eV
+
+        fast_temp_func = UnitSpline(psi_n, fast_temp_data)
+
+        np.savetxt("fast_pressure", fast_pressure)
+        np.savetxt("nalf", kinetics_data["NALF"][time_index, :])
+
+        result["alphas"] = Species(
```
arkabokshi: @bpatel2107  @LiamPattinson  Is there a way to update just the temperature? (Please ignore the error in populating the fields.)

bpatel2107: The alpha pressure can be taken from `2/3 * WALD` from the `jsp` file. `WALD` is the energy density of the alphas and the `2/3` converts it from energy density to a pressure

arkabokshi: Oh very helpful!  Hopefully the two are consistent.   :)

bpatel2107: I would define a `fast_ion_temp_func` from `WALD` and then when creating the fast ion `Species` object have a check for whether or not to use the thermal temp or fast ion temp 

i.e. something like
```python
thermal_ion_temp_func = kinetics_data["TI" ]
fast_ion_temp_func = 2/3 * kinetics_data["WALD" ] / kinetics_data["NALF"]
if species_name == "alpha":
    temp_func = fast_ion_temp_func
else:   
    temp_func = thermal_ion_temp_func
   
result[species["species_name"]] = Species(
    species_type=species["species_name"],
    charge=species["charge"],
    mass=species["mass"],
    dens=density_func,
    temp=temp_func,
    rot=rotation_func,
    ang=omega_func,
    rho=rho_func,
)
```

arkabokshi: Thanks for this.  What is the rationale behind checking for thermal ion temp for fast species?  EDIT:  I think you didn't mean fast_ion_temp can also be at thermal temp, instead to check species to append correct fast_temp for alphas.

bpatel2107: Right now we iterate through `possible_species` which is a dictionary of all the ion species with all the data needed to create a `Species` object. This just sets it properly when looping through `possible_species`

pyrokinetics/kinetics/jetto.py
```diff
@@ -4,7 +4,7 @@
 from ..constants import electron_mass, hydrogen_mass, deuterium_mass
 from ..units import ureg as units, UnitSpline
 from ..file_utils import AbstractFileReader
-
+from scipy.constants import elementary_charge
```
bpatel2107: We can get this from `..constants`, though you'll wanna take the magnitude

https://github.com/pyro-kinetics/pyrokinetics/blob/b1970ccbd4af754ec9f1a7c6971a40fcede7e73d/pyrokinetics/constants.py#L10

pyrokinetics/kinetics/jetto.py
```diff
@@ -156,6 +167,11 @@ def read_from_file(
 
             density_func = UnitSpline(psi_n, density_data)
 
+            if species["species_name"] == "alphas":
```
bpatel2107: This may be nit picky but do we want to call this `alpha` or `alphas`. We don't use the plural for electrons for example. Its a bit confusing as I guess the plural for helium is helium but I think we should be consistent

bpatel2107 (approved):
Overall this is good. A few minor things to be addressed

bpatel2107 (approved):
LGTM

PR #252, opened 2023-09-11
Document collisions

### LINKED ISSUE #251: TGLF collision bug?
`TGLF.XNUE` is the electron-ion collision frequency.  Therefore to map to the electron-electron collision frequency, such as in https://github.com/pyro-kinetics/pyrokinetics/blob/038eb7aee01ce05e5b2fa74334048dc74f2ab57e/pyrokinetics/gk_code/tglf.py#L319 we should have `nu_ee = ne/ni * xnue`?

arkabokshi:
- minor fix to jetto example doc
- adding a user guide that outlines the mapping between collision frequencies in various codes and Pyrokinetics convention


arkabokshi:
Please could @dake0795 and @ycamenen check for GENE and GKW?  Seems the mapping between the two codes is `GKW.coll_freq = GENE.coll * Rref/Lref * 2^1.5`?

bpatel2107:
You'll need to merge unstable back into this to get the readthedocs working again (they recent enforced the requirement for a specific yaml file in unstable). We've had to pin a dependency to allow for the 3.10 tests to pass

arkabokshi:
@bpatel2107 @dake0795 @ycamenen   I am happy with the state of the documentation.  Any thoughts?

bpatel2107:
Documentation looks good for CGYRO but for TGLF I would say that `XNUE` is the same as `NU_EE` and in the attention box say that although it is said to be the electron-ion collision frequency, the source code looks to be the electron-electron one. For GS2 it may be worth pointing out the `sqrt(2)` factor is already handled by pyro.

Maybe you want to use your institution email instead of personal?

dake0795:
Looks good on the GENE side! Sorry for taking so long with this. It would be nice to have some explanation in here of where the multiplicative constants come from in GKW and GENE (the numbers in the equation) but this is just personal preference!

dake0795 (approved):

PR #254, opened 2023-09-20
Enhancements/gene moments

bpatel2107:
Added the ability to read in moments from GENE. 

Need to be consisten with our definitions of the different moments and this is what GENE spits out.
image

image

We currently just have the density, full temperature and velocity moments. Adding the $\delta T = T_{1,||} + 2*T_{1,\perp} + 2 * n_1$ I think will get us to the right temperature moment, but I don't think we can technically get the full velocity from what we have here...

@dake0795 @arkabokshi @mgiacomin  FYI


mgiacomin:
@bpatel2107 I am not sure I agree with your definition of $\delta T$. Indeed, $\delta T$ should be defined as
$$\delta T = \frac{2}{3}\frac{m}{2n_0}\int v^2 \delta f\ d^3v - \frac{T_0n_1}{n_0}$$
from which one can define $\delta T_\perp$ and $\delta T_\parallel$ 
$$\delta T_\perp = \frac{m}{2n_0}\int v_\perp^2 \delta f\ d^3v - \frac{T_0n_1}{n_0}$$
$$\delta T_\parallel = \frac{m}{n_0}\int v_\parallel^2 \delta f\ d^3v - \frac{T_0n_1}{n_0}$$ 
This leads to $\delta T = \frac{2}{3}\delta T_\perp +\frac{1}{3}\delta T_\parallel$.

bpatel2107:
I guess the factor $1/3$ comes from a thermodynamic point of view. I was working off of this from @ajaycj1's recent paper where there is no factor

image

mgiacomin:
You can see the need for the factor $2/3$ if you take a Maxwellian distribution function
$$T = \frac{1}{n}\boldsymbol{\frac{2}{3}}\int \frac{m v^2}{2} f_M$$
I am not sure why that should be different for $\delta T$... 
@ajaycj1 ?

bpatel2107:
This is good to go. Just waiting to see if we need the $2/3$ or not. That would be a small change here 

https://github.com/pyro-kinetics/pyrokinetics/blob/e781ac4c04bb858789a47a198d519dcf2714bb7b/pyrokinetics/gk_code/gene.py#L1242-L1246

bpatel2107:
I think this is working and worth merging sooner rather than later. The temperature fluctuations look reasonable from a GA-STD style case


62ccaadc-aed2-4680-99e2-2731e8d191cf

ezgif com-gif-maker

ajaycj1:
I agree with @mgiacomin. Typo on my paper - sorry.
I stopped using github. I am available on trax.42@hotmail.com.

Issue #256, opened 2023-09-21
Proposal for plugin system


LiamPattinson:
I've been playing around with some of the more obscure Python packaging features, and I think I've found a way to improve the way we add new codes. Using _entry points_, we could design an interface for other libraries to hook into our code and register their own classes from within their own projects.

## Concept

Let's say we want to accept plugins for equilibrium reader classes. We could achieve this by adding something like the following to `equilibrium/__init__.py`:

```python
from importlib.metadata import entry_points

# the entry_points() function returns all entry point 'groups'
# Here, we only want the group called 'pyrokinetics.equilibrium'
# Note that this isn't a module path!
eq_entry_points = entry_points()["pyrokinetics.equilibrium"]

for entry_point in eq_entry_points:
    Equilibrium._factory[entry_point.name] = entry_point.load()
```

Meanwhile, if somebody were to write their own Pyrokinetics plugin that reads a new equilibrium type, they could register it with the `"pyrokinetics.equilibrium"` group by adding the following to their `pyproject.toml`:

```toml
[project.entry-points."pyrokinetics.equilibrium"]
vmec = "MyVMECLibrary.MyPyrokineticsPlugin:VMECReader"
```

This would add the class `VMECReader` to Pyrokinetics with the key `"vmec"`. If both packages were installed in the same Python environment, this would add VMEC functionality to Pyrokinetics without needing to explicitly import the VMEC library.

I made a little toy library called [`pyplugins`](https://github.com/LiamPattinson/pyplugins) which demonstrates this. It includes both an 'internal plugin', that ships with the core package, and an 'external plugin', which is a separate project. None of the libraries included directly `import` each other, but due to the way it's packaged it's possible to use all three via a common interface.

If we were to go all-in on a plugin architecture, it would be possible to reorganise Pyrokinetics as follows:

```bash
|__pyrokinetics
   |_ pyproject.toml
   |__ src
       |_ pyrokinetics # Define the core library, with no explicit reference to any codes
       |_ pyrokinetics-gs2 # GS2 plugin, adds input/output reader and normalisation scheme
       |_ pyrokinetics-cgyro # CGYRO plugin, see above
       |_ ... other GK codes ...
       |_ pyrokinetics-geqdsk # Plugin for G-EQDSK reading
       |_ pyrokinetics-transp # Define both equilibrium and kinetics plugins
```
[...]

PR #258, opened 2023-10-05
Add entry-points plugin system + update factories (again)

LiamPattinson:
I've added a plugin system as described in Issue https://github.com/pyro-kinetics/pyrokinetics/issues/256 that allows users to register their own readers for `Equilibrium`/`Kinetics`/`GKInput`/`GKOutput`. These are automatically added to Pyrokinetics if the user's library is installed in the same Python environment.

Testing a plugin system requires the existence of plugins, so I also created a repository of [example plugins](https://github.com/LiamPattinson/pyrokinetics-plugin-examples). This is currently on my personal account, but I can transfer it to the pyro-kinetics organisation if you'd like (thought this might break the PyPI integration -- they've moved away from using API tokens, and instead may require the creation of a pyro-kinetics PyPI organisation). The plugin package is installed if we use `pip install .[tests]` but otherwise isn't included.

Currently I have a test for `Equilibrium` plugins, and I still need to write tests for `Kinetics`, `GKInput`, and `GKOutput`.

---

I also happened across a better way to work with the factory/file reader bits of the code. Currently, you need to use a combination of inheritance and decorators to make a class a 'readable' or a 'reader':

- 'Readables' inherit `ReadableFromFileMixin` to get access to classmethods `from_file`, `supported_file_types`, and `reader`. The `readable_from_file` decorator adds a factory class variable.
- 'Readers' inherit `AbstractFileReader` to get the functions `read_from_file` and `verify_file_type`, and the `Readable.reader` decorator registers them with the readable's factory.

By adding the special method `__init_subclass__` to the two superclasses (renamed `ReadableFromFile` and `FileReader`), we can do away with the decorators and instead use keyword arguments in the class definition:

```python
class GEQDSKReader(FileReader, file_type="GEQDSK", reads=Equilibrium):
    ...
```

Along with simplifying the interface, this also means the code won't run unless you supply the keyword arguments, whereas currently it's possible for users to forget the decorators and end up with subtly broken code.

---

Aside: If you check the `publish.yml` CI job in the plugins package I made, I'm using the new and updated way of publishing to PyPI, plus using TestPyPI to check everything works as expected. I'd like to update Pyrokinetics to do the same, but that can wait for a future PR.


LiamPattinson:
Unrelated, but I think code cov still isn't working. It seems to only be testing example files in the docs for some reason. I think it might be fixed by adding an empty `__init__.py` to the top-level tests folder and changing the pytest command to `pytest --cov=pyrokinetics --cov-report=xml tests`, but I can play around in a new branch to figure it out. I might be able to merge the tests and coverage workflows while I'm at it too.

LiamPattinson:
I think I have this ready to go now.

bpatel2107 (approved):
Overall this looks really nice and is a cleaner way to set up `Readers` and `Readables`.

I think keeping the testing plugins separate is better and actually may be worth linking to in the docs to show a `real world` example of how to implement a plugin.

Happy for the publishing to be changed in a future PR

LiamPattinson:
Cheers @bpatel2107. I was planning on making some minor tweaks today, so could we hold off on merging for a day or so? I had a chat with @ZedThree about the separate plugin library, and he had some recommendations to make things a little cleaner.

LiamPattinson:
I think I'm good now. Tweaked the plugin library to escape a potential circular dependency problem and improve its CI, and updated this package so that the plugin tests will simply be skipped if there's a problem importing the plugins. Added a paragraph to the docs like you recommended.

In future, it may be preferable to pin the version of the plugin library to a specific git commit rather than a published version. For example, if future experimental pyrokinetics branches modify something that breaks compatibility with the plugin library, it might not be appropriate to release a new version of the plugin library, but it would instead be possible to pin it to a corresponding experimental plugin branch. It might be a bit fiddly to keep both packages synced up, but hopefully it won't be something we have to touch very often.

bpatel2107:
I don't see the changes to the docs when I load the readthedocs build

https://pyrokinetics--258.org.readthedocs.build/en/258/tutorials/file_readers.html

Something wrong awry? Maybe the build doesn't trigger on changes to `rst` files? Likely fine when merged into unstable...

LiamPattinson:
I'm not sure what's going on there as the changes are there when I open the page. Maybe it just took readthedocs a while to redirect links?

bpatel2107:
Its probably fine

PR #259, opened 2023-10-05
Fix to code coverage, minor test updates

LiamPattinson:
I think I might have gotten code coverage working in the same workflow as the regular tests. I also fixed some issues in the tests themselves:

- Hadn't run black on them in a while.
- In some cases test modules were importing functions from other test modules. This usually works because pytest adds the local path to `sys.path` when it runs, but it isn't intended behaviour, and it can cause weird import issues from time to time. ~~One such case is when you add an `__init__.py` to the top of the `tests` folder, which is sometimes needed to get code coverage to work (for reasons I can't understand)~~ [Edit: Never mind, I had this issue confused with another]. Pytest are moving to a new system that uses importlib instead of messing with path, so I've enabled that in `pyproject.toml`, and I've moved the shared functions in the tests folder into session-scoped fixtures in `conftest.py`.


LiamPattinson:
It only took all evening, but I finally got code coverage working. Turns out that when I split the workflow into separate test and coverage jobs, I'd forgotten to checkout the project in the second job, so Codecov didn't have a git project against which to compare the `coverage.xml` file. Whoops.

- Removed `.coverage.rc` and moved coverage setup into `pyproject.toml`.
- Controversial: moved the project to a 'src' layout instead of the 'flat' layout we'd used previously, so code is now in `src/pyrokinetics` instead of `pyrokinetics`. This is because `pytest --cov=pyrokinetics` gets confused with a flat layout, as it tries to determine coverage of the code in the local directory `./pyrokinetics` instead of comparing against the installed library, which is what we really should be testing. `import pyrokinetics` doing different things depending on whether the developer is sitting in the root directory or not can cause a bunch of other problems too. [This page in the pytest docs](https://docs.pytest.org/en/latest/explanation/goodpractices.html) explains it all pretty well.

bpatel2107:
This looks much better on the coverage!

I think I've come across this problem with `import pyrokinetics` not quite importing the right thing depending on the directory I was in so this seems like a clean fix. Just to check will `pip install -e .`  in the top level directory still work as normal? As long as  the overall functionality doesn't change this seems like a good move.

.github/workflows/tests.yml
```diff
@@ -17,16 +18,36 @@ jobs:
...
+    needs:
+    - pytest
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Download coverage report
+      uses: actions/download-artifact@v3
+      with:
+        name: "coverage-python-3.9"
```
bpatel2107: Is it worth bumping this up to 3.11 to ensure the IMAS test get covered too?

bpatel2107: Actually it is available for 3.9 so ignore this. We don't even test for 3.11 as is!

I believe that there will be updates to IMAS that will likely only be available for 3.10 but maybe we can cross that bridge we when get there. We could be proactive and switch it now

bpatel2107 (approved):
LGTM overall

.github/workflows/tests.yml
LiamPattinson: I went with 3.9 as it's the minimum supported version that still uses all the code. If you test with 3.8, code coverage gives you a 0% on the IMAS bits, so it's not a fair representation (there are workarounds to this but they're messy). I know that IMAS doesn't play well with 3.9 though, so I'd be happy to bump it to 3.10. 

I think it'd be a good idea to start testing 3.11 (and maybe even 3.12 if the GitHub runners support it). I was planning to add it in an upcoming rework of some of the packaging/testing/linting/CI tooling, but I suppose it might make sense to do it now while I'm already messing with the testing CI.

bpatel2107: Happy to include 3.11 and and consider dropping 3.8

LiamPattinson: I'd prefer to hold on to 3.8 for now since it's the built-in Python version for Ubuntu 20.04 and won't reach end-of-life until October next year, but since we're already having to provide a reduced feature set it might make sense to drop 3.8 earlier than that.

LiamPattinson:
> This looks much better on the coverage!
> 
> I think I've come across this problem with `import pyrokinetics` not quite importing the right thing depending on the directory I was in so this seems like a clean fix. Just to check will `pip install -e .` in the top level directory still work as normal? As long as the overall functionality doesn't change this seems like a good move.

As far as I can tell, both regular and editable installs should work fine, and both pytest and coverage work either way. It won't be possible to run tests without installing first, but I'd argue that's a feature.

PR #260, opened 2023-10-12
Fix downsizing in CGYRO

bpatel2107:
Downsizing `GKOutput` in `time` wasn't correctly working in CGYRO so now that is fixed for `fluxes`, `moments` and `fields`

Added some defaults in that were probably needed.

Probably worth adding some tests for this, but only really makes sense for nonlinear runs.

A new release after this would be a good idea, there have been a few notable changes.


bpatel2107:
This is good to go but i dont know why it is failing. On my local version (on 2 different systems) this passes without issue

LiamPattinson:
I can't tell what's going wrong here, as I've tried it using fresh local installs using all the same versions as the CI jobs, and all of them pass. Please could you try changing out the calls to `assert np.allclose(...)` to `np.testing.assert_allclose(...)`? The `numpy.testing.assert_*`  functions give better information as to what's causing the tests to fail, but you may need to set `atol` and `rtol` explicitly, as the defaults are much stricter than `np.allclose`.

It would also be helpful if the different codes, downsize numbers, and linear/nonlinear cases could be split into separate tests, as the failure of GENE when `downsize==2` here could be masking additional failures for the other codes when `downsize>=2`. We could minimise repeat code using `pytest.mark.parametrize`:

```python
@pytest.mark.parameterize("downsize", (2, 3, 4))
def test_cgyro_linear_output_downsize(downsize):
    # Test time values from linear CGYRO (can't do fields due to normalisation)
    cgyro = Pyro(
        gk_file=template_dir / "outputs/CGYRO_linear/input.cgyro", gk_code="CGYRO"
    )
    cgyro.load_gk_output()
    full_data = cgyro.gk_output
    
    cgyro.load_gk_output(downsize=downsize)
    downsize_data = cgyro.gk_output
    
    np.testing.assert_allclose(
        full_data["time"][::downsize],
        downsize_data["time"],
        atol=1e-8,
        rtol=1e-5,
    )
```

bpatel2107:
Good idea about the parameterising. The GENE fails mostly due to there not being enough output data, but I could probably just add some more so we can parameterise over code and downsize

bpatel2107:
I've parametrized this test for linear/nonlinear CGYRO and linear GENE. I had to modify the template so there was enough data to actually downsize.

This should be good to go now

bpatel2107:
Will merge now as the project coverage passes

PR #261, opened 2023-10-16
Handle rotation/rotation gradients better

### LINKED ISSUE #257: Rotation value and units in the GENE input file generation
The pyrokinetics code can generate GENE input file from the TRANSP output, which is fantastic. However, while using the code I noticed that in the GENE input file generated by pyrokinetics: 
  1. There is no toroidal rotation value (not the flow shear value, but the rotation, corresponding to Omega0_tor in GENE manual). 
  2. There are no explicit values for the normalization units, for instance, the values of Bref, Tref, nref in SI unit in the GENE input file.
I'm just wondering whether we can add these two function to the pyrokinetics code?
Thank you very much in advance

bpatel2107:
Initially we stored the toroidal rotation velocity and the angular velocity in `Kinetics`/`Species`/`LocalSpecies` but the angular velocity is more relevant so I've removed the double definition for clarity.

I've also include included the values for parallel velocity gradient and mach number when writing codes for inputs. So for each code we store

- $\omega_0$ which relates to the Mach number, stored in `LocalSpecies` 
- $\frac{\partial \omega_0}{\partial \rho}$ which relates to the $E\times B$ shear and parallel velocity gradient (PVG).

Given these two different uses for the derivative, I treat them slightly differently. The PVG is taken from `LocalSpecies` when writing files. The $E\times B$ shear is to be stored in `Numerics` and then when we write files it is taken from there. This will allow for them to be set independently of each other. So when loading files from a `Kinetics`/`Equilibrium` they are set consistently, but when reading from a GK file they can be different.

I've added them into each code as needed.

@daviddickinson it would be useful to decide what to do about `uprim` in GS2.

Partially fixes #257

### DIFF
--- a/src/pyrokinetics/gk_code/gs2.py
+++ b/src/pyrokinetics/gk_code/gs2.py
@@ -509,7 +509,7 @@ class GKInputGS2(GKInput, FileReader, file_type="GS2", reads=GKInput):
 
             self.data[species_key]["uprim"] = (
                 local_species[name]["domega_drho"]
-                / self.data["theta_grid_parameters"]["rmaj"]
+                * self.data["theta_grid_parameters"]["rmaj"]
             )
 
         self.data["dist_fn_knobs"]["mach"] = local_species.electron.omega0
--- a/src/pyrokinetics/gk_code/gs2.py
+++ b/src/pyrokinetics/gk_code/gs2.py
@@ -59,7 +59,6 @@ class GKInputGS2(GKInput, FileReader, file_type="GS2", reads=GKInput):
         "nu": "vnewk",
         "inverse_lt": "tprim",
         "inverse_ln": "fprim",
-        "inverse_lv": "uprim",
     }
 
     def read_from_file(self, filename: PathLike) -> Dict[str, Any]:
@@ -224,12 +223,6 @@ class GKInputGS2(GKInput, FileReader, file_type="GS2", reads=GKInput):
 
         ne_norm, Te_norm = self.get_ne_te_normalisation()
 
-        domega_drho = (
-            self.data["theta_grid_parameters"]["qinp"]
-            / self.data["theta_grid_parameters"]["rhoc"]
-            * self.data["dist_fn_knobs"].get("g_exb", 0.0)
-        )
-
         # Load each species into a dictionary
         for i_sp in range(self.data["species_knobs"]["nspec"]):
             species_data = CleverDict()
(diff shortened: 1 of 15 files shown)


src/pyrokinetics/gk_code/gs2.py
```diff
@@ -241,8 +234,14 @@ def get_local_species(self):
             for pyro_key, gs2_key in self.pyro_gs2_species.items():
                 species_data[pyro_key] = gs2_data[gs2_key]
 
-            species_data.vel = 0.0 * ureg.vref_most_probable
-            species_data.inverse_lv = 0.0 / ureg.lref_minor_radius
+            domega_drho = (
```
d7919: I don't think we want to use uprim here. The previous form with `g_exb` looks closer to what we'd want?

bpatel2107: If a input file has `g_exb` then that gets read directly into Numerics` here

https://github.com/pyro-kinetics/pyrokinetics/blob/995094d82afeeb5ab9ed931713ac12b20deed903/src/pyrokinetics/gk_code/gs2.py#L426-L430

And then written to file here

https://github.com/pyro-kinetics/pyrokinetics/blob/995094d82afeeb5ab9ed931713ac12b20deed903/src/pyrokinetics/gk_code/gs2.py#L575

So this part of the code is sort of bypassed when setting ExB shear.

The reasoning here is that when loading a GS2 input file in, we need to map uprim to something (if it is set) and the most appropriate thing I could see was `domega_drho`.

d7919: OK thanks, that makes sense. I'm not sure how widely used uprim is by those currently running GS2.

bpatel2107: Should I just remove it from GS2 entirely then?

d7919: It might be worth doing so and then we can see if anyone raises it as an issue/request in the future.

bpatel2107: I'll have to set `domega_drho` to 0 so when we convert between codes we don't accidentally add in the PVG term

d7919: Will `g_exb` still be set when doing the code comparison though?

bpatel2107: Yes, the ExB term is calculated either from a GK input file or the `Kinetics`/`Equilibrium` object and then stored directly in the `Numerics` object so this will be read in as normal

dake0795 (approved):
Looks good for GENE.

PR #263, opened 2023-10-18
Feature/write normalisations

bpatel2107:
Pyro now reads/writes normalisations for GS2/GENE input files. I have raised a request for this on TGLF/CGYRO but that may take a while so I wouldn't wait on that.

Found quite a few bugs along the way in normalisations as `rhoref` wasn't always set at the right time depending on how to loaded in the data. I've tidied up how data that depends on both `LocalSpecies` and `LocalGeometry` is loaded in so we can be more specific about how we load this data. 

This was branched off of  `modify/rotation_velocity` so we should probably wait on that first


dake0795 (approved):
Looks good on the GENE side.

PR #265, opened 2023-10-18
Feature/gacode reader

bpatel2107:
Add in readers for input.gacode files (https://gafusion.github.io/doc/input_gacode.html) for `Equilibrium` and `Kinetics` objects

### TODO
- [x] Add in unit tests


bpatel2107:
This is ready for review. Looks like adding in pygacode as a dependency breaks the docs building. Looks like it tries to compile fortran and fails...

bpatel2107:
@LiamPattinson @ZedThree The docs are failing her because of the `pygacode` library not building due to a lack of a fortran compiler. I tried setting `autodoc_mock_imports = ["pygacode"]` in the `docs/conf.py` file but that doesn't seem to have helped. Do you have any ideas on this?

ZedThree:
It's because `pip install pyrokinetics` will always try to install `pygacode` which fails on readthedocs.

Possible solutions:
- make `pygacode` an optional dependency, and move the `import pygacode` into the functions where it's needed
  - this might be more generally useful too, to reduce the overall pyrokinetics import time -- xarray did something similar as they were getting ludicrous import times
  - it does mean users might need to `pip install pyrokinetics[gacode]`, again, see xarray
- install gfortran on RTD
  - not sure how to do this off the top of my head, probably possible though
- fix pygacode to distribute binaries
  - I'm sure Liam and I could help with that, if they're interested :)

bpatel2107:
Would something like this work?
https://sphinx-autoapi.readthedocs.io/en/latest/index.html

It does require the latest version of `sphinx` so it may break something else. Will give it a go

ZedThree:
Sorry, this is failing a long time before running sphinx -- it's happening when installing pyrokinetics in the first place. You might be able to get RTD to install the docs dependencies separately, but I'm not sure that's the best solution

bpatel2107:
Blindly trying something worked! I switched over to a conda environment and had `pygacode` installed there. That worked fine as I think there is a compiler there. Then the rest of the proceeded as normal. @ZedThree thoughts?

See here  https://docs.readthedocs.io/en/stable/guides/conda.html#compiling-your-project-sources

LiamPattinson:
If the plan is to keep this as a non-optional dependency, I think we'll need to add something to the install section of the README to say that a Fortran compiler is required. I imagine most users of Pyrokinetics will already have one installed, but I don't think Ubuntu does by default, so it could catch out some new users.

I'm leaning towards the idea of moving some of the dependencies to optional-dependencies and moving imports inside the relevant functions. Normally I'm an "All imports at the top of the file" guy, but the import time is getting quite long. It takes about 2 seconds on my laptop, which isn't a total deal breaker, but it's noticeable compared to most Python libraries.

bpatel2107:
I've included this in the `README`. The issue of slow imports seems a bit tangential to this PR so I will raise a separate issue to tackle that but I agree the import time is getting very long now. 2 seconds is actually fast as doing `import pyrokinetics` took me over 30 seconds.

LiamPattinson (approved):

PR #267, opened 2023-10-26
Pin idspy-toolkit version to 0.4.3

bpatel2107:
Pin version of idspy-toolkits to 0.4.3 as the latest release seems to be missing a dependancy. I've raised an issue on the `idspy-toolkit` gitlab repo in the meantime

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,9 +4,11 @@ on:
   push:
     paths:
       - '**.py'
+      - 'pyproject.toml'
   pull_request:
     paths:
       - '**.py'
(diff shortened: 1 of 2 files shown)


bpatel2107:
@LiamPattinson @ZedThree  do you know if I can trigger the tests on this without changing the test workflow? Or should we just get it to run tests when the `pyproject.toml` file is changed along with any `.py` files

LiamPattinson:
I think having the tests run when the `pyproject.toml` has changed is a good idea.

LiamPattinson (approved):

PR #271, opened 2023-11-01
Update GS2 reader for next

### LINKED ISSUE #268: Update for GS2 next -- issue with `energy` grid lookup and conversion of stored input file
https://github.com/pyro-kinetics/pyrokinetics/blob/167e6824a7244fb1d8ad1db3488d3785903e1572/src/pyrokinetics/gk_code/gs2.py#L745

Tries to read `energy` to provide the energy grid data. With GS2 next we've fixed some issues around our writing of energy. This should now _always_ be `egrid` to get the energy dimension whilst `energy` is the normalised species dependent energy which is 2D.

The storage of the input file in the netcdf has also changed with our move to `neasyf` -- this currently causes problems when pyro tried to 'decode' the contents.

bpatel2107:
Fixes #268 

Update default source of `energy` grid in GS2. Note this isn't actually used for anything right now so it sort of pointless really.

Also update reader of new input file format. A simple check on the data type will allow for the new and older formats to both work.

### DIFF
--- a/src/pyrokinetics/gk_code/gs2.py
+++ b/src/pyrokinetics/gk_code/gs2.py
@@ -711,7 +711,10 @@ class GKOutputReaderGS2(FileReader, file_type="GS2", reads=GKOutput):
         else:
             # Old diagnostics (and eventually the single merged diagnostics)
             # input file stored as array of bytes
-            input_str = "\n".join((line.decode("utf-8") for line in input_file.data))
+            if isinstance(input_file.data[0], np.ndarray):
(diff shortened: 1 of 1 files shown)


src/pyrokinetics/gk_code/gs2.py
```diff
@@ -742,9 +749,9 @@ def _get_coords(
 
         # energy coords
         try:
-            energy = raw_data["energy"].data  # new diagnostics
+            energy = raw_data["egrid"].data  # new diagnostics
```
d7919: I think the comments should change here (possibly just removed entirely?)

d7919 (approved):

Issue #274, opened 2023-11-02
Optional dependencies to speed up import


bpatel2107:
Currently doing `import pyrokinetics` can take over 30 seconds, likely due to the large number of imports. We should look into having optional dependencies to speed up the import time and move imports so they are inside relevant functions

bpatel2107:
Found this nifty library to breakdown the import time of a module
https://github.com/nschloe/tuna

Running this on the unstable branch

```
python -X importtime -c "import pyrokinetics" 2> pyrokinetics.log
tuna pyrokinetics.log
```

shows that 1/3 of the import time is on `xarray` so speeding that up would be helpful if possible. Another 1/3 comes from `jetto_tools`

![image](https://github.com/pyro-kinetics/pyrokinetics/assets/15210802/a341a702-bcaa-4fd2-b926-fe7a7e5e46f6)
![image](https://github.com/pyro-kinetics/pyrokinetics/assets/15210802/fd635373-1574-4903-ae90-9b77e26631df)


Note doing this on the `feature/gacode_reader` branch adds essentially no time

ZedThree:
Might not be much we can do about xarray -- it's worth updating it though, as they have sped up their import time recently.

There's some low hanging fruit in `jetto_tools` by just not importing `IPython.embed` everywhere.

OMAS is a little trickier, in that we would need to change their build system and get them to distribute binaries instead of dynamically running cython on import. But that is possible.

Just moving the `jetto_tools` import into the functions that use it is a significant speed up for us.

Our `units` module takes a long time, but that's because we have to create the units registry. I'm not sure of a nice way around that off the top of my head.

LiamPattinson:
This is a neat tool! Interestingly, I get very different results on my own machine:

![import_time](https://github.com/pyro-kinetics/pyrokinetics/assets/14332134/805730d3-840b-45bf-acdd-955a53685fe3)

To break it down:

- Xarray 15%
- Matplotlib 11%
- pint 11%
- pint xarray 18% (?!)
- Our own units work ~15%
- jetto_tools 16%
- The rest

I'm surprised that almost half of the import time is coming just from the units. I think we should be able to get a significant speed up by moving the Matplotlib/Xarray/Jetto stuff only to the functions that use them, but I don't know about speeding up our own units. If the time is being taken up defining unit systems for each GK code, perhaps these could be defined lazily only if the user is using those codes?

[...]

PR #275, opened 2023-11-03
Move slow imports within relevant functions

LiamPattinson:
Move slow imports (xarray, matplotlib, jetto-tools) inside the functions that use them. Roughly halves the import time for the project, and the remaining time is almost entirely taken up by units stuff:

faster_imports

I found it difficult to parse some of the import sections while I was searching for the right things to edit, so I sorted the import sections with [`isort`](https://pycqa.github.io/isort/index.html) and added it to the autoformatting CI job.

I also found the new GACODE features don't work on my machine. Looks like it can't find Python headers, and I can't immediately figure out what it's trying to do. I moved it to be an optional import.

Resolves Issue https://github.com/pyro-kinetics/pyrokinetics/issues/274.

(Sorry that this changed quite a lot more than I was planning, in hindsight I could have split the changes into 3 separate commits)

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -16,24 +16,28 @@ jobs:
   black:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
       with:
         ref: ${{ github.head_ref }}
     - name: Setup Python
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
         python-version: 3.x
-    - name: Install black
+    - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install black
+        pip install black isort
     - name: Version
       run: |
         python --version
         black --version
+        isort --version
     - name: Run black
       run: |
         black src tests
+    - name: Run isort
+      run: |
+        isort src
     - uses: stefanzweifel/git-auto-commit-action@v4
       with:
         commit_message: "[skip ci] Apply black changes"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,12 +45,10 @@ dependencies = [
     "simplejson >= 3.17.6",
     "gitpython >= 3.1",
     "jetto-tools",
+    "pygacode",
 ]
 
 [project.optional-dependencies]
-gacode = [
-  "pygacode",
-]
 docs = [
     "sphinx >= 5.3",
     "sphinx_autodoc_typehints >= 1.19",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,10 +45,12 @@ dependencies = [
     "simplejson >= 3.17.6",
     "gitpython >= 3.1",
     "jetto-tools",
-    "pygacode",
 ]
 
 [project.optional-dependencies]
+gacode = [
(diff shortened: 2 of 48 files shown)


LiamPattinson:
Managed to fix the problem with `pygacode` on my system -- turns out that when I installed Python 3.10, I did `apt install python3.10` instead of `apt install python3.10-dev`, so I didn't get the necessary headers. Shall I revert the move to an optional dependency?

bpatel2107:
Yes I think having one optional dependency seems odd so I think reverting is a good idea.

ZedThree:
Hmm, it might be wise to get in touch with the GA folk to see if they would like some help from PlasmaFAIR to make `pygacode` a bit easier to install and use

bpatel2107:
Happy to send them an email about this

LiamPattinson:
Reverted the `pygacode` changes. The tests are now taking 3 times as long, so is that something we could work on? Or is that just how long `pygacode` takes?

bpatel2107:
Wouldn't have thought it would take that much longer. Could be these lines here
```python
        read_gacode = f"from pygacode import expro; expro.expro_read('{filename}', 0)"
        try:
            subprocess.run(["python", "-c", read_gacode], check=True)
        except subprocess.CalledProcessError:
            raise ValueError(f"KineticsReaderGACODE could not read {filename}")
```

Essentially when the file being read in fails, `pygacode`  will just segfault because of the fortran. To catch this kind of error I had to do this awkward test so maybe that is the issue?

LiamPattinson:
Oh, that explains why Ubuntu kept warning me of seg faults while it was running :laughing:. Yeah, it might be that it's calling that every time it tries to figure out the file type for both equilibria and kinetics (plus I think it always runs first as GACODE comes first alphabetically in both cases). We could speed it up by writing our own minimal file parser that just checks for the presence of a few features. I think that's an issue for a separate PR though.

bpatel2107 (approved):
This is much faster than before! LGTM

LiamPattinson:
Managed to halve the test time just by reordering the imports in the equilibrium and kinetics `__init__.py`. I've told isort to ignore these files, as the order in which the imported modules are listed also affects the order in which each code is tested for automatic file type inference. Code formatting tools aren't normally known for imposing performance penalties, but in Python, all things are possible.

bpatel2107:
That last commit made little difference to my version
image

LiamPattinson:
The last commit shouldn't have changed anything in terms of the import time, but it did improve the unit tests. As I'd introduced `isort` to get all the imports in the project in a consistent ordering, it swapped the order of imports in the `__init__.py` files. The issue is, lines like:

```python
from . import geqdsk  # noqa
```

have the side effect of registering file readers with all the factory business under the hood. The order in which they're registered is the order in which they're tested when we read files without providing `gk_code`, `eq_type`, or `kinetics_type`. GACODE just happens to be alphabetically first for both equilibria and kinetics, so `isort` registered it first in both cases, and the subprocess call to attempt reading the files with `pygacode` was being called a _lot_. By forcing `isort` to ignore those files and manually reordering the imports, I got the unit tests to run in half the time.

bpatel2107:
Ah I'm with you now! Will merge now

Issue #276, opened 2023-11-06
Error when using GS2 grid_option = 'range'


mehfoos:
**Brief description**: When scanning over multiple binormal wavenumbers (or toroidal modes for that matter), using [grid_option = 'range'](https://gyrokinetics.gitlab.io/gs2/page/namelists/index.html#kt_grids_knobs-grid_option) in GS2, pyrokinetics throws an error.

**Versions affected**: This happens on recent commits (including SHA 167e6824a7244fb1d8ad1db3488d3785903e1572 and 6b475bff34135a5ff8ee0aa8a7fc600e79b1c9f4), but does not seem to occur with older version (SHA 408e25a20c4d12e8ca8233b3d772504dc404f16b, for instance).

**Possible cause**: David had mentioned that it may be to do with the shape of the phase array, which should be [ntheta0, naky] in size.

**Some additional details**: The last error traced back to a file outside of library dependencies seems to be in pyrokinetics/gk_code/gk_output.py -> _normalise_linear_fields (not sure if this is the root cause; there may be issues upstream; I've pasted the error thrown when trying to run)
This only seems to happen in scanning/range simulations, and not if I just use one value for n0,or ky in GS2.
Where does it fail: It is traced to the multiplication of phase by amplitude in:
`fields[field_name] *= phase / amplitude`
In my specific case, with 30 points of binormal wavenumbers (naky = 30) I've noted some properties of phase and amplitude (phase is complex): phase.shape is (1, 30, 1, 30). amplitude.shape is (1, 30). type(phase) is <class 'pint.util.Quantity'>. type(amplitude) is<class 'numpy.ndarray'>.
This ultimately results in an error thrown by pint's quantity.py: ValueError: operands could not be broadcast together with shapes (193,1,30,246) (1,30,1,30) (193,1,30,246).

bpatel2107:
Thanks for the detailed bug report, made it much easier to see what's happening. Essentially `pyro` doesn't fully support running multiple linear simulations with `range`. 

What is happening here is that `pyro` recognises that this is a linear simulation and is attempting to normalise the output fluxes using the fields. To do that it has to multiply by a phase and divide by an amplitude to get it in the correct normalisation (which was added recently hence it working before but not now). The amplitude should be being calculated correctly but the phase doesn't seem to be. The fix should be relatively straightforward as we just need the array sizes to match up as there is a single phase/amplitude for each `ky` 

See section 5.5.3 here for details
[...]

PR #277, opened 2023-11-08
Bugfix/multiple linear ky

bpatel2107:
This allows for linear runs with multiple `ky` and `kx` to be loaded in pyro as it was previously breaking when trying to normalise/rotate the fields. Addresses #276. Due to this new format, the `growth_rate_tolerance` can now be a function of `ky` and `kx`  so the `GKOutput` format needed to change to accommodate then hence the new test data.

@mehfoos can you see if that works for you?

@LiamPattinson Not sure why tests are failing due to `pyloidal` Was there a recent release?


LiamPattinson:
We updated Pyloidal yesterday, and moved the `cocos_transform` function to a `Transform` class. If you set `pyloidal == 0.1.0` in `pyproject.toml` it should sort out the problem for now. I'll try to make a PR to update to 0.2.0 later this week.

d7919:
Are the tolerance changes needed here?

bpatel2107:
Well sort of. The changes to calculating via the `xarray` methods are needed to allow for it to be `kx` `ky` dependent. The other change is to bring it inline with the GKDB definition as it was different.

d7919 (approved):
Looks good to me although I've not had chance to actually test it out yet.

bpatel2107:
Will wait on @mehfoos to confirm but it works for me

mehfoos:
> Will wait on @mehfoos to confirm but it works for me

Thanks for tagging me to check if it's fixed. I'm traveling and will only be able to confirm when I'm back over the weekend, but I'm okay with this being closed if we're fairly confident that it has been fixed. Thank you.

mehfoos:
> Will wait on @mehfoos to confirm but it works for me

Sorry about the delay. Just tried with the test case that had failed in the past, and can confirm that it does work now. Thank you!

PR #279, opened 2023-11-09
Update pyloidal usage to 0.2.0

LiamPattinson:
Updates `Equilibrium` to use the `pyloidal` 0.2 interface. This requires switching the function `cocos_transform` to the class `Transform`, which implements the transformation factors as attributes instead of a dict.


bpatel2107 (approved):
LGTM

PR #280, opened 2023-11-10
Bugfi/eigenfunction normalisations

bpatel2107:
The eigenfunctions were being re-normalised to the total field amplitude but this is already done to the fields so it was being applied twice. This is corrected here

Eigenfunctions before
image

Eigenfunctions after
image

Can see the total amplitude roughly adds up to 1.


bpatel2107:
Initial field normalisation happens here
https://github.com/pyro-kinetics/pyrokinetics/blob/b6974417ec477ecd85a664076f797160f396053f/src/pyrokinetics/gk_code/gk_output.py#L510-L511

PR #281, opened 2023-11-16
Enhancements/kperp example

bpatel2107:
Adds the ability to calculate $k_\perp \rho_s$ from the metric terms which works reasonably well

image
image

The issue I'm facing is that when I compare to the output from GS2 (or if I manually spit out the value from CGYRO) the agreement only works when I set

$k_\perp \rho_s = \sqrt{g^{\alpha\alpha}} * k_y  * C_y = \sqrt{g^{yy}} k_y$

It should be something like 

$k_\perp \rho_s = \sqrt{ (g^{yy} * k_y)^2  + 2k_yk_x g^{xy}  + g^{xx} * k_x^2 }$

with $g^{xx} = g^{rr}$, $g^{xy} = C_yg^{r\alpha}$ and $g^{yy} = C_y^2 g^{\alpha\alpha}$

Seems like the value for `k_perp` in CGYRO is as above but the GS2 one is slightly different. We should figure out which one to "use".

@whornsby82 FYI


d7919:
What ntheta in GS2 are you using for these tests?

tests/local_geometry/test_metric_terms.py
```diff
@@ -191,3 +192,31 @@ def test_jetto_ffprime(tmp_path):
...
+    pyro = Pyro(gk_file=gs2_file)
+
+    gs2_output = Dataset(gs2_file.with_suffix(".out.nc"))
+
+    bunit_over_b0 = pyro.local_geometry.bunit_over_b0
+    theta = gs2_output["theta"][:].data
+
+    theta_even = np.linspace(min(theta), max(theta), len(theta))
+    k_perp_gs2_uneven = (
+        gs2_output["kperp2"][0, 0, :].data ** 0.5 / 2**0.5 / bunit_over_b0
```
d7919: sqrt rather than `**0.5`?

bpatel2107:
> What ntheta in GS2 are you using for these tests?

The template file had `ntheta=24` so that. The other case I show has 32

bpatel2107:
Looks like the theta spacing is the key here. Pyro matches the codes when they use equal spacing in theta (CGYRO `CONSTANT_STREAM_FLAG= 0` and GS2 `equal_arc = F` which makes senses as it uses equal spaced `theta` 

image

But they definitely all use $k_\perp \rho_s = g^{\alpha\alpha}$ which is isn't quite what I was expecting...

d7919:
> But they definitely all use k⊥ρs=gαα which is isn't quite what I was expecting...

Isn't this because you're in ballooning space with an extended theta domain? If you switch to flux tube in GS2 I would expect that you would need to use the other components to reconstruct the extended domain values correctly?

bpatel2107:
So I've reverted `metric.py` back to its original form and do the calculation in flux tube as  follows

image


The result is as follows with basically perfect agreement. Here is a case wit `ky=0.1` and `theta0=0.6pi` so pretty happy with this
image

@d7919 @hgd511 let me know if there's anything else you'd want to change

src/pyrokinetics/gk_code/gk_output.py
```diff
@@ -690,17 +690,10 @@ def _eigenvalues_from_fields(
...
-        field_amplitude = np.sqrt(np.trapz(square_fields, theta, axis=0)) / (2 * np.pi)
-        eigenfunctions = np.zeros(
-            (len(fields.coords),) + square_fields.shape, dtype=complex
-        )
+        eigenfunctions = np.zeros((len(fields.coords),) + fields.shape, dtype=complex)
+
         for ifield, field in enumerate(fields.values()):
-            eigenfunctions[ifield] = field.magnitude / field_amplitude
-        # TODO are these dims correct?
+            eigenfunctions[ifield] = field.magnitude
```
d7919: This looks like an unrelated change?

src/pyrokinetics/local_geometry/metric.py
```diff
@@ -130,17 +128,19 @@ def __init__(self, local_geometry: LocalGeometry, ntheta=None, theta=None):
 
         # poloidal average of Jacobian * g^zetazeta: <Jacobian * g^zetazeta>,
         # e.g. the denominator of equation 16
-        self.Y = (
-            integrate.trapezoid(self.Jacobian / self.R**2, self.regulartheta)
-            / self.theta_range
+        self.Y = integrate.trapezoid(self.Jacobian / self.R**2, self.regulartheta) / (
+            2 * np.pi
```
d7919: Was there any harm in keeping the theta range a calculated quantity? Flux tubes _can_ be run with > 2pi extent.

src/pyrokinetics/gk_code/gk_output.py
bpatel2107: This seems to be from #280 but not sure why its come up here

bpatel2107: Probably merged it wrong

tests/local_geometry/test_metric_terms.py
```diff
@@ -201,22 +201,18 @@ def test_k_perp(tmp_path):
...
+    nperiod = pyro.numerics.nperiod + 1
     theta0 = pyro.numerics.theta0
 
-    k_perp_pyro = pyro.metric_terms.k_perp(ky, rho, theta0)
+    theta_pyro, k_perp_pyro = pyro.metric_terms.k_perp(ky, theta0, nperiod)
+
+    k_perp_gs2 = np.interp(theta_pyro, theta_gs2, k_perp_gs2)
 
     # check within 2%
     assert np.all(np.isclose(k_perp_gs2, k_perp_pyro, rtol=2e-2))
```
d7919: Could the tolerance be tighter?

src/pyrokinetics/local_geometry/metric.py
bpatel2107: Not sure there's a harm in it but @hgd511 was keen on keeping it between -pi to pi. There's no harm in increasing it though unless I've missed something

tests/local_geometry/test_metric_terms.py
bpatel2107: That's as good as it got. 
image

Not sure what the differences are but it could some time to work it out. Could drop it to 1.5% and it would pass

bpatel2107: I could run GS2 with `equal_arc = False`, that might improve it

d7919: An error of 1.5% seems pretty good for what the test is, but it might be worth seeing if the error drops as the resolution in GS2 is increased. One could increase ntheta to improve the resolution in the calculation _and_ hopefully reduce the interpolation errors, whilst one could increase ntheta_geometry to only improve the resolution in the calculation (i.e. so you could check if it's the interpolation or the calculation at fault).

src/pyrokinetics/local_geometry/metric.py
hgd511: I think keeping the theta domain for the metric terms between -pi to pi makes equation 155 easier to understand, as one can clearly see the role of kx (~theta0) and the number of 'turns' (~2 M pi, for Mth turn). It also works for when one interprets the flux tube geometry as a single turn and multiple kx, i.e M = 0 and kx = 2 pi nx / Lx. Happy to discuss if there's a better way of doing it!

Could be worth keeping theta_range as a quantity for plotting against ballooning angle (although I assume this already exists for eigenfunctions etc).

tests/local_geometry/test_metric_terms.py
bpatel2107: `equal_arc=.false.` made no difference. 

`ntheta_geometry=1024` dropped the difference to <0.7%
image

`ntheta=64` (from 24) dropped it to <0.2%
image


Doing both `ntheta=64` and `ntheta_geometry=1024` 
image

Interestingly the increased `ntheta_geometry` drops the difference when |theta| > pi.

I imagine we could keep increasing and get better and better matches. Do you have any preference on the values used? Note this file is also the basic GS2 template so will impact the default settings of other runs.

src/pyrokinetics/local_geometry/metric.py
d7919: I was just thinking that given that a flux tube "cell" doesn't have to be 2 pi in length it might help comparisons to codes if the metric calculation can also be constructed from these "extended cells".

tests/local_geometry/test_metric_terms.py
d7919: That's perhaps a little bit suspicious -- the error got worse around theta~0 for ntheta=64 going from ntheta_geometry = 64 to ntheta_geometry = 1024, which may suggest we're not converging?

Might it be worth adding in a run with ntheta = 1024?

Of course probably don't want to go down this rabbit hole and <1% error is probably fine. I'd suggest adding ntheta_geometry = something reasonably large in the template as it shouldn't case issues for anything other than ideal_ball (which requires ntheta == ntheta_geometry currently, although I'm hoping to avoid this).

src/pyrokinetics/local_geometry/metric.py
bpatel2107: I can add back in the range but leave the default as from -pi to pi

bpatel2107:
I've increase ntheta=64 in the GS2 file and dropped the tolerance to 0.2%. Made pyro calculate the range of theta so you can put in any range of theta but the default goes from -pi to pi.

Will merge now unless any other comments

PR #284, opened 2023-11-24
Modify/remove transp template

bpatel2107:
We had 2 TRANSP netCDF template files, with the larger being 46MB which is unnecessary in term of the storage. I've removed the larger one and updated the tests to use values from the smaller one (17MB).


bpatel2107:
An alternative option would be to move the template input/output files somewhere else and have pyro pull them when needed.

PR #286, opened 2023-11-28
Add Dockerfile

fcasson:
On a system with docker daemon running, images can be built with

```
docker build . -t pyrokinetics:0.5
docker push pyrokinetics:0.5
```

The push assumes we first setup the namespace pyrokinetics on dockerhub.  I'm not sure how to do this (not sure if free) so I put examples under my username https://hub.docker.com/repository/docker/cassonf/pyrokinetics for now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,11 @@
 from python:3.9
 
 RUN apt-get update
-RUN apt-get -y install gfortran
+RUN apt-get -y install gfortran && apt-get clean
 
 COPY ./ /pyrokinetics
-RUN cd /pyrokinetics && pip install .
+RUN cd /pyrokinetics && git describe --tags > VERSION
+RUN cd /pyrokinetics && pip install --no-cache-dir . && rm -rf .git
+
+CMD [ "python" ]
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,5 +7,5 @@ COPY ./ /pyrokinetics
 RUN cd /pyrokinetics && git describe --tags > VERSION
 RUN cd /pyrokinetics && pip install --no-cache-dir . && rm -rf .git
 
-CMD [ "python" ]
+CMD [ "ipython" ]
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-from python:3.9
+from python:3.11
 
 RUN apt-get update
 RUN apt-get -y install gfortran && apt-get clean
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,8 @@
 from python:3.9
 
+RUN apt-get update
+RUN apt-get -y install gfortran
 
 COPY ./ /pyrokinetics
 RUN cd /pyrokinetics && pip install .
 
-ENTRYPOINT python3
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+from python:3.9
+
+
+COPY ./ /pyrokinetics
+RUN cd /pyrokinetics && pip install .
+
+ENTRYPOINT python3
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -1,0 +1,34 @@
+name: Publish Docker Image
+
+on:
+  workflow_dispatch:
+  release:
+    types: [published]
+
+jobs:
+  publish-docker:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Docker meta
+        id: meta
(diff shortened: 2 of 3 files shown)


fcasson:
At present the pip install does not work due to issues with gacode

```
#7 94.26 Successfully built pyrokinetics cleverdict fortranformat prominence
#7 94.26 Failed to build pygacode
#7 94.26 ERROR: Could not build wheels for pygacode, which is required to install pyproject.toml-based projects
#7 94.27 WARNING: You are using pip version 22.0.4; however, version 23.3.1 is available.
```

d7919:
I think @LiamPattinson is working on the pygacode issue.

fcasson:
This now builds and I put an image on dockerhub

Not sure what is the use case but default goes to an ipython terminal where you just do `import pyrokinetics` to get started:

```
docker run -it --rm cassonf/pyrokinetics:0.5.1-pre
```
or 
```
singularity run -i docker://cassonf/pyrokinetics:0.5.1-pre
```
Or if you want to look around in bash

```
singularity run -i docker://cassonf/pyrokinetics:0.5.1-pre /bin/bash
```

With singularity / [apptainer ](https://apptainer.org/) home dirs are mounted by default so you should be able to run a script from your home like this:

```
singularity run docker://cassonf/pyrokinetics:0.5.1-pre "python /home/userid/myscript.py"
```

With docker you would need to use -v args to mount dirs for external scripts or data into the container.

Other simple examples can be inferred by reading the docs for the python image:  https://hub.docker.com/_/python

fcasson:
> I think @LiamPattinson is working on the pygacode issue.

This was just because I needed a Fortran compiler, now fixed.

LiamPattinson:
Rather than putting images up on DockerHub, I'd recommend using the GitHub Container Registry (GHCR), as then we can manage everything just via GitHub and more easily tie the Docker releases to our GitHub/PyPI releases. I worked on a project recently that needed both Docker and Singularity images, and I came up with [this workflow](https://github.com/PlasmaFAIR/epoch_containers/blob/main/.github/workflows/publish.yml) to achieve it. In this case, I don't think we'll need a separate Singularity container, but it may be necessary if we run into directory/executable permissions issues.

fcasson:
Sounds great.  Did you just volunteer to do that?

I'm just trying to satisfy a request from the referee to add a dockerfile.  I'm not sure if putting images onto a registry was also requested, but since some people don't have ability to build docker images, it seemed useful.  I have no strong feelings about which registry to use, and what you say makes sense.

LiamPattinson:
I'd be happy to get the workflow up and running, but I think @bpatel2107 would need to set up a personal access token first. [This page](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry) describes how to work with the container registry directly, but to use the relevant bit of the workflow I posted earlier I think we'd just need two new secrets added to the project:

- `GHCR_USERNAME` -- probably just `bpatel2107`
- `GHCR_TOKEN` -- to keep things simple, I normally just generate a classic one with `write:packages` activated ([see here](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#creating-a-personal-access-token-classic) for help on creating them)

Once that's done, let me know and I'll get the automatic publishing set up. It could wait until another PR though.

Dockerfile
```diff
@@ -0,0 +1,11 @@
+from python:3.9
```
LiamPattinson: Would it be possible to bump this to Python 3.10 or 3.11? The `idspy-dictionaries` dependency recommends that we use at least 3.10, and 3.10 is the default on Ubuntu 22.04. 3.11 doesn't offer too many benefits over 3.10, but apparently it might run a little faster, which is nice.

Unfortunately 3.12 isn't an option, as they removed the `distutils` library and `pygacode` needs it to build (something I'm hoping to work on in the near future!).

LiamPattinson (approved):
Looks good besides one optional change! An auto-publish workflow could be added as its own PR.

Dockerfile
fcasson: Done

fcasson:
python:3.11 build here:

```
docker pull cassonf/pyrokinetics:0.5.1-py3.11
```

LiamPattinson:
I made a [PR](https://github.com/pyro-kinetics/pyrokinetics/pull/291) into this branch which adds a publishing workflow and a description to the `README`. It shouldn't conflict in any way, so no worries if you'd rather merge this now and add the other stuff later.

bpatel2107:
We can merge that in to this first

bpatel2107:
If everyone's happy then I can merge? @ZedThree?

ZedThree (approved):

Issue #287, opened 2023-11-29
Missing brackets on 1/2pi?


d7919:
https://github.com/pyro-kinetics/pyrokinetics/blob/7b625d7d2d497a53200969d21eef8ee74e67510b/src/pyrokinetics/gk_code/gk_output.py#L708

Should 1/ 2 * pi be `1 / (2*pi)`?

bpatel2107:
Yep that's a missing bracket
Should be 
![image](https://github.com/pyro-kinetics/pyrokinetics/assets/15210802/b978d7c7-8f9b-48bd-a109-221a5efcd209)

from https://gitlab.com/gkdb/gkdb/raw/master/doc/general/IOGKDB.pdf

Shouldn't impact eigenvalues luckily

PR #288, opened 2023-11-29
Update gk_output.py to fix issue #287

d7919:


bpatel2107:
Tests are failing due to new normalisation making the reference values out by a factor of `pi**2`. I can generate new results

bpatel2107 (approved):

PR #289, opened 2023-11-29
Update gs2.py to fix calculation of eigenvalues in non-wstar_units runs

d7919:
Slightly surprised this wasn't caught in the tests so perhaps there's some motivation to add extra tests?

### DIFF
--- a/src/pyrokinetics/gk_code/gs2.py
+++ b/src/pyrokinetics/gk_code/gs2.py
@@ -802,7 +802,7 @@ class GKOutputReaderGS2(FileReader, file_type="GS2", reads=GKOutput):
         ky = raw_data["ky"].data
 
         # time coords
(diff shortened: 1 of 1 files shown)


bpatel2107:
We can add a test comparing the pyro value to the GS2 value were we manually change it by `sqrt(2)` though I think the the template as `wstar_units = .true.`

bpatel2107:
@mgiacomin looks like you were about 30 minutes too slow

bpatel2107 (approved):

mgiacomin:
Ah sorry I hadn't seen that David pushed the fix already.

bpatel2107:
When we start storing template files on zenodo or something similar we can create more tests covering these cases but until then I can merge for now

PR #290, opened 2023-11-29
Time_divisor should be 1 when wstar_units is not used

mgiacomin:
The `time_divisor` should be 1 when `wstar_units` is `false`.  Normalisation conversion is taken care somewhere else.


bpatel2107:
This is the same as #289

PR #291, opened 2023-11-29
Docker workflow

LiamPattinson:
Adds a workflow that publishes a Docker container to GHCR whenever we make a new release. Also adds a small description of how to use the `Dockerfile` in the `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,32 @@ $ pyro --help
 $ pyro convert --help
 ```
 
+## Docker
+
+Pyrokinetics provides a `Dockerfile` from which you can build and run Docker containers.
+To do so, you must have [Docker](https://docs.docker.com/engine/install/) installed on
+your system. To build a local container:
+
+```bash
+$ docker build . -t pyrokinetics
+```
+
+It can then be run using:
+
+```bash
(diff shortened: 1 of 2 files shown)


bpatel2107 (approved):

README.md
```diff
@@ -109,6 +109,32 @@ $ pyro --help
...
+$ docker run -it --rm -v ./path/to/local:/mymount pyrokinetics
+```
+
+where:
+
+- `-it` runs an interactive shell.
+- `--rm` deletes the Docker instance after use.
+- `-v ./path/to/local:/mymount` mounts the local directory `./path/to/local` to the
+  directory `/mymount` within the Docker container.
+
```
fcasson, line 134: For me the main use case I see for the docker image would be to get all the deps quickly onto a system where I don't have root access (if I have root access, installing pyro is probably easier than running a docker build!).  To use pyro on such a system with no root access and a bad old python install, I would want to run singularity and pull an image which we already built and deployed to the registry.  So an example of that I think is useful to some users, e.g.

```
singularity run -i docker://regpath/pyrokinetics
```

Issue #285, opened 2023-11-28
GS2 geometric outputs for comparison


d7919:
Now that pyrokinetics is able to calculate the majority of relevant geometric coefficients it would be helpful if it could provide the following quantities for testing GS2's geometry calculation:

`bmag, gradpar, grho, cvdrift, gbdrift, cvdrift0, gbdrift0, gds2, gds21, gds22, rplot, zplot, aplot ,rprime, zprime, aprime, bpol`

PR #292, opened 2023-11-29
Enhancements/ideal ballooning solver + GS2 geometric terms

bpatel2107:
Calculates geometric terms needed by GS2 as requested in #285 as an option for a `Diagnostic`. This requires a few additions to ``MetricTerms`. @hgd511 feel free to rename things that you feel are somewhat mislabelled.

A test has been written comparing the GS2 outputs in the netCDF and the `Pyro` values. The following is calculated

`bmag, gradpar, grho, cvdrift, gbdrift, cvdrift0, gbdrift0, gds2, gds21, gds22, rplot, zplot, aplot ,rprime, zprime, aprime, bpol`

This shows good agreement for the template case (<3% with `gbdrift` and `cvdrift` being the limiting factor on a lower tolerance and will probably improve with a better GS2 resolution) but the agreement of `aprime` is significantly poorer for a shaped equilibria case (the rest are fine!).

The agreement for a CBC style case but not a shaped case is suspicious. Here is the template case 
image

And here is the full shaped case 
image

@d7919 do you think there could be a bug in GS2 here? I can't see anything obvious here. Use defaults for `shift`/`kappa`/`kappri`/`tri`/`tripri` did help it match. My definition could be missing something but I can't see what.

This arose as a by product from wanting to do the ideal ballooning calculation (which is unaffected by `aprime`). We used the algorithm in https://github.com/rahulgaur104/ideal-ballooning-solver to actually do the calculation. An example script is included along with some documentation. An example is shown below showing good agreement with GS2's `ideal_ball`

image


bpatel2107:
@rahulgaur104 @jfparisi @fcasson @whornsby82 FYI might be of interest

rahulgaur104:
This looks great Bhavin! I am glad that we can finally calculate the growth rate of both stable and unstable modes. 

I have invited you and Liam to look at a repo that repeats the exact same exercise with geqdsk files so we can do more comparisons if you like. I take the coefficients directly from GS2 and compare the marginal stability curve with the growth rate contours.

In a few weeks, I'll share a geometry module that can do s-alpha for a general 3D VMEC equilibrium. This would give us the ability to do gyrokinetics with tokamak and stellarator VMEC files.

On a separate note, you should consider using multiprocessing for getting the s-alpha contours, it will speed up the growth rate calculation even further. [Here's an example.](https://github.com/rahulgaur104/VMEC2GK/blob/master/src/bishoper_ball.py#L335)

src/pyrokinetics/diagnostics/__init__.py
```diff
@@ -248,3 +251,195 @@ def _invfft(f, x, y, kx, ky):
...
+    )
+
+    # Method without M is approx 3 X faster with Arnoldi iteration
+    # Perhaps, we should try dstemr as suggested by Max Ruth. However, I doubt if
+    # that will give us a significant speedup
+    w, v = eigs(A, 1, sigma=sigma0, v0=vguess, tol=5.0e-7, OPpart="r")
+    # w, v  = eigs(A, 1, sigma=1.0, tol=1E-6, OPpart='r')
+    # w, v  = eigs(A, 1, sigma=1.0, tol=1E-6, OPpart='r')
+
+    # ## Richardson extrapolation
```
rahulgaur104, line 488: Please replace Richardson extrapolation in the comment with Variational refinement. I'll do the same in my repo.

tests/diagnostics/test_gs2_geometry.py
```diff
@@ -0,0 +1,39 @@
...
+    ignore_keys = ["dpdrho"]
+
+    for key in geometry_terms.keys():
+        if key not in ignore_keys:
+            # GS2 bug in aplot
+            if key == "aplot":
+                gs2_aplot = gs2_data[key][:].data
+                gs2_theta = gs2_data["theta"][:].data
+                gs2_dpsidr = 1.0 / gs2_data["drhodpsi"][:].data
+                gs2_aplot = np.where(
```
d7919, line 31: This fix shouldn't be required once https://bitbucket.org/gyrokinetics/gs2/pull-requests/887 is merged (and specifically https://bitbucket.org/gyrokinetics/gs2/commits/ac63354fd05c42b86a899e19c9100d1adedbf3ca)

tests/diagnostics/test_gs2_geometry.py
```diff
@@ -0,0 +1,39 @@
...
+                    abs(gs2_theta) > np.pi, gs2_aplot / gs2_dpsidr, gs2_aplot
+                )
+
+                gs2_geo_term = np.interp(pyro_theta, gs2_theta, gs2_aplot)
+            else:
+                gs2_geo_term = np.interp(pyro_theta, gs2_theta, gs2_data[key][:].data)
+
+            assert_allclose(gs2_geo_term, geometry_terms[key], rtol=3e-2, atol=1e-8)
```
d7919, line 39: Does this need aprime to be ignored (or is it not included in the keys tested) given the current disagreements or is it just fine for the test case as it has default shaping?

src/pyrokinetics/diagnostics/__init__.py
```diff
@@ -248,3 +251,195 @@ def _invfft(f, x, y, kx, ky):
             )
         )
         return np.real(value)
+
+    def gs2_geometry_terms(self, ntheta_multiplier: int = 10):
```
d7919, line 255: This is great, thanks!

tests/diagnostics/test_gs2_geometry.py
bpatel2107: Fine for default shaping.

hgd511:
Because in the toroidal system, we have $\nabla \alpha = \frac{\partial \alpha}{\partial \varrho} \nabla \varrho + \frac{\partial \alpha}{\partial \vartheta} \nabla \vartheta +\frac{\partial \alpha}{\partial \zeta} \nabla \zeta$, then dotting with the gradient of a flux function $C = C \left(\varrho\right)$, we have $\nabla C = \frac{\partial C}{\partial \varrho} \nabla \varrho$ and so 

$\nabla \alpha \cdot \nabla C = \frac{\partial C}{\partial \varrho} \left( \frac{\partial \alpha}{\partial \varrho} g^{\varrho \varrho} + \frac{\partial \alpha}{\partial \vartheta} g^{\varrho \vartheta} \right)$.

If we want to divide by the magnitude of $\nabla C$ to make it a unit vector,  we have $\left | \nabla C \right | = \left |\frac{\partial C}{\partial \varrho} \right | \sqrt{g^{\varrho \varrho}}$, thus

$\nabla \alpha \cdot \frac{\nabla C}{\left | \nabla C \right |} = \mathrm{sgn}\left(\frac{\partial C}{\partial \varrho}\right) \frac{1}{\sqrt{g^{\varrho \varrho}}} \left( \frac{\partial \alpha}{\partial \varrho} g^{\varrho \varrho} + \frac{\partial \alpha}{\partial \vartheta} g^{\varrho \vartheta} \right)$.

Perhaps our current calculation of `aprime` is missing the $g^{\varrho \vartheta}$ contribution? $g^{\varrho \vartheta}$ is zero in circular but non-zero for shaping, so makes sense with what we're seeing.

bpatel2107:
Well look at that
image

src/pyrokinetics/diagnostics/__init__.py
```diff
@@ -248,3 +251,199 @@ def _invfft(f, x, y, kx, ky):
...
+
+        gamma, X_arr, dX_arr, g_arr, c_arr, f_arr = gamma_ball_full(
+            dpdrho, theta, bmag, gradpar, cvdrift, gds2
+        )
+
+        return gamma
+
+
+def gamma_ball_full(
+    dPdrho, theta_PEST, B, gradpar, cvdrift, gds2, vguess=None, sigma0=0.42
```
d7919, line 395: If one takes gds21, gds22 and cvdrift0 then you should be able to include the theta0 dependence here trivially I think with `gds2 --> gds2 + 2 * theta0 * gds21 + theta0**2 * gds22` and `cvdrift --> cvdrift + theta0 * cvdrift0`

d7919 (approved):
Looks great. 

I always get nervous by code like `a / b * c` and would rather `a * c / b` or `(a / b) * c` to help avoid errors sneaking by, but appreciate this is perhaps a style question

bpatel2107:
> This looks great Bhavin! I am glad that we can finally calculate the growth rate of both stable and unstable modes.
> 
> I have invited you and Liam to look at a repo that repeats the exact same exercise with geqdsk files so we can do more comparisons if you like. I take the coefficients directly from GS2 and compare the marginal stability curve with the growth rate contours.
> 
> In a few weeks, I'll share a geometry module that can do s-alpha for a general 3D VMEC equilibrium. This would give us the ability to do gyrokinetics with tokamak and stellarator VMEC files.
> 
> On a separate note, you should consider using multiprocessing for getting the s-alpha contours, it will speed up the growth rate calculation even further. [Here's an example.](https://github.com/rahulgaur104/VMEC2GK/blob/master/src/bishoper_ball.py#L335)

@rahulgaur104  Thanks for the prompt reply! Pyrokinetics can also read in from GEQDSK files via the FreeQDSK python library. and then directly generate the required coefficients without having to call `GS2` or any fortran code which make installation quite simple. We are also able to use other local geometry representation like MillerTurnbull, Miller eXtended Harmonic and 2 different fourier representations where we can directly calculate the geometry coefficients in python and then perform the ideal ballooning calculation for each one. We can compare the GEQDSK interpolation in GS2 to pyrokinetics but I would imagine the python one may be more robust.

I would be keen to get VMEC support into pyrokinetics so any suggestions there would be appreciated!

I think pyrokinetics should allow for the option to calculate the IBM boundary and any MPI parallelisation can be done on top of that, though I think including an example script may be a good if you have a good way of doing it already. But parallelisation would be a clear option to speed this up.

src/pyrokinetics/diagnostics/__init__.py
bpatel2107: We can add that to the function here and give `theta0` as an argument. Seems the best way to handle this.

src/pyrokinetics/diagnostics/__init__.py
```diff
@@ -248,3 +251,199 @@ def _invfft(f, x, y, kx, ky):
...
+
+        # B_mag and derivatives
+        B_mag = metric.B_magnitude
+        dB_dr = metric.dB_magnitude_dr
+        dB_dtheta = metric.dB_magnitude_dtheta
+
+        # Drifts
+        gbdrift = -2 * (dB_dr - g_rt / g_tt * dB_dtheta) / B_mag
+        press = -2 * dpdrho * (jacob / np.sqrt(g_tt)) / B_mag / dpsidrho
+        gbdrift0 = 2 * metric.dqdr * g_at / g_tt * dB_dtheta / B_mag
```
LiamPattinson: Would it be possible to reorder equations like this so that there's only 1 division?

```python
press = -2 * dpdrho * jacob / (np.sqrt(g_tt) * B_mag * dpsidrho)
gbdrift0 = 2 * metric.dqdr * g_at * dB_dtheta / (B_mag * g_tt)
```

I find it difficult to read when multiplies and divides are mixed on the same line, and you also might see some speed up since floating point divides are much slower than multiplies. If it isn't possible to do a full reordering due to broadcasting rules, brackets would still make the order of operations clearer.

src/pyrokinetics/diagnostics/__init__.py
```diff
@@ -248,3 +251,199 @@ def _invfft(f, x, y, kx, ky):
...
+        gamma, X_arr, dX_arr, g_arr, c_arr, f_arr = gamma_ball_full(
+            dpdrho, theta, bmag, gradpar, cvdrift, gds2
+        )
+
+        return gamma
+
+
+def gamma_ball_full(
+    dPdrho, theta_PEST, B, gradpar, cvdrift, gds2, vguess=None, sigma0=0.42
+):
```
LiamPattinson, line 396: Please could we add type hints and docstrings before the merge?

LiamPattinson (commented):
I can't comment on the science-side of this, but it all looks really impressive! I just have a couple of requests for the code -- sorry if these are already on your radar!

I've had a scan through the changes, and the only sections I can see that would be suitable for a multiprocessing speed up are in the example file -- @bpatel2107 @rahulgaur104 is that where you mean? If so,  I think [`tqdm.process_map()`](https://tqdm.github.io/docs/contrib.concurrent/) would be a good option, as it provides a nicer interface to Python multiprocessing and even gives you a progress bar. Let me know if you'd like me to have a go at this.

src/pyrokinetics/diagnostics/__init__.py
rahulgaur104: @LiamPattinson  I hope this helps

    """
    Ideal-ballooning growth rate finder.
    This function uses a finite-difference method
    to calculate the maximum growth rate against the
    infinite-n ideal ballooning mode. The equation being solved is
    d / d theta (g d X / d theta) + c * X - lam * f * X = 0, g, f > 0
    where
    g = gradpar  * gds2  / B,
    c =  (1/ gradpar) * dPdrho  * cvdrift,
    f = gds2 / B^3 * 1/gradpar,
    are needed along a field line to solve the ballooning equation once.

    gds2 = dpsidrho**2 * |grad alpha|^2, alpha = phi - q (theta - theta_0) is the field line bending term
    gradpar = b dot grad theta_PEST is the parallel gradient,
    cvdrift = dpsidrho**2 * (b cross (b dot grad b)) dot grad alpha, is the curvature drift
    Params
    ---------
     dPdrho: float
    Gradient of the total plasma pressure w.r.t the radial coordinate rho
     theta_PEST: numpy array N
    Vector of grid points in a straight field line theta (PEST coordinates)
      B: numpy array size N
     The magnetic field strength
      gradpar: numpy array size N
      parallel gradient 
      cvdrift: numpy array size N
     Geometric factor in curvature drift
      gds2: numpy array
     Field line bending term
      vgiess: numpy array size( N-2) 
     starting guess for the eigenfunction
      sigma0: float
     starting guess for the eigenvalue (square of the growth rate). 
     Must always be greater than the final value. Choose a large value

      Returns
      ------------
     gam: Float
     The maximum value of the square of the ballooning growth rate

  """

LiamPattinson:
I've played around a little, and it looks like the code should already be partially running in parallel. On my machine, it seems that the calls to SciPy's `eigs` (which just passes things to ARPACK) are multithreaded. However, the serial parts of the code dominate, and this doesn't actually provide much speed-up.

To get around this, we can set `OMP_NUM_THREADS=1`, and then use `tqdm` to do Python-level multiprocessing in the examples file. I have a branch [here](https://github.com/pyro-kinetics/pyrokinetics/blob/enhancements/ideal_ballooning_solver_parallel/docs/examples/example_ideal_ballooning.py) that demonstrates it, and I was able to get around a 3x speed up with this. However, if you don't set `OMP_NUM_THREADS=1`, you instead get a ridiculous slow-down due to the competition for system resources (on my machine, 7 seconds without multiprocessing, 2.5 seconds with multiprocessing and `OMP_NUM_THREADS=1`, 4 minutes with multiprocessing and `OMP_NUM_THREADS` unset).

Let me know if you'd like me to make a PR into here.

rahulgaur104:
Yes, you have to set the number of threads to one otherwise it ends up in GIL (Global Interpreter Lock) because it tries to multithread. You can prevent this from your python code by using
`os.environ["OMP_NUM_THREADS"] = "1"`
I have obtained a speedup of ~50x with 92 cores on a cluster so scaling shouldn't be a problem.

src/pyrokinetics/diagnostics/__init__.py
LiamPattinson: Cheers, that'll do it! It'll need a little ReSTructued text formatting, but otherwise looks great.

src/pyrokinetics/diagnostics/__init__.py
bpatel2107: Yep that's fine

LiamPattinson:
> Yes, you have to set the number of threads to one otherwise it ends up in GIL (Global Interpreter Lock) because it tries to multithread. You can prevent this from your python code by using `os.environ["OMP_NUM_THREADS"] = "1"` 

Thanks, that's good to know. I didn't realise it still worked by setting it within a Python script, but it turns out it just needs to be before the import of NumPy/SciPy.

src/pyrokinetics/diagnostics/__init__.py
bpatel2107: Added in

src/pyrokinetics/diagnostics/__init__.py
bpatel2107: Do any other terms change with `theta0`? Maybe put this into `get_gs2_geometry` is more appropriate?

bpatel2107:
I think the optimisation can be done in a separate PR as we are moving away from the realm of just getting it to work! A separate issue would be a good place to discuss.

src/pyrokinetics/diagnostics/__init__.py
rahulgaur104: No. For ballooning and local gyrokinetic solvers, it's just gds2 and cvdrift. The function here assumes that you are adding the theta0 terms before passing gds2 and cvdrift to the solver.

bpatel2107: Okay I'll put it in `get_gs2_geometry` then!

PR #293, opened 2023-12-06
Bugfix/docker publishing issues

LiamPattinson:
Fixes for the Docker publishing workflow.

[docker/build-push-action](https://github.com/docker/build-push-action) does not make use of the project's working directory, but instead gets its own copy of the project via the [build command](https://docs.docker.com/engine/reference/commandline/build/#git-repositories). To minimise the size of the resulting image, this does not copy across the full commit history in `.git`, with the side-effect that `setuptools-scm` can't figure out the Git tags at build time.

This fix sets the `ARG` `SETUPTOOLS_SCM_PRETEND_VERSION`, which overrides the local version in the Docker file if set like so:

```bash
$ docker build . --build-arg SETUPTOOLS_SCM_PRETEND_VERSION=v0.6.0 -t pyrokinetics
```

If building locally and the `--build-arg` is excluded, `setuptools-scm` will grab the version from the Git tags as usual. In the CI job, the build arg is set using `${{ github.ref_name }}`, and on new releases this will be set to the release tag.

Also contains some edits to the `Dockerfile`  that I made while trying to find alternative solutions. I think it's a little easier to read now, so I left the changes in.

I wasn't able to verify that this definitely works, as `${{ github.ref_name }}` resolved to my branch name in testing, and we'll need to make a new release to be sure.

### DIFF
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -2,13 +2,13 @@ name: Publish Docker Image
 
 on:
   workflow_dispatch:
+  push:
   release:
     types: [published]
 
 jobs:
   publish-docker:
     runs-on: ubuntu-latest
-
     steps:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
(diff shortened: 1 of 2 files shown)


ZedThree (approved):
Might be wise to try and get the action to run on PRs (and only push on release?)

ZedThree:
Oh, these tests are "required" but only run on changes to `**.py`

LiamPattinson:
> Might be wise to try and get the action to run on PRs (and only push on release?)

Good idea, I've now changed the workflow so that it runs on push, pushes the Docker container only on release, and sets a fallback version on other events.

bpatel2107:
Does this workflow trigger on any push?

LiamPattinson:
Yes, but only to check that a docker image can be built using GitHub's test runners. It shouldn't publish it anywhere unless we make a new release. It would also be really easy to disable it on pushes if you'd prefer to reduce the number of workflows being triggered.

bpatel2107:
Do we need to re-release 0.6.0 then?

ZedThree:
No, we can just do 0.6.1

PR #294, opened 2023-12-06
wip: highk/dbs synthetic diagnostic

juanruizruiz:
This is the initial work on the implementation of the synthetic highk/dbs diagnostic. More details to come :-)


bpatel2107:
@quinntpratt @jcandy FYI this may be of interest to you!

bpatel2107:
@LiamPattinson @ZedThree I think this works as but just need tidying up. Any thoughts on that would be appreciated.

@juanruizruiz Would be useful to think of a way to test this so we know if someone breaks it in the future.

src/pyrokinetics/diagnostics/synthetic_highk_dbs.py
```diff
@@ -0,0 +1,670 @@
...
+        eps = rho / R00
+
+        # compute k-resolutions (approx at the OMP): cf. Ruiz Ruiz PPCF 2020, eq. (13)
+        if syn_filter == "gauss":
+            self.dkx0 = 2 / (wR * np.sqrt(grr0)) * rhos_sim / 1e2
+            self.dky0 = 2 * kappa * q / (wZ * dalpha_dtheta0) * rhos_sim / 1e2
+            # elif filter == 'beam':
+            # we need further work on this
+
+    def get_syn_fspec(self, ik, t1, t2, savedir, if_save):
```
bpatel2107: You can probably just have this run on all the `ky` at once. I think utilising the built in xarray methods would be a good idea too

ZedThree:
Can we convert the example into a test?

Some quick thoughts:
- get rid of the `print` statements -- useful for development and debugging, less useful for production. Maybe add a `__str__` method, or something like `describe/summary` that returns a string to give an overview of the object
- using more xarray objects might simplify most of the plotting. The plotting is also better off in separate routines to the calculations
- add a few helper functions to simplify repeated patterns, for example `np.sum(np.sum(np.abs(phikxkyf) ** 2 * F2, 0), 0)` could be wrapped up into a function that takes `phikxkyf` and `F2`

src/pyrokinetics/diagnostics/synthetic_highk_dbs.py
```diff
@@ -0,0 +1,670 @@
...
+import xarray as xr
+import xrft
+from numpy.typing import ArrayLike
+
+import pyrokinetics as pk
+from pyrokinetics import Pyro, template_dir
+from pyrokinetics.typing import PathLike
+
+
+class Synthetic_highk_dbs:
```
ZedThree: More pythonic name would be something like:

```suggestion
class SyntheticHighkDBS:
```

src/pyrokinetics/diagnostics/synthetic_highk_dbs.py
```diff
@@ -0,0 +1,670 @@
...
+    simdir: PathLike    #
+    savedir: PathLike   #
+
+    # Steps:
+    # 1. Inputs are diagnostic specific (diagnostic, filter, k, location, resolution, local rhos)
+    # 2. Load equilibrium, kinetics files
+    # 3. Map (kn, kb) to (kx, ky) for all k's / channels specified in 1.
+    # 4. Load GK output data
+    # 5. Filter sim data : synthetic spectra
+
```
ZedThree: Please could you expand this docstring to include a description of the purpose of the class, along with what all the arguments and attributes/members mean?

Look at [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) for the preferred docstring style for pyro.

src/pyrokinetics/diagnostics/synthetic_highk_dbs.py
```diff
@@ -0,0 +1,670 @@
...
+        self.pyro = pyro
+        self.eq = pk.read_equilibrium(eq_file)
+        meter = pyro.norms.units.meter
+        self.psin = self.eq._psi_RZ_spline(Rloc * meter, Zloc * meter) / (
+            self.eq.psi_lcfs - self.eq.psi_axis
+        )
+        pyro.load_local(psi_n=self.psin, local_geometry="Miller")
+        self.geo = pyro.local_geometry
+        pyro.load_metric_terms()
+        self.m = pyro.metric_terms
```
ZedThree: Try not to abbreviate things too much:

```suggestion
        self.geometry = pyro.local_geometry
        pyro.load_metric_terms()
        self.metric = pyro.metric_terms
```

src/pyrokinetics/diagnostics/synthetic_highk_dbs.py
```diff
@@ -0,0 +1,670 @@
...
+        self.thetaloc = thetatmp[tmp_ind]  # np.interp(Zloc, Ztmp, thetatmp)
+
+        print("psin = " + str(self.psin))
+        print("theta/pi = " + str(self.thetaloc / np.pi))
+        print("Rloc = " + str(Rloc) + " [m]")
+        print("Zloc = " + str(Zloc) + " [m]")
+
+    def mapk(self):
+        import matplotlib.pyplot as plt
+        import numpy as np
```
ZedThree: Imports are usually better at the top of the file

src/pyrokinetics/diagnostics/synthetic_highk_dbs.py
```diff
@@ -0,0 +1,670 @@
...
+
+        # maps k from (kn, kb) to k in GK code : (kR, kZ) in the future
```
ZedThree: This is better off as a docstring

src/pyrokinetics/diagnostics/synthetic_highk_dbs.py
```diff
@@ -0,0 +1,670 @@
...
+        # in pyrokinetics, (kx, ky) is the same as CGYRO: kx=2*pi*p/Lx, ky=n*q/r
+
+        thetaloc = self.thetaloc
+        kn0 = self.Kn0_exp
+        kb0 = self.Kb0_exp
+        geo = self.geo
+        syn_filter = self.syn_filter
+        m = self.m
+        wR = self.wR
+        wZ = self.wZ
```
ZedThree: I don't think these are modified locally, so no need to make local copies, just use `self.*` directly

src/pyrokinetics/diagnostics/synthetic_highk_dbs.py
```diff
@@ -0,0 +1,670 @@
...
+        dalpha_dtheta = m.dalpha_dtheta
+        dalpha_dr = m.dalpha_dr
+        Jr = m.Jacobian
+
+        dRdr = m.dRdr
+        dRdtheta = m.dRdtheta
+        dZdr = m.dZdr
+        dZdtheta = m.dZdtheta
+        R = m.R
+        q = m.q
```
ZedThree: Again, I don't think you need to make local copies here

src/pyrokinetics/diagnostics/synthetic_highk_dbs.py
```diff
@@ -0,0 +1,670 @@
...
+        thetatmp = self.geo.theta[self.geo.Z > self.geo.Z0]
+        Rtmp = self.geo.R[self.geo.Z > self.geo.Z0] * self.a_minor  # [m]
+        Ztmp = self.geo.Z[self.geo.Z > self.geo.Z0] * self.a_minor  # [m]
+        tmp_ind = np.argmin(np.abs(Rtmp - Rloc))
+        self.thetaloc = thetatmp[tmp_ind]  # np.interp(Zloc, Ztmp, thetatmp)
+
+        print("psin = " + str(self.psin))
+        print("theta/pi = " + str(self.thetaloc / np.pi))
+        print("Rloc = " + str(Rloc) + " [m]")
+        print("Zloc = " + str(Zloc) + " [m]")
```
ZedThree, line 120: Rather than calling `print` here, add a `__repr__` method (for a "plain" representation that could recreate the object) or `__str__` (for a "nicer" representation).

Also, rather than `"string" + str(something)` use an f-string:

```suggestion
        print(f"psin = {self.psin}")
        print(f"theta/pi = {self.thetaloc / np.pi}")
        print(f"Rloc = {Rloc} [m]")
        print(f"Zloc = {Zloc} [m]")
```

src/pyrokinetics/diagnostics/synthetic_highk_dbs.py
```diff
@@ -0,0 +1,670 @@
...
+        plt.title("Avg. " r"$-k_x \ \tilde{P}_s(f)$, ch = " + str(ik), fontsize=fsize)
+        plt.xlabel(r"$\omega \ [a/c_s]$", **axis_font)
+        plt.tick_params(labelsize=fsize)
+
+        if if_save:
+            if not os.path.exists(savedir):
+                os.mkdir(savedir)
+                os.chdir(savedir)
+            else:
+                os.chdir(savedir)
```
ZedThree: I would suggest not to call `os.chdir` in functions, it can be quite surprising when used in scripts!

In fact, I would suggest replacing `savedir` with something like `save_as` or `filename` and just writing straight to that file. This way the user has full control over where the graphs are saved, as they can include any directory in the filename -- it does mean the user is responsible for ensuring the directory exists, but I think that's expected.

src/pyrokinetics/diagnostics/synthetic_highk_dbs.py
LiamPattinson: We recently moved Matplotlib inside only the functions that use it to reduce the import time, so I'd argue to remove that from the top of the file and leave it inside the plotting functions, but most other libraries like NumPy should be at the top.

bpatel2107 (approved):
Overall this is working as expected. There are some smaller things that could be re-named as such but some parallel development is being planned and it will be useful to have this in the next release. 

A future PR will be opened on detailed documentation for this which can address some of these issues.

PR #296, opened 2024-02-22
Enhancements/merge species

arkabokshi:
The idea is merge various kinetic species - such that it preserves quasi-neutrality - into a single specie.  The basic algorithm is as follows ($\sum$ is over all the species $s$ we wish to merge):

- effective charge $Z_m = \sum Z_s n_s / \sum n_s$
- effective number density $n_m = \sum Z_s n_s / Z_m$
- effective mass $M_m = \sum M_s n_s / \sum n_s$
- effective inverse length scale $1/L_{n_m} = (\sum Z_s n_s (1 / L_{n_s}) )  /  (Z_m n_m)$

The script asks the user to enter the species they wish to merge and performs the merge.  Is it better to have it as an example script or embed it more deeply into Pyrokinetics?


arkabokshi:
Alternatively, we can perform the merge action on the input files generated by Pyro?

docs/examples/example_merge_species.py
```diff
@@ -0,0 +1,77 @@
+from pyrokinetics import Pyro, template_dir
+from pint import UnitRegistry
+import numpy as np
+
+# data path
+spr_path = '/home/ab1209/pyrokinetics/SPR-JETTO/qmin_264_imp'
```
d7919: This might work better as an executable example where we provide relevant equilibrium and profile files as with tests/other examples so users can expect the example script to work unlike here where you've hard-coded a path from your local system.

docs/examples/example_merge_species.py
```diff
@@ -0,0 +1,77 @@
...
+# inform species
+for count,species in enumerate(kin.species_data.keys()):
+
+    charge = kin.species_data[f'{species}'].charge(psi_n*ureg.dimensionless)
+    print(f'{species}, Z: {charge.magnitude}')
+
+
+# check with users species to merge
+spec_list = []
+nspec = int(input("Enter number of species to merge : "))
```
d7919: Fine for an example, but we might want this functionality to be scriptable -- move the functionality into a function which takes all required input as an argument. That way example scripts like this can still read user input (although perhaps better using an argument parser than relying on input interaction) but the code can easily be reused in automated workflows.

d7919:
This PR adds a single example file but ends up touching a whole bunch of other files just with black formatting changes. I guess this is automatically applied, just a bit surprising that it has found changes to make on files you've not touched!

docs/examples/example_merge_species.py
d7919: If this could reuse existing example files then that might be even nicer (and definitely shouldn't add the files that you're currently referencing!)

d7919:
I'm a little torn on if this should be an example / tutorial vs built-in functionality. Being built-in may make it easier to build clear workflows and ensure pyro can continue to work with the system (i.e. knows what the new profiles are etc.) but it then starts to move pyrokinetics away from the current state where one can (hopefully) be convinced that the input files generated by pyro are a consistent representation of the source equilibrium.

docs/examples/example_merge_species.py
```diff
@@ -0,0 +1,77 @@
...
+print('Enter species name, e.g. electrons')
```
d7919: Electron probably isn't the best example (will you ever want to merge electrons with anything?)

docs/examples/example_merge_species.py
```diff
@@ -0,0 +1,77 @@
+from pyrokinetics import Pyro, template_dir
```
d7919, line 1: You don't use this

arkabokshi:
Thanks for the comments @d7919.  Not sure why it touched the formatting on the other files.

I am personally inclined to make this an example/tutorial. This is something the users must decide to do and further check themselves if the _difference_ the merge introduces is acceptable in their scenario. I will fold in your comments and make this an example for now.  If people feel strongly otherwise, it can be extended.

docs/examples/example_merge_species.py
bpatel2107: Yes agreed about using existing template files

docs/examples/example_merge_species.py
bpatel2107: Adding a method to the kinetic species object seems like the cleanest way to do this so its built into pyro

bpatel2107:
I agree with @d7919 as this is likely something to be done a lot so should be done in a consistent way. Pointing out changes is of course needed and the user can decided then but having it as an example opens up paths to user error and makes testing more complicated.

LiamPattinson:
I've had a go at implementing a `merge_species` function on `LocalSpecies`, though I know there are a few issues:
- I don't think taking the weighted density average over the collision frequency is correct, but I'm also not sure how else to handle it.
- Also not sure if I'm handling the gradients properly.
- Needs a unit test and a short example written up for the docs

bpatel2107:
The only things we should do the weighted average over is the density and density gradient. I think the best way of doing this would be to have a `base_species` that the other species are merged into. Weighted averages of other terms don't really make sense and generally that is how people would want to use it.

Ideally we would have a kwarg for the `base_species` as a str and the kwarg for a list of species to be absorbed by the base. Something like

```python
    def merge_species(self, base_species: str, merge_species Iterable[str]) -> None:
        """
        Merge species listed in `merge_species` into the `base_species`. Performs a weighted average depending on the
        densities of each species to preserve quasineutrality
        Parameters
        ----------
        base_names
            Names of species that will absorb other species
        merge_species
            List of the species names to be merge.
        Raises
        ------
        ValueError
            If there is no species with a given name.
        """
        if base_species not in self.names:
            raise ValueError(f"Unrecognised base_species name {base_species}")

        unrecognised = [name for name in merge_species if name not in self.names]
        if unrecognised:
            raise ValueError(f"Unrecognised merge_species names {', '.join(unrecognised)}")

        new_dens = (self[base_species].dens * self[base_species].z + sum(self[name].dens * self[name].z for name in merge_species)) / self[base_species].z

        new_inverse_ln = (self[base_species].dens * self[base_species].z * self[base_species].inverse_ln + sum(self[name].dens * self[name].z * self[name].inverse_ln for name in merge_species)) / (self[base_species].z * new_dens)

        self[base_species].dens = new_dens
        self[base_species].inverse_ln = new_inverse_ln

        self.remove_species(*names)
        self.check_quasineutrality()
```


@arkabokshi @d7919 thoughts? You guys have done this more. Maybe we can give the option to change `z` and `mass` but take the `base_species` by default?

d7919:
I think being able to specify the charge and mass would be helpful. Consider trying to merge a 50:50 D-T system -- probably just want a single DT species with mass of 2.5 times the proton mass? Similarly, merging two different ionisation states of the same impurity might be something we would want to do?

Perhaps possible to calculate the density weighted average mass and charge as one option? Would it also make sense to doing something similar for the temperature (e.g. work out the thermal energy density of the species to be merged and then ensure the merged species has the same)?

Probably need to check that base_species is not in merge_species. 

Would it be any cleaner to return a species instance representing the merge?

arkabokshi:
@bpatel2107  Do we need documentation for this?

arkabokshi:
I think the charge state `z` _has_ to be updated to ensure quasi-neutrality. `mass` is probably optional, but makes sense to update it like @d7919  points out.  `temp` and `inverse_lt` are optional, but perhaps not needed at this stage?  We _typically_ deal with only three temperatures (that of electrons, ions and alphas) and it is unusual to merge any of these?

d7919:
> I think the charge state `z` _has_ to be updated to ensure quasi-neutrality. `mass` is probably optional, but makes sense to update it like @d7919 points out. `temp` and `inverse_lt` are optional, but perhaps not needed at this stage? We _typically_ deal with only three temperatures (that of electrons, ions and alphas) and it is unusual to merge any of these?

It isn't _necessary_ to update `z`  -- one could decide to treat all impurities as Carbon, for example and would need to simply adjust the density of the resulting species to represent the same total charge density as the merged species. It's not clear that this is a common use case, but it's certainly possible. Perhaps a more likely example is where we start with multiple charge states of a heavy impurity and we want to set up a case where we assume the impurity is now fully ionised -- here we could have the fully ionised species as the base one and the partially ionised as the 'merge_species', whilst another case would be to merge all states into a single species with an effective charge. I guess the point is that there are different approaches that we might want to support and the question is then how do we best support this consistently -- do we just take the properties from the base species, do we add flags to activate different treatments, do we require the user to "fix up" after the merge, ....?

bpatel2107:
I would say people would prefer to have a species with a physical `z`/`mass` and then change the `dens`/`inverse_ln` to ensure quasi-neutrality. An averaged `z` and `mass` would result in a non-physical species and generally people would want all species to represent an actual element with the fudge happening in the density gradient. 

It is definitely possible to do this but you just need to be slightly more careful when modifying the density and its gradient.

I would definitely not touch anything that doesn't have to be modified to maintain quasineutrality (like `temp`/`inverse_lt`). People can just manually change that if they want.

bpatel2107:
Also we always check for quasineutrality at the end so we have to make sure it is satisfied when doing this. I don't want a method to generate a non-quasineutral plasma

fcasson:
Disagree. Common use case is we have one or two average impurities which represent all the others. This does not need to have a real Z or A, but some codes like NEO only accept integer Z which is a pain

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

arkabokshi:
Flags are certainly easy to add.  Consider the two example use cases:

- 50-50 D-T mix:  Should the user provide a flag to update the mass or should it simply return 2.5 proton mass?  I think the default behaviour should at least be to do a density average.
- 50-50 Be-C mix:  What should be the default behaviour?

bpatel2107:
Okay so there is a lot of contention here so flexibility is the best option. I propose the following. Have the following options as kwargs: `new_z`, `new_mass` which have default as `None`.

When left as `None` we take the weighted average of charge and mass. Easiest to do this by default as it can be fiddily and setting these directly to a species is straight forward. If `new_z` is set then we have that value be used when calculating the quasineutrality. The mass isn't involved in the quasineutrality but we may as well set it here too.


```python
    def merge_species(
        self,
        base_species: str,
        merge_species: Iterable[str],
        new_z: float = None,
        new_mass: float = None,
    ) -> None:
        """
        Merge multiple species into one. Performs a weighted average depending on the
        densities of each species to preserve quasineutrality.

        Parameters
        ----------
        base_species:  str
            Names of species that will absorb other species
        merge_species: Iterable[str]
            List of species names to be
        new_z: float
            Charge of new species - Default of None results in weighted average
        new_mass: float
            Mass of new species - Default of None results in weighted average

        Raises
        ------
        ValueError
            If there is no species with a given name.
        """

        if base_species not in self.names:
            raise ValueError(f"Unrecognised base_species name {base_species}")

        unrecognised = [name for name in merge_species if name not in self.names]
        if unrecognised:
            raise ValueError(
                f"Unrecognised merge_species names {', '.join(unrecognised)}"
            )

        if base_species not in merge_species:
            merge_species.append(base_species)

        if new_z is None:
            new_dens = sum(self[name].dens for name in merge_species)
            new_z = (
                sum(self[name].dens * self[name].z for name in merge_species) / new_dens
            )

        else:
            new_z *= self[base_species].z.units
            new_dens = (
                sum(self[name].dens * self[name].z for name in merge_species) / new_z
            )

        new_inverse_ln = sum(
            self[name].dens * self[name].z * self[name].inverse_ln
            for name in merge_species
        ) / (new_dens * new_z)

        if new_mass is None:
            new_mass = (
                sum(self[name].mass * self[name].dens for name in merge_species)
                / new_dens
            )
        else:
            new_mass *= self[base_species].mass.units

        self[base_species].dens = new_dens
        self[base_species].z = new_z
        self[base_species].inverse_ln = new_inverse_ln
        self[base_species].mass = new_mass

        merge_species.remove(base_species)

        self.remove_species(*merge_species)
        self.update_pressure()
        self.check_quasineutrality()

```

This seems like the best compromise @arkabokshi @fcasson @d7919 thoughts?

d7919:
Thanks @bpatel2107 that looks good to me (note slight typo on merge_species doc string if planning on copy-pasting to a commit).

arkabokshi:
Thanks @bpatel2107, yes looks like a good solution.  Seems we need tests on the new methods as well.  Shall complete these in the second half today.

LiamPattinson:
I've added unit tests for `LocalSpecies.merge_species()` that cover all paths through the function, and slightly changed the logic:

- If the user adds the same species names to `merge_species` multiple times, the duplicates are discarded rather than preferentially weighting those species.
- `check_quasineutrality()` now returns a `bool`, as it's useful to be able to check this in the tests without having to detect whether a warning has been raised.

I chose to merge two carbon isotopes to keep the maths easy, but unfortunately it means that the `keep_base_species_z=True/False` tests aren't as informative as they could be. It might be worth altering the test to use two different elements instead.

arkabokshi (commented):
Looks good @LiamPattinson, thanks! Small typo in the last test. It should be `merge_species` and not `base_species` in the docstring?

arkabokshi:
Testing deuterium and carbon12 merge. Note that `keep_base_species_z = True` does not maintain the total pressure.

bpatel2107:
When we have `keep_base_species_z=False` everything works as it should.

I see the issue with the pressure when `keep_base_species_z=True`

Do we think that we should maintain the full pressure when we merge things? Right now what we do is maintain the level of quasi-neutrality rather than the total pressure. 

We could maintain the total pressure by doing this when `keep_base_z=True`

```python
            new_dens = (self.pressure - (
                sum(self[name].dens * self[name].temp if name not in merge_species else 0 for name in self.names))
            ) / self[base_species].temp
```

This maintains total pressure but will changes the level of quasineutrality.

The total pressure is rarely used though. We could add this as an option and leave as False by default. Thoughts @arkabokshi? I've coded it up so I can push this if we think this is worth it.

arkabokshi:
@bpatel2107  If we want to maintain the pressure and quasineutrality, should we instead adjust the base species temperature?  Regardless, we should flag this in the documentation with admonitions I feel.

bpatel2107:
Hmm this is getting a bit too complicated. I feel like adjusting the temperature starts taking us away from the whole point of this (which was to maintain quasi-neutrality when removing a species). I think the best thing is to leave this as is and stress the change in pressure when we set `keep_base_z=True`.

Anything beyond that the user will have to figure out on their own I think.

bpatel2107:
Okay I'm happy to merge now

bpatel2107 (approved):

PR #297, opened 2024-03-04
Replace "parameters" with "knobs" in GS2

d7919:
The parameters namelist in GS2 has been deprecated for a while and v8.2 will be removing this. Currently we attempt to get/set beta and zeff using parameters. This commit changes that.

We could probably add a helper similar to `_get_beta` for zeff as well. Probably not much work to add something like

~~~~python
def _get_knobs_parameters(self, var, default = 0.0):
      has_parameters = "parameters" in self.data.keys()
      local_default = default
      if has_parameters:
               local_default = self.data["parameters"].get(var, local_default)
      return self.data["knobs"].get(var, local_default)
~~~~

It would even be possible to take a list of namelist to search in order if we wanted to generalise further (but likely no need).


bpatel2107 (approved):
This looks good to me. Not sure why it reformatted the other files though...

PR #301, opened 2024-03-12
Enhancements/imas update v3.4

bpatel2107:
Update pyrokinetics to read/write the newer `idspy_toolkit=0.6.3` and `idspy_dictionaries=3400.2.0`

~Failing due to a weird git repo issue with it not being able to get the git hash of the project.~
This is passing all but one test due to an issue on the `idspy_toolkit` end when writing a namelist with an empty section.

Currently the full linear data and the full nonlinear fields is being stored. Only the time-averaged fluxes are stored but if the data dictionary add in this function then we can update easily. This also is the case for the nonlinear moments which are not supported at all.


bpatel2107:
@ycamenen would you be able to test this with a case you already have an IDS output in HDF5 format? We support GENE/CGYRO/GS2 so any examples of those would be fine I think.

arkabokshi:
@bpatel2107 There may be a cleaner way of doing this?

bpatel2107:
Okay so this PR is good to go (once again a big PR)!

Thanks to @ycamenen I have an example IMAS IDS file for GKW which I compare to the reading in the same data via the inputs files directly our testing and we get a perfect match! Furthermore I have a roundtrip test. Pyro -> IDS -> Pyro for GKW and that also matches so we should be good to go. I also read the GKW data in, write the IDS and then compare to the example IDS we have.

There are a few points of note

- Dropped testing an support of python 3.8
- Currently 1 test is failing due to the value of the MXH shaping coefficients in the template IDS which I believe may not be correct. 
- There was a small bug in the linear outputs of fluxes/moments where they were not being re-normalised to the full field amplitude squared $|A_f|^2$ (the fields were correctly handled). This meant that we were off by a constant factor for each of these so I have had to re-generate the golden answers. But with the match to the IMAS IDS we should be good now
- Given that you might not have the minor radius when loading in GKW output, I've included an option to set the GKOutput convention when loading in data. So you can do `pyro.load_gk_output(output_convention="GKW")` and it won't try and convert it to the pyro Convention. You could also do something like `pyro.set_ref_ratios(aspect_ratio=3)` before loading the output which would let you convert to the pyro norm.
- We likely need to double check the definition of `nu_ab` in the GKW input file. In pyro we have

$\nu_s = \frac{\sqrt{2}\pi n_s Z_s^4 ln(\Lambda)} {(4\pi\epsilon_0)^2 T_s^{1.5} m_s^{0.5}} \frac{L_{ref}}{v_{ref}}$

GKW has this which is very close to what we want but out by a factor of $\sqrt{\frac{T_s}{T_{ref}} \frac{m_{ref}}{m_s}}$
image

I added that factor in when mapping but could use another eye on it, @arkabokshi what do you think? I do match the value in the IDS so I am inclined to believe that is correct.

We should merge #311 before merging this (or merge this into #311)

ZedThree:
Probably best get #311 into main first before this. Given the size of the two PRs, I'd definitely advise against merging #311 directly into this.

Could we pull out the GKW changes into a separate PR? It adds a _lot_ of files, so it would probably be easier to review as two separate PRs. Or maybe there's something else we can do about all the binary files?

bpatel2107:
> Probably best get #311 into main first before this. Given the size of the two PRs, I'd definitely advise against merging #311 directly into this.
> 
> Could we pull out the GKW changes into a separate PR? It adds a _lot_ of files, so it would probably be easier to review as two separate PRs. Or maybe there's something else we can do about all the binary files?

Yeah not a huge fan of the GKW outputs. We could zip up the files so its just one? Not sure if that's easier or not

Pulling out the GKW files should be possible but may get a little fiddily

bpatel2107:
Will merge now and deal with issues as they arise!

PR #302, opened 2024-03-14
method to check quasi-neutrality on density gradients

arkabokshi:
`GENE` checks this condition is satisfied.  `GS2` does not. Not sure about `CGYRO`?

### DIFF
--- a/src/pyrokinetics/local_species.py
+++ b/src/pyrokinetics/local_species.py
@@ -134,7 +134,6 @@ class LocalSpecies(CleverDict):
 
         self.set_zeff()
         self.check_quasineutrality(tol=1e-3)
-        self.check_quasineutrality_gradient(tol=1e-3)
 
     def set_zeff(self) -> float:
         """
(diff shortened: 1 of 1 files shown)


bpatel2107:
Good spot here! I think we can just include the check on the gradients into the `check_quasineutrality` method that already exists, adds fewer lines.

Something like

```python
def check_quasineutrality(self, tol=1e-2):
        """
        Checks quasi-neutrality is satisfied and raises a warning if it is not
        """
        error = 0.0
        error_gradients = 0.0

        for name in self.names:
            species = self[name]
            error += species["dens"] * species["z"]
            error_gradients +=  species["dens"] * species["z"] * species["inverse_ln"]
           
        error = error / (self["electron"]["dens"] * self["electron"]["z"])
        error_gradients = error_gradients / (self["electron"]["dens"] * self["electron"]["z"] * species["inverse_ln")

        if abs(error) > tol or abs(error_gradients) > tol:
            warnings.warn(
                f"Currently local species violates quasi-neutrality in the density by {error.magnitude} and in the gradients by {error_gradients}"
            )
```

What you've down now is fine but this is more concise I think

bpatel2107 (approved):

arkabokshi:
@bpatel2107 agreed and combined. However using a different definition of the error to be consistent with how `GENE`, for example, would quantify the violation of quasi-neutrality.

bpatel2107:
In most scenarios I think this is fine, most of the time `electron.dens`*`electron.z` is usually -1 any way

PR #303, opened 2024-03-15
Update normalisation.py to allow for the conversion of original base …

bpatel2107:
…units

Initially when setting normalising reference values (`lref`, `nref` `tref` etc.), pyro would create a new unit (`lref_minor_radius_input0000`) which it would then be possible to convert to SI units (`meter`). However you could not convert the original base unit `lref_minor_radius`) to the new unit or SI which led to failures when setting the unit and then converting say the output to a different normalisation. This fixes that (had a feeling there was a issue like this for a while but finally figured it out this morning... very sneaky bug).

If you load directly from `Equilibrium`/`Kinetics` then this is not an issue as the base units are set from the get go. This only affects scenarios where the reference values are set later on down the line.

Added tests to check for this.

### DIFF
--- a/src/pyrokinetics/normalisation.py
+++ b/src/pyrokinetics/normalisation.py
@@ -430,7 +430,7 @@ class SimulationNormalisation(Normalisation):
         bref_B0_sim_unit = getattr(self.units, bref_B0_sim)
         self.context.add_transformation(
             "[bref]",
-            bref_B0_sim,
+            bref_B0_sim_unit.dimensionality,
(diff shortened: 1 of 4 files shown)


bpatel2107:
For some reason the density comparison didn't work even though the  values are the same when you manually check it. Tried a few different value for density but no luck. Still work overall so I'm a little confused but its fine.

ZedThree (approved):
LGTM, thanks @bpatel2107 !

PR #305, opened 2024-03-19
Enhancements/add gkw

arkabokshi:


src/pyrokinetics/gk_code/gkw.py
```diff
@@ -0,0 +1,418 @@
...
+
+from ..constants import deuterium_mass, electron_mass, pi, sqrt2
+from ..local_geometry import LocalGeometry, LocalGeometryMiller, default_miller_inputs
+from ..local_species import LocalSpecies
+from ..normalisation import SimulationNormalisation as Normalisation
+from ..normalisation import convert_dict, ureg
+from ..numerics import Numerics
+from ..templates import gk_templates
+from ..typing import PathLike
+from .GKInput import GKInput
```
bpatel2107:
```suggestion
from .gk_input import GKInput
```

This should help with the test

src/pyrokinetics/gk_code/gkw.py
```diff
@@ -0,0 +1,418 @@
...
+from ..local_species import LocalSpecies
+from ..normalisation import SimulationNormalisation as Normalisation
+from ..normalisation import convert_dict, ureg
+from ..numerics import Numerics
+from ..templates import gk_templates
+from ..typing import PathLike
+from .gk_input import GKInput
+
+
+class GKInputGKW(GKInput):
```
bpatel2107: See lines here to register this correctly.

https://github.com/pyro-kinetics/pyrokinetics/blob/4278eb9f2de1223d9d63dbad706ad09d79029979/src/pyrokinetics/gk_code/gs2.py#L30

This is needed I think

```suggestion
class GKInputGKW(GKInput, FileReader, file_type="GKW", reads=GKInput):
```

arkabokshi:
Thanks for the fixes @bpatel2107!

bpatel2107:
So far so good! Once we add the outputs then this should be good to merge

src/pyrokinetics/templates/input.gkw
```diff
@@ -0,0 +1,134 @@
...
+ lorentz = .false.
+ selfcollcon = .true.
+ freq_override = .true.
+ coll_freq =  0.00350139972
+ freq_input = .false.
+/
+
+&PYROKINETICS_INFO
+ major_r = 3.0 
+ minor_r = 1.0
```
bpatel2107, line 133: This should be 0.5 to match `eps` in `GEOM`

bpatel2107:
Here it looks like you've had some issues with the reference length scale and had to add in some additional terms. We could likely side-step this entire issue by giving everything units and so far the only thing missing units is `LocalGeometry`. Beyond that the only terms that actually need units are `local_geometry.Rmaj` `local_geometry.rho` and `local_geometry.Z0`. 

If we gave these units of `ureg.lref_major_radius` then we wouldn't really need to bother with any of these issues and the only time there would be a problem is when try to convert to a different code that doesn't accept that (though technically all of the are fine with it). Might be worth doing that to make everything cleaner

bpatel2107:
Merging this PR into #301 as that needs GKW for testing!

PR #306, opened 2024-03-22
Modified the kperp diagnostic to work on any nperiod 

whornsby82:
(other than 0 and 1, exception handing?)


bpatel2107:
Good spot. I think we can parameterise the test so we scan over nperiod as an input and make sure k_perp matches within the region that the GS2 output has

bpatel2107:
Added in the parameterised test and it looks good to go!

PR #310, opened 2024-04-04
Enhancements/clean documentation

arkabokshi:
A bugfix and removing some legacy code to make the documentation more accessible.  This will _always_  remain a WIP, so can be merged at any appropriate point!


arkabokshi:
@bpatel2107 Should we merge this into `enhancements/imas_v3.4` instead?  I can build the `GKW` documentation on top of this.

bpatel2107:
No this is already becoming a massive PR which isn't ideal so separate is fine I think. Just make the GKW relevant docs changes on that

Issue #308, opened 2024-04-02
When querying parameters, they should always provide a unit, even if no unit is present


whornsby82:
When querying parameters, the return is sometimes and xarray Quantity (e.g local_species['ion1'].inverse_lt gives, 5.166073892153321 / lref_minor_radius) and sometimes just a number (e.g q, s_hat).  This should be consolidated to a single type of return to make automation cleaner

bpatel2107:
This is done now so closing. Just a note, when setting a variable if a unit is not specified then it will maintain the same unit.

PR #311, opened 2024-04-05
Enhancements/auto detect normalisations

bpatel2107:
This PR sets up an automatic determination of the `NormalisationConvention` i.e., which values are used as the normalising quantities (`tref = T_electron` or `tref = T_deuterium` etc.). It does this by search through all the species in an input file, looks for which values are 1 and then tries to determine what those species are. 

1. First it looks at the mass of the negative charged species to figure out `mref` and also which species is the electron. The main options are `electron` `hydrogen` and `deuterium`. Currently we don't support anything else but could if there is a case for it.
2. Next it looks at which species have a density/temperature of 1.0. If the electron value is 1, then it sets it to the electron, otherwise if `mref` species happens to be 1 then it will use that. Can't see many cases where it won't be either the electron or the main ion. `tref` can then also change `vref` accordingly. _TODO: Still need to add impact on `rhoref`_
3. Next we look at `lref` and `bref` which in some cases are fixed but can be changed in some codes like GS2. `beta_ref` is also changed as needed.
4. If the determined values are different to the default values then pyro will create a new `Convention` where these values are used as the normalising units (GENE is a special case as `minor_r = 1` corresponds to the pyrokinetics `Convention` so we just force that). We also define how to map back to the standard pyrokinetics `Convention`.

This would allow for any convention to be used and the units automatically converted. This also applies to the outputs as well. This also is needed for GKW which only supports `lref = major_radius` which we don't really handle.

I've added some unit testing for a ~GS2~ all GK code cases from a mock input.

I've fixed a sneaky little issue where you could not redefine a Unit if it had already been defined once before using the hidden method `ureg._build_cache`. This likely could simplify the conversion of Simulation to Physical Units without the need for a Context but I can leave that to later. I've added a test anyway for it.

`LocalGeometry` and `Numerics` both now have units (I think basically everything now has `Units`!). `Numerics` is definitely tidier as I copied some useful methods from the  `GKOutputArgs` as it was a `dataclass` object. I'm wondering if `LocalGeometry` and `LocalSpecies` could be built in a similar way though they are more complicated that `Numerics` or any of the `GKOutputArgs`

I've also upgraded the version of `pint` to 0.23. Currently there are some small issues with `pint`  not supporting `np.roll` or `np.outer` which we used a few times in `LocalGeometry` so I made a PR on `pint` adding `roll` and will make one for `outer` eventually. Hopefully these can be added in 0.24 when released. The major issue here is that this version does not support python 3.8. We could do a pinning thing where we change the version depending on which python we use?

The method `_get_normalisation` (the old version of this which got the reference values is now named `get_references`) does the check for each code independently. Ideally we should use a modified method which returns a simple dictionary of all the relevant values and then a base class method which actually determines what everything should be so we use the same code again. Not a massive issue though.

Things to be added:

- [x] For this to be fully functional we need units in the `LocalGeometry` class (`Numerics` would be useful too`). This will help address #308 as everything probably should have a unit, even if dimensionless. Also needed for full GKW implementation #305 
- [x] Unit testing for other codes


src/pyrokinetics/diagnostics/__init__.py
```diff
@@ -109,14 +109,14 @@ def poincare(
         theta_metric = np.linspace(0, 2 * np.pi, 256)
         self.pyro.load_metric_terms(theta=theta_metric)
         nskip = len(geo.theta) // ntheta
-        bmag = np.sqrt((1 / geo.R) ** 2 + geo.b_poloidal**2)
+        bmag = np.sqrt((1 / geo.R.m) ** 2 + geo.b_poloidal.m**2)
```
ZedThree, line 112: This makes me a little concerned, but I don't understand the full context here. Are the quantities here all dimensionless?

src/pyrokinetics/gk_code/cgyro.py
```diff
@@ -322,7 +345,7 @@ def get_local_geometry_mxh(self) -> LocalGeometryMXH:
         ne_norm, Te_norm = self.get_ne_te_normalisation()
         beta = self.data.get("BETAE_UNIT", 0.0) * ne_norm * Te_norm
         if beta != 0:
-            mxh.B0 = 1 / (mxh.bunit_over_b0 * beta**0.5)
+            mxh.B0 = 1 / beta**0.5
         else:
             mxh.B0 = None
```
ZedThree: For a separate PR, but smells like we could use a helper function to cut down on some duplication here?

src/pyrokinetics/gk_code/cgyro.py
```diff
@@ -493,33 +527,147 @@ def get_numerics(self) -> Numerics:
...
+                "Pyro currently requires an electron species in the input file"
+            )
+
+        if len(temp_index) == 0 or len(dens_index) == 0:
+            raise ValueError("Cannot find any reference temperature/density species")
+
+        if not found_electron:
+            raise TypeError(
+                "Pyro currently only supports electron species with charge = -1"
+            )
```
ZedThree: The condition is the same as at L605 -- copy-paste error?

Also, probably more like `ValueError` than `TypeError`?

src/pyrokinetics/gk_code/cgyro.py
```diff
@@ -493,33 +527,147 @@ def get_numerics(self) -> Numerics:
...
+                    electron_index = i_sp + 1
+                    found_electron = True
+
+                if np.isclose(self.data[f"DENS_{i_sp+1}"], 1.0):
+                    dens_index.append(i_sp + 1)
+                if np.isclose(self.data[f"TEMP_{i_sp+1}"], 1.0):
+                    temp_index.append(i_sp + 1)
+
+        if not found_electron:
+            raise TypeError(
```
ZedThree:
```suggestion
            raise ValueError(
```

src/pyrokinetics/gk_code/cgyro.py
```diff
@@ -493,33 +527,147 @@ def get_numerics(self) -> Numerics:
...
+        me_mh = (electron_mass / hydrogen_mass).m
+
+        if np.isclose(e_mass, 1.0):
+            references["mref_species"] = "electron"
+        elif np.isclose(e_mass, me_md, rtol=0.1):
+            references["mref_species"] = "deuterium"
+        elif np.isclose(e_mass, me_mh, rtol=0.1):
+            references["mref_species"] = "hydrogen"
+        else:
+            raise ValueError("Cannot determine reference mass")
```
ZedThree: Could expand this error message and say that only electron/deuterium/hydrogen are currently supported? (Also, should we think about supporting tritium?)

src/pyrokinetics/gk_code/cgyro.py
```diff
@@ -493,33 +527,147 @@ def get_numerics(self) -> Numerics:
...
+
+        if electron_index in dens_index:
+            references["nref_species"] = "electron"
+        else:
+            for i_sp in dens_index:
+                if np.isclose(self.data[f"MASS_{i_sp}"], 1.0):
+                    references["nref_species"] = references["mref_species"]
+
+        if references["nref_species"] is None:
+            raise ValueError("Cannot determine reference density species")
```
ZedThree: Potential improvement for more user-friendly error message, could add something like "no species masses were 1.0 (found MASS_0 = 1.5, MASS_1 = 2.7, ...)"? But obviously, you know, with better English than that

I like nice, friendly, explicit error messages like that, as they make fixing bugs much easier!

src/pyrokinetics/gk_code/cgyro.py
```diff
@@ -493,33 +527,147 @@ def get_numerics(self) -> Numerics:
...
+            for i_sp in temp_index:
+                if np.isclose(self.data[f"TEMP_{i_sp}"], 1.0):
+                    references["tref_species"] = references["mref_species"]
+
+        if references["nref_species"] is None:
+            raise ValueError("Cannot determine reference density species")
+
+        rmaj = self.data["RMAJ"]
+
+        if rmaj == 1:
```
ZedThree: Does this need to use `np.isclose`, or is this likely to be exactly 1?

src/pyrokinetics/gk_code/cgyro.py
```diff
@@ -493,33 +527,147 @@ def get_numerics(self) -> Numerics:
...
+            references["lref"] = "major_radius"
+        else:
+            references["lref"] = "minor_radius"
+
+        if references == default_references:
+            return {}
+        else:
+            self.norm_convention = f"{self.code_name.lower()}_bespoke"
+            return references
```
ZedThree: Why return an empty dict if it matches the default references, rather than always returning the actual references used? It feels like that would give more consistent behaviour, but maybe I'm missing something!

src/pyrokinetics/gk_code/gene.py
```diff
@@ -140,7 +142,13 @@ def verify_file_type(self, filename: PathLike):
         expected_keys = ["general", "geometry", "box"]
         self.verify_expected_keys(filename, expected_keys)
 
-    def write(self, filename: PathLike, float_format: str = "", local_norm=None):
+    def write(
+        self,
+        filename: PathLike,
+        float_format: str = "",
+        local_norm=None,
+        code_normalisation=None,
```
ZedThree: Would be nice to add type hints here if possible!

src/pyrokinetics/gk_code/gene.py
```diff
@@ -178,27 +191,51 @@ def get_local_geometry(self) -> LocalGeometry:
...
+        minor_r = self.data["geometry"].get("minor_r", 0.0)
+        major_R = self.data["geometry"].get("major_r", 1.0)
+
+        if minor_r == 1.0:
+            self.norm_convention = "pyrokinetics"
+        elif major_R == 1.0:
+            self.norm_convention = "gene"
+        else:
+            raise ValueError(
+                "Pyrokinetics can only handle GENE simulations with either minor_r=1.0 or major_R = 1.0"
```
ZedThree: I really like error messages that give you as much info as possible:
```suggestion
                f"Pyrokinetics can only handle GENE simulations with either minor_r=1.0 (got {minor_r}) or major_R = 1.0 (got {major_R})"
```

src/pyrokinetics/gk_code/gs2.py
```diff
@@ -253,27 +280,25 @@ def get_local_species(self):
...
-            species_data.z *= ureg.elementary_charge
-            species_data.inverse_lt *= ureg.lref_minor_radius**-1
-            species_data.inverse_ln *= ureg.lref_minor_radius**-1
+            species_data.dens *= convention.nref
+            species_data.mass *= convention.mref
+            species_data.nu *= convention.vref / convention.lref
+            species_data.temp *= convention.tref
+            species_data.z *= convention.qref
+            species_data.inverse_lt *= convention.lref**-1
+            species_data.inverse_ln *= convention.lref**-1
```
ZedThree, line 284: I really really like this change, this feels very clean and "obvious" -- just use the convention to add the correct units!

src/pyrokinetics/gk_code/cgyro.py
```diff
@@ -493,33 +527,147 @@ def get_numerics(self) -> Numerics:
...
-    def get_normalisation(self, local_norm: Normalisation) -> Dict[str, Any]:
+    def get_reference_values(self, local_norm: Normalisation) -> Dict[str, Any]:
         """
-        Reads in normalisation values from input file
+        Reads in reference values from input file
 
         """
         return {}
 
+    def _get_normalisation(self):
```
ZedThree: As this also sets `self.norm_convention`, should this perhaps be named something more like `_detect_normalisation`? Definitely should mention it modifies `self` in the docstring though

src/pyrokinetics/local_geometry/local_geometry.py
```diff
@@ -299,6 +308,79 @@ def from_gk_data(cls, params: Dict[str, Any]):
...
+            attribute = getattr(self, key)
+
+            if hasattr(attribute, "units"):
+                new_attr = attribute.to(val)
+            elif attribute is not None:
+                new_attr = attribute * val
+
+            setattr(self, key, new_attr)
+
+    def _generate_local_geometry_units(self, norms):
```
ZedThree, line 358: Should mention this changes `self.unit_mapping` in the docstring.

Could also bike-shed the name, but it's fine

src/pyrokinetics/local_geometry/local_geometry.py
```diff
@@ -429,14 +519,25 @@ def get_bunit_over_b0(self):
 
         theta = np.linspace(0, 2 * pi, 256)
 
-        R, Z = self.get_flux_surface(theta=theta, normalised=True)
+        R, Z = self.get_flux_surface(theta=theta)
+
+        # Roll doesn't work on pint quantities...
```
ZedThree: Could add `TODO` with pint version number, so this could be found and fixed at a later date?

src/pyrokinetics/normalisation.py
```diff
@@ -234,13 +236,31 @@ def __init__(
...
+            # Construct name of beta_ref dimension
+            bref_type = str(bref).split("_")[1]
+            beta_ref_name = f"beta_ref_{nref_species[0]}{tref_species[0]}_{bref_type}"
+            self.beta_ref = getattr(ureg, beta_ref_name)
+        else:
+            self.beta_ref = betaref
 
         self.qref = ureg.elementary_charge
 
+    def __repr__(self):
```
ZedThree, line 253: Nice, love a good `__repr__`!

src/pyrokinetics/normalisation.py
```diff
@@ -364,6 +384,73 @@ def default_convention(self, convention):
...
+
+        if ne != 1.0:
+            self.units.define(
+                f"nref_{convention_dict['nref_species']} = {te ** -1} nref_electron",
+                units=True,
+            )
+
+        del convention_dict["rgeo_rmaj"]
+        del convention_dict["te"]
+        del convention_dict["ne"]
```
ZedThree: Should probably make a copy rather than modify the input argument

src/pyrokinetics/normalisation.py
```diff
@@ -364,6 +384,73 @@ def default_convention(self, convention):
         self._current_convention = self._conventions[convention]
         self._update_references()
 
+    def add_convention_normalisation(self, name=None, convention_dict=None):
```
ZedThree, line 391: If you want to go even further, could add a `TypedDict` to explicitly document what keys `convention_dict` should have. Possibly more effort than it's worth though!

src/pyrokinetics/normalisation.py
```diff
@@ -534,24 +642,35 @@ def set_ref_ratios(
...
-                f"lref_major_radius = {local_geometry.Rmaj} lref_minor_radius"
-            )
+            try:
+                aspect_ratio = local_geometry.Rmaj.to(self.pyrokinetics.lref).m
+                self.define(
+                    f"lref_major_radius = {aspect_ratio} lref_minor_radius",
+                    context=True,
+                )
+            except (PyroNormalisationError, pint.DimensionalityError):
+                warnings.warn(
```
ZedThree: Should this be a warning or an error?

src/pyrokinetics/numerics.py
```diff
@@ -163,3 +203,51 @@ def from_json(
...
             numerics_dict.pop("_metadata")
         return Numerics(**numerics_dict, title=overwrite_title)
+
+    def items(self) -> Generator[Tuple[str, Any], None, None]:
+        """Dict-like items iteration"""
+        return zip(iter(self), self.values())
+
+    def units(self, name: str, c: ConventionNormalisation) -> pint.Unit:
+        if name not in self.names:
+            raise ValueError(f"The coord '{name}' is not recognised")
```
ZedThree:
```suggestion
            raise ValueError(f"The coord '{name}' is not recognised (expected one of {Self.names}")
```

src/pyrokinetics/pyro.py
```diff
@@ -1695,7 +1723,10 @@ def load_local_species(self, psi_n: float, a_minor: Optional[float] = None) -> N
             )
 
         if a_minor is not None:
-            self.norms.set_lref(minor_radius=a_minor)
+            if isinstance(a_minor, PyroQuantity):
+                self.norms.set_lref(minor_radius=a_minor)
+            else:
+                raise ValueError("a_minor must be specified with units")
```
ZedThree:
```suggestion
            if not isinstance(a_minor, PyroQuantity):
                raise ValueError("a_minor must be specified with units")
            self.norms.set_lref(minor_radius=a_minor)          
```

Ok, I get this is super minor, but I think it's a good habit to write checks this way round!

tests/local_geometry/test_mxh.py
```diff
@@ -215,47 +215,54 @@ def test_grad_r(generate_miller, parameters, expected):
...
-        "shat": 0.7706147138551124,
-        "kappa": 3.0302699173285554,
-        "shift": -0.5766834602024067,
-        "s_kappa": -0.20110564435448555,
-        "dZ0dr": 9.223273885642161e-05,
-        "sn": [0.0, 0.45903873, -0.06941584, 0.00112094],
-        "cn": [-1.07040432e-04, 6.73097121e-05, 7.55332714e-07, 8.19418442e-06],
-        "dsndr": [0.0, 0.32807204, -0.02038408, -0.02555297],
-        "dcndr": [2.32569249e-04, -2.70991934e-04, 3.30192292e-05, 4.42607392e-05],
+        "B0": 2.197104321877944 * units.tesla,
```
ZedThree, line 229: For another PR, we should maybe look at trying to reduce the duplication across the miller tests

ZedThree (commented):
Looks really good @bpatel2107! It's great that basically everything has units now, and I really like the automatic determination of the convention from input files, big win there.


> The major issue here is that this version does not support python 3.8.

We should definitely consider dropping support for Python 3.8. Strictly, if we were to follow [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table) we should only support 3.10+!

This is a big PR so I have many comments, sorry! Some of them are probably even useful

src/pyrokinetics/gk_code/cgyro.py
bpatel2107: Agreed this could be done automatically now

tests/local_geometry/test_mxh.py
bpatel2107: Yes a lot are similar!

src/pyrokinetics/gk_code/cgyro.py
bpatel2107: Yes can delete
```suggestion
```

src/pyrokinetics/normalisation.py
bpatel2107: Error for sure

src/pyrokinetics/local_geometry/local_geometry.py
bpatel2107: 1 TODO for each roll or just a single TODO somewhere. Same is true for `np.outer`

src/pyrokinetics/local_geometry/local_geometry.py
bpatel2107: I don't know what bike-shedding is...

src/pyrokinetics/gk_code/cgyro.py
bpatel2107: Could add tritium!

src/pyrokinetics/gk_code/cgyro.py
bpatel2107: Could be done but I use this to check if we need to add in another Convention. If it exists then I create one, and if not I didn't want to recreate the same convention.

I could probably change this to None

src/pyrokinetics/local_geometry/local_geometry.py
ZedThree: Something like "getting bogged down in the details" :) See e.g. https://alexkondov.com/what-is-bikeshedding/

The name is fine, just not how I'd personally name it, so not worth spending much time worrying about!

src/pyrokinetics/gk_code/cgyro.py
bpatel2107: Close is fine

src/pyrokinetics/gk_code/cgyro.py
bpatel2107: Done

src/pyrokinetics/gk_code/cgyro.py
bpatel2107: Done

src/pyrokinetics/diagnostics/__init__.py
bpatel2107: Basically sort of. 

`Bmag = sqrt( Btor**2 + Bpol**2)`

Here we have `Btor = Bref * Rmaj / R`

Before we had `R` in units of `a_minor` so what was originally there was `Btor = a_minor / R` not `Rmaj`

`Bpol` is already in units of `Bref` so overall maybe it should be

`Bmag  = sqrt ((Bref * Rmaj/R)**2 + Bpol**2)`

@mgiacomin thoughts?

bpatel2107:
@ZedThree Made suggestion as requested;

bpatel2107:
ba

bpatel2107:
@ZedThree Made suggestion as requested. Move the actual logic of finding the normalisation convention into the base GKInput class which seems safer and avoids repeated calculations.

Things for a future PR:
- Remove duplication in LocalGeometry testing
- Helper function setting `B0` from `beta`
- Explicit `TypedDict` when making a convention
- Better handling of units in poincare plots

bpatel2107:
Was missing `rhoref` but added in now

ZedThree (approved):
LGTM! I think only thing missing is something in the docs or tutorials about this?

bpatel2107:
Agreed but can put that as a separate PR as docs aren't necessary for releases

bpatel2107:
Added some final tests when actually creating the `Convention` and ensuring the mappings are correct. I have also made it such that when you set any value in `Numerics`, `LocalSpecies` or `LocalGeometry`, if no unit is specified then it preserves the same unit as it had before.

This has also higlighted something where the definition of `ky` in CGYRO/TGLF $k_y = \frac{n q}{r}$ is different to the other codes GS2/GENE $k_y = \frac{n}{a}\frac{\partial \rho}{\partial \psi}$. We sort of swept this under the rug before but now that it everything has units we need to map the CGYRO $k_y$ to the actual pyro $k_y$ which is easy as they are different by a factor of $\frac{B_{unit}}{B_0}$. 

The only annoyance is that when you set `ky` with CGYRO, it doesn't correspond to the actual `ky` in the simulation in units of `rhoref_unit`. Though it does happen to exactly equal when `rhoref_pyro` is used. So doing `pyro.numerics.ky = x pyro.norms.units.rhoref_pyro` will put `x` in your input file.

In any case I think I am finally finished with this.

bpatel2107:
I'm going to merge this now and work out any teething issues before next release. Other PRs could use this now!

PR #314, opened 2024-04-16
Add tutorial for using Pyrokinetics on HPC systems

LiamPattinson:
A while back, I recall talking about a potential feature that could automatically dispatch gyrokinetics jobs to a remote HPC system (although I don't think an issue was raised for this). Having looked into it a bit, I think it would be very difficult to design a generic tool for doing this, so instead I've written a short guide on how it could be achieved for a simple toy problem. It first explains how to install Pyrokinetics on an HPC machine, and then gives a very rough overview of two tools that can be used to automate Pyrokinetics runs:

- [QCG-PilotJob](https://qcg-pilotjob.readthedocs.io/en/develop/): A job scheduler that runs within a single Slurm job. Allows you to queue up a large number of jobs and manages resource allocation for you.
- [HPC Rocket](https://svenmarcus.github.io/hpc-rocket/): A tool for dispatching jobs to remote HPC machines without the need to SSH in. Mostly intended as a CI tool, but it can used as a regular command line tool too.

Hopefully this might be useful to somebody! If not, I've at least had some fun playing around with new tools.

### DIFF
--- a/docs/tutorials/using_pyrokinetics_on_hpc_systems.rst
+++ b/docs/tutorials/using_pyrokinetics_on_hpc_systems.rst
@@ -399,8 +399,34 @@ We can set up a remote job run by creating a new directory, adding the scripts i
    continue_if_job_fails: true
 
 In order for this to work, we'll need to export our login password to the environment
-variable ``PASSWORD`` before running. There are also options to automate the login
-procedure using SSH keys.
+variable ``PASSWORD`` before running. If you have SSH keys set up on your chosen
+machine, it would be preferable to instead pass the location of your private key file:
+
+.. code-block:: yaml
+
+   # Replaces password line:
+   private_keyfile: ~/.ssh/my_private_key_file
+
+If your machine also requires multi-factor authentication, you'll need to add the
+following to the top of the file ``~/.ssh/config``:
+
+.. code-block::
+
+   ControlMaster auto
+   ControlPath ~/.ssh/connection-%h_%p_%r
+   ControlPersist 4h
+
+Before submitting any jobs, you'll also need to log in to create a 'master session':
+
+.. code-block:: bash
+
+   ssh -o ServerAliveInterval=30 -fN username@hpc.machine.com
+
+Including ``-fN`` will result in your session sitting idle in the background, while
+``-o ServerAliveInternal=30`` will ping the server every 30 seconds to ensure your
+connection stays alive. If you aren't using multi-factor authentication, it should be
+sufficient to just provide a password or private key file, and there's no need to create
+a master session beforehand.
 
 With our YAML file set up, we can then dispatch a new remote job using:
 
--- a/docs/tutorials/using_pyrokinetics_on_hpc_systems.rst
+++ b/docs/tutorials/using_pyrokinetics_on_hpc_systems.rst
@@ -407,6 +407,13 @@ machine, it would be preferable to instead pass the location of your private key
    # Replaces password line:
    private_keyfile: ~/.ssh/my_private_key_file
 
+If your private key requires a password to unlock, before submitting jobs you'll first
+need to add it to your keychain using:
+
+.. code-block:: bash
+
+   $ ssh-add -k ~/.ssh/my_private_key_file
+
 If your machine also requires multi-factor authentication, you'll need to add the
 following to the top of the file ``~/.ssh/config``:
 
@@ -420,7 +427,7 @@ Before submitting any jobs, you'll also need to log in to create a 'master sessi
 
 .. code-block:: bash
 
-   ssh -o ServerAliveInterval=30 -fN username@hpc.machine.com
+   $ ssh -o ServerAliveInterval=30 -fN username@hpc.machine.com
 
 Including ``-fN`` will result in your session sitting idle in the background, while
 ``-o ServerAliveInternal=30`` will ping the server every 30 seconds to ensure your
--- a/docs/tutorials/using_pyrokinetics_on_hpc_systems.rst
+++ b/docs/tutorials/using_pyrokinetics_on_hpc_systems.rst
@@ -74,7 +74,7 @@ our equilibrium files and a GS2 template file:
 
 .. code-block:: bash
 
-   $ scp pyrokinetics_template.in *.geqdsk username@hpc.com:scratch/pyrokinetics/
+   $ scp pyrokinetics_template.in *.geqdsk username@hpc.machine.com:scratch/pyrokinetics/
 
 
 The following script at ``~/scratch/pyrokinetics/pyrokinetics.sh`` will be used to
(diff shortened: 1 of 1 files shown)


docs/tutorials/using_pyrokinetics_on_hpc_systems.rst
```diff
@@ -0,0 +1,395 @@
...
+collect all commands into an ``sbatch`` script, and to submit this to the Slurm job
+scheduling system. Depending on your system, you may need to create this in your
+'scratch' or 'work' space -- the area on your file system where it's permitted to save
+large amounts of data, and where items may be periodically deleted to free up space.
+We'll work in the directory ``~/scratch/pyrokinetics/``, and we'll begin by copying over
+our equilibrium files and a GS2 template file:
+
+.. code-block:: bash
+
+   $ scp pyrokinetics_template.in *.geqdsk username@hpc.com:scratch/pyrokinetics/
```
d7919:
```suggestion
   $ scp pyrokinetics_template.in *.geqdsk username@hpc.machine.com:scratch/pyrokinetics/
```

docs/tutorials/using_pyrokinetics_on_hpc_systems.rst
```diff
@@ -0,0 +1,395 @@
...
+     -o tdotp_lowq0.geqdsk.d/tdotp_lowq0.in
+   pyro convert GS2 pyrokinetics_template.in --eq tdotp_highq0.geqdsk --psi 0.5 \
+     -o tdotp_highq0.geqdsk.d/tdotp_highq0.in
+   pyro convert GS2 pyrokinetics_template.in --eq tdotp_negtri.geqdsk --psi 0.5 \
+     -o tdotp_negtri.geqdsk.d/tdotp_negtri.in
+
+   # Perform runs
+   # ------------
+
+   srun $gs2 tdotp_lowq0.geqdsk.d/tdotp_lowq0.in
```
d7919, line 133: Note slurm has a (sysadmin configurable) maximum number of srun commands which can appear in a single job submission. The default is something like 40,000 so unlikely to hit this, but might be worth a mention somewhere?

docs/tutorials/using_pyrokinetics_on_hpc_systems.rst
```diff
@@ -0,0 +1,395 @@
...
+   $ squeue --me
+
+Assuming all goes well, this should generate three new GS2 input files -- one for each
+equilibrium file -- and then run each of them sequentially with 16 MPI processes each.
+The data will be available afterwards in the directories ``tdotp_*.geqdsk.d/``, and
+should be copied back from the system for analysis:
+
+.. code-block:: bash
+
+   $ scp username@hpc.com:scratch/pyrokinetics/tdotp_lowq0.geqdsk.d/tdotp_lowq0.out.nc .
```
d7919:
```suggestion
   $ scp username@hpc.machine.com:scratch/pyrokinetics/tdotp_lowq0.geqdsk.d/tdotp_lowq0.out.nc .
```

docs/tutorials/using_pyrokinetics_on_hpc_systems.rst
```diff
@@ -0,0 +1,395 @@
...
+   #!/bin/bash
+
+   # Slurm settings
+   # --------------
+
+   #SBATCH --job-name=pyro_job      # Job name
+   #SBATCH --mail-user=me@email.com # Where to send mail reporting on the job
+   #SBATCH --mail-type=END,FAIL     # Mail events (NONE, BEGIN, END, FAIL, ALL)
+   #SBATCH --nodes=1                # Number of compute nodes to run on
+   #SBATCH --ntasks-per-node=24     # Number of MPI processes to spawn per node
```
d7919, line 303: Difference wrt simple case -- might be worth keeping this the same as before to avoid "noise"

d7919: Ah I see from the python that you're asking each job to use 8 cores so I guess  the intention here is to run all three cases at the same time -- might be worth some text up front to indicate (e.g. before the python script) to say what approach to running you're targeting here?

docs/tutorials/using_pyrokinetics_on_hpc_systems.rst
```diff
@@ -0,0 +1,395 @@
...
+Before using this, note that we'll still need to SSH into the HPC machine in order to
+set up a Python environment and install all the software we'll need.
+
+We can set up a remote job run by creating a new directory, adding the scripts in
+:ref:`sec-advanced-hpc` and any equilibrium files we want, and then adding the file
+``pyro_job.yaml``:
+
+.. code-block:: yaml
+
+   host: hpc.com
```
d7919:
```suggestion
   host: hpc.machine.com
```

docs/tutorials/using_pyrokinetics_on_hpc_systems.rst
```diff
@@ -0,0 +1,395 @@
...
+   user: username
+   password: $PASSWORD
```
d7919, line 371: This feels a little dubious (and presumably won't work with systems requiring MFA?) -- might it be worth documenting how to set up a ControlMaster config (e.g. https://support.york.ac.uk/s/article/How-to-create-a-ControlMaster-SSH-session-so-that-you-only-need-to-authenticate-once-for-multiple-SSH-sessions) so that users just authenticate once and then hpc-rocket can reuse this?

LiamPattinson:
Thanks for the review! I've fixed up all of the inconsistencies, added some details on `srun` usage, and added some info on using SSH keys instead of supplying a password to HPC Rocket (I think a password might make more sense when you're running it as a CI job, as you can just save your password to a GitHub secret, but you're right that SSH keys are preferable in all other cases).

docs/tutorials/using_pyrokinetics_on_hpc_systems.rst
```diff
@@ -385,8 +399,34 @@ We can set up a remote job run by creating a new directory, adding the scripts i
...
+Before submitting any jobs, you'll also need to log in to create a 'master session':
+
+.. code-block:: bash
+
+   ssh -o ServerAliveInterval=30 -fN username@hpc.machine.com
+
+Including ``-fN`` will result in your session sitting idle in the background, while
+``-o ServerAliveInternal=30`` will ping the server every 30 seconds to ensure your
+connection stays alive. If you aren't using multi-factor authentication, it should be
+sufficient to just provide a password or private key file, and there's no need to create
```
d7919, line 435: Presumably one would need to unlock the private key file (e.g. with ssh-add) if it is not passwordless.

d7919 (approved):

LiamPattinson:
I've added a few lines about unlocking private key passwords beforehand. I'd actually been having issues with my keychain for a while, and this seems to have fixed it, so cheers for the prompt!

bpatel2107:
This looks fantastic! @whornsby82 @lorenzozanisi FYI may be of interest

Is it worth including info on how to pull the output data from the HPC cluster? I think that would close the whole loop though it is a bit out of the scope.

LiamPattinson:
There's a bit in the 'Simple usage' section about `scp`-ing output files back, but I guess I never explained how it works for HPC Rocket. I've added a small section to the end that explains how to ensure HPC Rocket pulls files back to you after the batch job has finished.

bpatel2107 (approved):

PR #315, opened 2024-04-18
update how ions are named

arkabokshi:
Naming all the ions "ion" breaks in `GENE3.0` (but works with `GENE2.0`).


bpatel2107:
Looks good. Depending on the order of the species in the input file you might get something like `ion0`, `ion2`, `ion3` etc. which is a little odd so maybe we include an ion_counter so it goes `ion1`, `ion2`, `ion3` etc.

arkabokshi:
Wondering if it is better to name the ions by their descriptive species name instead?

bpatel2107:
We probably could looking at the mass/charge but it's not always obvious. In #311 we determine the reference values so it becomes a little easier. My concern are edge cases where you have species with different charge states. Even more edge would be an isotope with a different mass.

I wonder if there is a python library function that can determine an element from its charge and mass?

bpatel2107:
Well here's something

```python

import numpy as np
from periodictable import elements

def determine_isotope(charge, mass):
    for element in elements:
        for isotope in element:
            if np.isclose(isotope.mass, mass, rtol=0.05) and np.isclose(isotope.number, charge, rtol=0.05):
                return f"{isotope.name}{int(isotope.mass)}"

    return "ion"

charge = 6
mass = 12

isotope = determine_isotope(charge, mass)
print(f"Isotope: {isotope}")
```

Which gets you `carbon12`

~I think it may fall down when you get isotopes of heavier elements that aren't fully ionised...~

It definitely falls down with heavier elements

```python
charge = 41
mass = 184

isotope = determine_isotope(charge, mass)
print(f"Isotope: {isotope}")

```

doesn't find tungsten41.

Maybe we could do a second loop if not fully ionised?

arkabokshi:
Thanks Bhavin.  I'll have a go at this later today.  Probably a method in local_species?

bpatel2107:
Throw it in the `normalise` method. That will always get called when reading in a file and have it as a kwarg flag to determine the name or not.

arkabokshi:
Sure! But then we'd also need the ion counter in case the name is not being determined?

bpatel2107 (approved):
Looks good to me. Can leave the fancy naming for another PR

Issue #255, opened 2023-09-20
LocalGeometry 'tracer_efit' not implemented for GENE


joeyschmidt97:
Hello,

I am trying to use pyrokinetics to read GENE files but am running into an error in that the geometry option 'tracer_efit' is not implemented for GENE code. Is there a way to suppress this error or a workaround? I really only need to read the output data (nrg, field, mom, etc.) files and not anything to do with the profiles.

Here's the code I am trying to run:
```
from pyrokinetics import Pyro

# Point to GENE input file
gene_file = 'filepath_to_simulation/scanfiles0000/parameters_0001'

# Load in file
pyro = Pyro(gk_file=gene_file, gk_code="GENE")

# Load in GENE output data
pyro.load_gk_output()
data = pyro.gk_output
```

Here's the error I am getting:
```
----> pyro = Pyro(gk_file=gene_file, gk_code="GENE")

File [~/.local/perlmutter/3.9-anaconda-2021.11/lib/python3.11/site-packages/pyrokinetics/pyro.py:867](https://vscode-remote+ssh-002dremote-002bperlmutter-002dp1-002enersc-002egov.vscode-resource.vscode-cdn.net/global/homes/fusion_research/~/.local/perlmutter/3.9-anaconda-2021.11/lib/python3.11/site-packages/pyrokinetics/pyro.py:867), in Pyro.read_gk_file(self, gk_file, gk_code, no_process)
    865 # Set LocalGeometry, LocalSpecies, Numerics, unless told not to.
    866 if "local_geometry" not in no_process:
--> 867     self.local_geometry = self.gk_input.get_local_geometry()
...
--> 189     raise NotImplementedError(
    190         f"LocalGeometry type {geometry_type} not implemented for GENE"
    191     )

NotImplementedError: LocalGeometry type tracer_efit not implemented for GENE
```

And in the geometry part of my GENE file. My initial thought is that the g-file (geomfile) is a relative path (loaded in wherever the GENE simulation directory is)

```
&geometry
magn_geometry = 'tracer_efit'
q0       =    7.1213618
shat     =    3.9870173
geomfile = 'g129015.00409_450'
edge_opt =    8.0000000
minor_r  =    1.0000000
major_R  =    1.0208728
rhostar  =   0.47368176E-02
dpdx_term= 'full_drift'
dpdx_pm  =   0.48974139E-01
norm_flux_projection  = F
/
```

Let me know if more information is required for this issue. Any help is appreciated!

[...]

PR #319, opened 2024-04-24
Load in tracer_efit as "circular" LocalGeometry with q and shat

bpatel2107:
Fixes issue #255

Not sure if this is the best solution but it does let you read in GENE inputs with geometry set to `tracer_efit` **assuming** $q$ and $\hat{s}$ are defined in the input file. If not then when you load in eigenfunctions the mapping from flux tube to ballooning space won't work, see 

https://github.com/pyro-kinetics/pyrokinetics/blob/enhancements/gene_tracer_efit/src/pyrokinetics/gk_code/gene.py#L1169-L1183

@dake0795 @joeyschmidt FYI

Ideally I'd want to have all the things in a `LocalGeometry` object (`R`, `Z`, plus derivatives w.r.t `theta` and `r`) but this is fine. Note you can't convert to CGYRO units as we don't know the value of `Bunit/B0` without those derivatives mentioned above.

EDIT: This should also work for `slab` and `s_alpha`. Note this only works with reading in data. Writing out new files will fail


bpatel2107:
GENE v3.0 input can have an issue with setting `FCVERSION` in the `parameters_XXXX` files and that can have brackets i.e. `FCVERSION = GNU Fortran (Spack GCC) 9.4.0\` which breaks f90nml in a bad way. Brackets are reserved for lists in fortran namelists so I don't think changing f90nml is going to happen.

Hacky fix is to convert the file to a str, read in, remove brackets and then use the `_read_str` method. Not ideal but a functional placeholder until fixed in GENE.

PR #325, opened 2024-04-30
Enhancements/add gkw v2

bpatel2107:
Separate PR for including GKW. Branched from #301 

Major change is of course including GKW and giving the option to choose the normalisation of the outputs. I dropped the number of time point for the template data but it is still 200 odd files...


src/pyrokinetics/gk_code/gkw.py
```diff
@@ -0,0 +1,1331 @@
...
+        me_mh = (electron_mass / hydrogen_mass).m
+
+        if np.isclose(e_mass, 1.0):
+            references["mref_species"] = "electron"
+        elif np.isclose(e_mass, me_md, rtol=0.1):
+            references["mref_species"] = "deuterium"
+        elif np.isclose(e_mass, me_mh, rtol=0.1):
+            references["mref_species"] = "hydrogen"
+        else:
+            raise ValueError("Cannot determine reference mass")
```
ZedThree: To-do for future PR: pull out as helper function?

bpatel2107: Added to match other codes and included in tests

ZedThree:
Are the ~170 output files only really needed for the tests (and possibly examples)? Some options for dealing with them:

1. zip/tar them, and extract them during the tests with a fixture. The downside is we then have an opaque binary file, but it seems that this is already the case for most of the files, but we would only have the one. It does also add a small layer of complexity over the tests and examples. We could potentially incorporate the ability to read directly from a zip file/tarball if necessary?
2. use [git LFS](https://git-lfs.com) to track them. This keeps them out of the main repo at least, but potentially makes things a bit more complicated -- I'm not very familiar with git LFS so I'm not sure what this is like in practice.
3. just deal with it. I guess my main objections to this really boil down to "it's untidy and annoying". Presumably these files won't change very often, so hopefully won't make very many PRs difficult to browse?
4. use fewer output timesteps? It looks like there are 26 files per output timestep, so only keeping 4 or 5 timesteps would make a big difference.

It also looks like GKW supports HDF5 output, but we'd still need to keep some binary files for the tests. Do we also need to add support for the HDF5 output?

bpatel2107:
We are currently only keeping 6 timesteps and I chose that as that was the minimum needed to get a meaningful `growth_rate_tolerance` on the output using the default averaging window (final 20% in time of the data).

We can of course zip them which make a single file but it is annoying to deal with but that only needs to be dealt with one off in the in testing. 

Adding in the ability to read HDf5 is probably useful but will require adding a decent amount to the `GKOutputReader` in `GKW` @arkabokshi is the HDf5 output a single file with all the data?

The best solution is to move all of the outputs to a completely separate repo `pyrokinetics_gk_outputs` (like we did with the plugin testing) so none of the outputs are there and only pull that repo for the testing.

arkabokshi:
I was advised that the `HDF5` format in GKW could be "buggy" and best to use as  setup currently. Maybe @ycamenen can comment on any latest developments?

arkabokshi (approved):

bpatel2107:
@ZedThree managed to zip up the outputs without too much issue so we are only adding 1 file now!

Issue #326, opened 2024-05-01
Tests taking a very long time


LiamPattinson:
Following PR https://github.com/pyro-kinetics/pyrokinetics/pull/311, tests are taking much longer to finish (about 20 minutes on the GitHub CI runners, vs 4 minutes before).

To figure out what's going on, I tried running `pytest --duration=20`, which tells you the time taken to run the slowest 20 tests. I was on commit 7576e7d0, but I get similar results running from the latest commit on `unstable`. The output was:

```
==== slowest 20 durations ====
159.46s call     tests/test_pyroscan.py::test_compare_read_write_pyroscan
82.67s call     tests/test_pyroscan.py::test_apply_func
50.91s call     tests/test_pyroscan.py::test_create_single_run
30.30s call     tests/test_pyroscan.py::test_format_run_name
25.20s call     tests/test_deepcopy.py::test_deepcopy_pyroscan[electron_temp_gradient-values2]
22.71s call     tests/test_deepcopy.py::test_deepcopy_pyroscan[ky-values0]
22.69s call     tests/test_deepcopy.py::test_deepcopy_pyroscan[kappa-values1]
13.88s setup    tests/equilibrium/test_gacode.py::test_compare_gacode_geqdsk_attrs[R_major]
13.71s call     tests/local_geometry/test_miller_turnbull.py::test_load_from_eq
13.19s call     tests/equilibrium/test_gacode.py::test_read
10.31s call     tests/test_deepcopy.py::test_deepcopy_pyro
8.27s call     tests/local_geometry/test_transp_cdf_vs_geqdsk.py::test_compare_transp_cdf_geqdsk
4.55s setup    tests/test_roundtrip.py::test_compare_roundtrip_exb[gs2-cgyro]
4.50s call     tests/cli/test_convert.py::test_convert_all_opts[True]
4.37s call     tests/cli/test_convert.py::test_convert_with_kinetics_and_eq[TRANSP]
4.29s call     tests/cli/test_convert.py::test_convert_all_opts[False]
4.05s call     tests/cli/test_convert.py::test_convert_with_kinetics_and_eq[JETTO]
3.83s call     tests/cli/test_convert.py::test_convert_with_kinetics_and_eq[SCENE]
3.58s call     tests/local_geometry/test_fourier_cgyro.py::test_load_from_eq
3.45s setup    tests/test_roundtrip.py::test_compare_roundtrip[gs2-cgyro]
```

[...]

PR #328, opened 2024-05-07
Bugfix: Slow deepcopies in tests

LiamPattinson:
Fixes Issue https://github.com/pyro-kinetics/pyrokinetics/issues/326

When deepcopying `SimulationNormalisation`, we call the line:

https://github.com/pyro-kinetics/pyrokinetics/blob/3f9f09d96ab890e78852d39d5adb2a395c8a4e62/src/pyrokinetics/normalisation.py#L357

This deep-copies the following dict:

https://github.com/pyro-kinetics/pyrokinetics/blob/3f9f09d96ab890e78852d39d5adb2a395c8a4e62/src/pyrokinetics/normalisation.py#L325-L328

`ConventionNormalisation` does not have a custom `__deepcopy__`, but it does contain the following line in its `__init__`:

https://github.com/pyro-kinetics/pyrokinetics/blob/3f9f09d96ab890e78852d39d5adb2a395c8a4e62/src/pyrokinetics/normalisation.py#L932

This means each deepcopy was duplicating the pint registry 6 times -- once for each `Convention` in `NORMALISATION_CONVENTIONS`. Somehow this was interacting with the routines that rebuild Pint's cache in functions such as `set_lref` (still not sure what's going on there) to cause the tests to run about 5 times slower than before.

This PR adds a custom `__deepcopy__` to `ConventionNormalisation` that avoids deepcopying the Pint registry, and cleans up the `__deepcopy__` in `SimulationNormalisation`.

### DIFF
--- a/src/pyrokinetics/normalisation.py
+++ b/src/pyrokinetics/normalisation.py
@@ -935,10 +935,7 @@ class ConventionNormalisation(Normalisation):
         self._update_system()
 
     def __deepcopy__(self, memodict):
-        """Overrides deepcopy behaviour to perform regular copy.
-
-        Performing a full deepcopy would duplicate the Pint registry itself, which
-        can lead to significant performance problems."""
+        """Overrides deepcopy behaviour to perform regular copy of the Pint registry."""
         new_obj = object.__new__(type(self))
         for k, v in self.__dict__.items():
             if k == "_registry" or k == "_system":
--- a/src/pyrokinetics/normalisation.py
+++ b/src/pyrokinetics/normalisation.py
@@ -354,23 +354,10 @@ class SimulationNormalisation(Normalisation):
         new_object = SimulationNormalisation("COPY")
         new_object.name = self.name
         new_object.units = self.units
-        new_object._conventions = copy.deepcopy(self._conventions)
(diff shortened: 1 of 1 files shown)


bpatel2107 (approved):
That seems to have cut the tests run time by over half! Seems to run in about 8 minutes now and before the update with automatic detection of Convention it was running in about ~4m30. Do we think it can go further down?

LiamPattinson:
I've merged in the file-caching stuff I did last week, and then tried rerunning the profiling. This time I used the `pytest-profiling` package, and got this neat `.svg` that breaks everything down:

combined

Around 50% of the runtime is being spent on `_build_cache`, and 20% on reading input files (hard to avoid that, though it might be possible to optimise individual file reads). We could avoid that if we had a system that always defines a new normalisation rather than modifying an existing one, but I think it'd be better to save that until another PR.

bpatel2107:
We could check in that unit exists and then create a new `SimulationNormalisation` if it does. I did something similar here https://github.com/pyro-kinetics/pyrokinetics/commit/3e181f97d9f90cecfc3f7e88b053c848db3dc97f which kind of worked. 

~Will raise an issue on the `build_cache` issue~ Will bring this up in the existing issue

In any case this has significantly sped things up!

Issue #330, opened 2024-05-09
pyro_to_ids fails when outside of pyrokinetics directory


jmcclena:
fails due to not finding git repo

branch: `enhancements/imas_update_v3.4`

error: 
```
File [~/miniconda3/lib/python3.11/site-packages/pyrokinetics/databases/imas.py:270](http://localhost:8888/lab/tree/~/miniconda3/lib/python3.11/site-packages/pyrokinetics/databases/imas.py#line=269), in pyro_to_imas_mapping(pyro, comment, time_interval, reference_values, ids)
    262 ids_properties = {
    263     "provider": "pyrokinetics",
    264     "creation_date": str(datetime.now()),
    265     "comment": comment,
    266     "homogeneous_time": 1,
    267 }
    269 ids_properties = gkids.IdsProperties(**ids_properties)
--> 270 repo = git.Repo(search_parent_directories=True)
    271 sha = repo.head.object.hexsha
    272 code_library = [
    273     {
    274         "name": "pyrokinetics",
   (...)
    278     }
    279 ]

File [~/miniconda3/lib/python3.11/site-packages/git/repo/base.py:276](http://localhost:8888/lab/tree/~/miniconda3/lib/python3.11/site-packages/git/repo/base.py#line=275), in Repo.__init__(self, path, odbt, search_parent_directories, expand_vars)
    273 # END while curpath
    275 if git_dir is None:
--> 276     raise InvalidGitRepositoryError(epath)
    277 self.git_dir = git_dir
    279 self._bare = False

InvalidGitRepositoryError: [/Users/mcclenaghan/programming/python/jupyter_notebooks](http://localhost:8888/lab/tree/mgkdb.ipynb)
```

From what I've read, I don't think there is a way to know the commit hash of an installed package. Would it make sense to wrap this logic in a try-except?

bpatel2107:
Yeah I saw this too at times I had a fix but that broke the testing see here

https://github.com/pyro-kinetics/pyrokinetics/actions/runs/8264030375/job/22606734500
https://github.com/pyro-kinetics/pyrokinetics/pull/301/commits/0bfa669930b34c78bde6f113fa49a6d5eb237454

I can add a try-except and we store the version of pyrokinetics anyway which _should_ be enough

LiamPattinson:
[...]

PR #331, opened 2024-05-09
Replace `setuptools-scm` with `versioningit`, store `__commit__` alongside `__version__`

LiamPattinson:
An attempt to solve Issue https://github.com/pyro-kinetics/pyrokinetics/issues/330

Currently uses `gitpython` to get the commit hash of the project at runtime, but this won't work if you're using a `pip` installed package or working outside the git repo.

This replaces `setuptools-scm` with the alternative [`versioningit`](https://versioningit.readthedocs.io/en/stable/). It's more configurable than `setuptools-scm`, and allows you to set a template for the `_version.py` file. I've configured it to mimic the `setuptools-scm` versioning scheme, and it will also store the full commit hash at build/install.

It may be possible to achieve the same outcome with `setuptools-scm`, but I wasn't able to figure it out from the docs.

### DIFF
--- a/.github/workflows/docker_publish.yml
+++ b/.github/workflows/docker_publish.yml
@@ -10,6 +10,11 @@ jobs:
   publish-docker:
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
 
@@ -29,6 +34,7 @@ jobs:
       - name: Build and push Docker image
         uses: docker/build-push-action@v5
         with:
+          context: .
           push: ${{ github.event_name == 'release' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/python_publish.yml
+++ b/.github/workflows/python_publish.yml
@@ -8,9 +8,11 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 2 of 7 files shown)


LiamPattinson:
It looks like it doesn't work on the GitHub test runners. I'll see if there's some way to configure them to get this working.

LiamPattinson:
Looks like I (almost) got it working, at the cost that the CI jobs now need to clone the full repo instead of just the latest commit. It'd be nice if we could avoid this though. @ZedThree Are you aware of another solution for this? I had a quick play around with [`versioneer`](https://pypi.org/project/versioneer/) but didn't get very far with it.

The latest issue seems to be that it can't find output files for testing for some reason... not sure what's going on there.

ZedThree:
It looks like it works similarly to `setuptools-scm`, in that it calculates number of commits since the last tag, which necessarily requires a non-shallow clone of the repo. So it's probably the case that we were setting a fallback version for `setuptools-scm`, and need to do something similar for `versioningit`. This is going to be the case for any tool we use.

LiamPattinson:
Good catch, I didn't realise I'd dropped that. Added it back now. 

Looking back through some old actions on other PRs, it looks like both the tests and docker build were installing with a fallback tag. While this is fine for the tests, I've left the full checkout for docker builds, as anybody using them would otherwise end up with incorrect metadata.

Still no idea why the tests are failing to find output files in the templates directory, and I've so far failed to reproduce the error locally.

LiamPattinson:
It turns out `setuptools-scm` also helps manage your package data alongside providing versioning tools, so I've added that back as a dependency and it looks like it fixed the problem.

bpatel2107 (approved):
Works for me when I re-install with pip

PR #332, opened 2024-05-09
Bugfix/eigenvalues units

bpatel2107:
How this slipped through the cracks I have no idea (though I do outline how below). The growth rate and frequency had inverted units 😱 

Essentially we normally calculate the eigenvalues directly from the fields and in doing so we dropped the units during the calculation (because `phi` `apar` and `bpar` have different units) and then apply the units back on afterwards. So the actual value of the growth rate and frequency was technically correct but if we were to convert the output to a different convention then it would convert the wrong way. But as long as you didn't want to convert the units to a different normalisation you would be fine.

However, this bug is particularly bad when you would not be able to get the eigenvalues from the fields so would read from the gk output directly. In that case it reads in the eigenvalues, applies the code's local convention and then tries to convert it to the pyro convention and that would fail to give you the correct value.

In #325 we created the option to specify the Convention of the outputs in Pyro for GKW so this becomes even more likely to be an issue and in #301 we actually need to convert from pyro to GKW convention.

The fix is straightforward and a test has been included to directly compare to the GS2 output omega. The tolerance is reasonably high due to the fact that this run has not converged but before they would be different by a factor ~4

### DIFF
--- a/tests/gk_code/test_gk_output_reader_gs2.py
+++ b/tests/gk_code/test_gk_output_reader_gs2.py
@@ -8,6 +8,7 @@ import xarray as xr
 import numpy as np
 import pytest
 from types import SimpleNamespace as basic_object
+import netCDF4 as nc
 
 
 @pytest.fixture(scope="module")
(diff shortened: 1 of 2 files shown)


ZedThree:
The diff is a bit confusing, it looks like it's adding GKW again? I think the only new commit is actually [89cf7df](https://github.com/pyro-kinetics/pyrokinetics/pull/332/commits/89cf7df84feaf63bfd3fda2a65f5ea3f4e41a412)?

bpatel2107:
Yeah I just merged `add_gkw_v2` so I thought it would update but yes that is the only difference. Maybe I can close and re open it.

tests/gk_code/test_gk_output_reader_gs2.py
```diff
@@ -83,10 +84,40 @@ def test_infer_path_from_input_file_gs2():
...
+    """Can we match growth rate/frequency from netCDF file"""
+
+    path = template_dir / "outputs" / "GS2_linear"
+    pyro = Pyro(gk_file=path / "gs2.in", name="test_gk_output_gs2")
+    pyro.load_gk_output()
+
+    netcdf_data = nc.Dataset(template_dir / "outputs" / "GS2_linear" / "gs2.out.nc")
+
+    cdf_mode_freq = netcdf_data["omega"][-1, 0, 0, 0]
+    cdf_gamma = netcdf_data["omega"][-1, 0, 0, 1]
```
ZedThree: Best practice is to use `with nc.Dataset`:

```suggestion
    with nc.Dataset(path / "gs2.out.nc") as netcdf_data:
        cdf_mode_freq = netcdf_data["omega"][-1, 0, 0, 0]
        cdf_gamma = netcdf_data["omega"][-1, 0, 0, 1]
```

(also reuses `path`)

ZedThree (approved):
LGTM. Maybe just a small tidy up of the test.

I'm a bit confused why this test isn't exact though? Is it not just doing a roundtrip of the data through our conversions? Or is there an actual calculation going on in Pyro?

I think it might also be worth either rebasing this branch on top of latest `unstable`, or merging `unstable` in here

bpatel2107:
The won't be the same because GS2 doesn't spit out data every time step, only every `nwrite` which is specified in the input file. If you spit it out every time step then it should be closer. This is less of an issue when you are converged as they will tend to agree very well.

Another smaller point is that the sum of the fields is done to calculate the eigenvalues here

https://github.com/pyro-kinetics/pyrokinetics/blob/unstable/src/pyrokinetics/gk_code/gk_output.py#L670-L680

The specific normalisation choice here can slightly impact the value you use. For example GS2 uses `|B|` when normalising `bpar` rather than a fixed `B` so the actual `bpar` in the simulations is slightly different. Maybe GS2 corrects for this when doing the sum but I don't know.

LiamPattinson:
> How this slipped through the cracks I have no idea

Maybe `git blame` can help us out here...

```
b5683402 pyrokinetics/gk_code/gk_output.py     (Liam Pattinson  2023-07-06 18:31:52 +0100 346)         return c.lref / c.vref
```

Ah, that explains it. Sorry guys!

PR #338, opened 2024-05-13
GK code stella input file conversion feature

mrhardman:
Initial attempt to make the input file converting feature of pyrokinetics work for the GK code stella.

Tests have been carried out for converting between GS2 and stella input files, but these need to be documented.

Making this PR to make sure that the work is not lost, and to request reviews and test suggestions from the core developers.

Nonlinear and linear example input files are provided in the /templates/ folder.

This version is written for the branch https://github.com/stellaGK/stella/tree/development/apar2plusbpar (stella currently requires significant merging between development branches, and this has B||, and so has the appropriate electromagnetic field input flags).


src/pyrokinetics/templates.py
```diff
@@ -7,11 +7,13 @@
 gk_cgyro_template = template_dir / "input.cgyro"
 gk_gene_template = template_dir / "input.gene"
 gk_tglf_template = template_dir / "input.tglf"
+gk_stella_template = template_dir / "input.stella"
 gk_templates = {
     "GS2": gk_gs2_template,
     "CGYRO": gk_cgyro_template,
     "GENE": gk_gene_template,
     "TGLF": gk_tglf_template,
+    "stella": gk_stella_template,
```
bpatel2107: To best consistent this should be uppercase

```suggestion
    "STELLA": gk_stella_template,
```

src/pyrokinetics/gk_code/__init__.py
```diff
@@ -6,6 +6,7 @@
 from .gk_input import GKInput, read_gk_input, supported_gk_input_types
 from .gk_output import GKOutput, read_gk_output, supported_gk_output_types
 from .gs2 import GKInputGS2, GKOutputReaderGS2  # noqa
+from .stella import GKInputSTELLA  # , GKOutputReaderSTELLA  # noqa
```
bpatel2107: Should we add the `OutputReader` too?

bpatel2107 (commented):
@mrhardman  thanks for this! There have been some changes to `Numerics` `LocalGeometry` (they now have units!) and the way the normalisation is detected so it may be faster if I tweak this to work with the latest version. But this looks great so far.

Does the output reader work?

src/pyrokinetics/gk_code/__init__.py
mrhardman: The output reader probably does not work at the moment. I will try to get a minimal version working this week.

src/pyrokinetics/templates.py
mrhardman: OK, fine for me. The stella name convention is to use lower case for the code name, but I am happy either way here so long as it's obvious (which it might be from your documentation, once I update it).

mrhardman:
> @mrhardman thanks for this! There have been some changes to `Numerics` `LocalGeometry` (they now have units!) and the way the normalisation is detected so it may be faster if I tweak this to work with the latest version. But this looks great so far.

Please do the merge from unstable in to the stella_new branch, if you feel you have time. I started but realised I was out of my depth.

> Does the output reader work?

Not yet but I can start on it this week.

src/pyrokinetics/templates.py
mrhardman: Hopefully the last commits address this concern @bpatel2107?

mrhardman:
@bpatel2107 The code can now read the output fluxes, so please review when you are ready.

bpatel2107 (approved):
Looks great, thanks @mrhardman, I've tweaked the tests so we are good to merge now! Thanks for you contribution!

mrhardman:
Before me merge, I might have spotted an issue:

When I try to convert the attached GS2 input to stella with

```
from pyrokinetics import Pyro

# Point to GS2 input file
gs2_template = "tear_miller_gs2.in"

# Load in GS2 file
pyro = Pyro(gk_file=gs2_template)

# Switch to STELLA
pyro.gk_code = "STELLA"

# Write STELLa input file
pyro.write_gk_file(file_name="tear_miller_stella.in")
```

I get a ky variable in stella that is `0.86...` when it should be `0.8`. What have I done wrong?

[pyrokinetics.zip](https://github.com/pyro-kinetics/pyrokinetics/files/15375617/pyrokinetics.zip)

bpatel2107:
This actually comes from the fact that you have `r_geo != rmaj` which changes the reference B field. But you have happened to have stumbled on a bug related to #344. As of right now `numerics.ky` has units of `[1 / rhoref]`, so when you convert from the GS2 file (which uses `Bgeo` not `B0`here ) to the STELLA file we convert from `rhoref_Bgeo` to `rhoref_B0`.

We are confident that the `aky` **should** actually be the same in this scenario but we are thinking of the solution as it is not clear due to the slightly messy definition where $k_y = \frac{n_0}{L_{ref}} \frac{\partial r_{N}}{\partial \psi_N}$ which changes when `Bref` changes. So `ky` doesn't actually depend on `Bref` due to the definition of `drho_dpsi` having a `Bref` which cancels it out. 

This may involve changing either the units of `ky` or the definition and I am not sure which is more important.

mrhardman:
I would guess that stella should also use a `Bgeo = I(psi)/Rgeo` normalisation for the magnetic field strength -- if that allows a quick fix I would take that option. Otherwise, if you think this is a symptom of a wider problem, happy for the merge to go ahead and I will report further issues separately.

bpatel2107:
It would not allow for a quick fix but there is a way to handle this in the meantime. I will merge and then make a branch with this hacky fix. This is definitely a symptom of a larger problem though

Issue #316, opened 2024-04-19
v0.6.2 release


bpatel2107:
I think we are well overdue for a new release! Wait for the following to be merged and then we can crack on with a release

- [ ] #301
- [ ] #311 
- [x] #315

LiamPattinson:
I agree, now looks like a good time! It would definitely be good to have a release with the `pygacode` alternative you wrote, as currently it can lead to `pip install pyrokinetics` failing on some systems.

LiamPattinson:
I just raised PR https://github.com/pyro-kinetics/pyrokinetics/pull/340 which updates the PyPI publishing workflow to the new standards. All three of the bullet points in the top post have been achieved, so is there anything else that we need before the next release? We're adding quite a lot and there are some changes to the user interface, so I'd argue this should be v0.7.0 instead of v0.6.2.

Should we also organise a development meeting sometime soon? It'd be good to catch up and discuss what our priorities should be beyond the next release.

LiamPattinson:
Cheers for merging PR https://github.com/pyro-kinetics/pyrokinetics/pull/345! Are there any other tasks that need doing before releasing v0.7.0?

bpatel2107:
Technically no but I am very close to making a big change on `GKOutput`. But that can probably be for v0.7.1 but when that is merged I'll want to do another release basically immediately

LiamPattinson:
No issue with doing another release after new features have been added!

I'm personally working on a fix to Issues https://github.com/pyro-kinetics/pyrokinetics/issues/326 and https://github.com/pyro-kinetics/pyrokinetics/issues/337, and I'm hoping to have a demonstration ready tomorrow or early next week. It includes a rewrite and merging of `SimulationNormalisation` and `ConventionNormalisation` into a single simplified class, and rethinks how units are assigned and updated. However, the full implementation would affect a large proportion of the project, and will need to be split over multiple issues and pull requests (it would probably be best to manage using GitHub Projects). The sum total of those updates would probably require a v0.8.0 release, but if I can compartmentalise each of the prerequisite steps well enough, it may also result in versions 0.7.2, 0.7.3, etc.

PR #340, opened 2024-05-15
Update publishing workflow

LiamPattinson:
In advance of our imminent next release (Issue https://github.com/pyro-kinetics/pyrokinetics/issues/316), I've updated the PyPI publishing workflow to meet the new standards. The previous method of storing PyPI usernames and passwords with GitHub secrets is now obsolete, and PyPI now prefers to use 'trusted publishers' -- [this page](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/) explains how to configure everything. I think @bpatel2107 or @ZedThree will have to configure things on the PyPI side.

Things changed in this update:
- Now builds a distribution on every push
- Calls `twine check` to ensure the packages are being built correctly
- Uploads built packages as an artefact which can be downloaded from the action
- Introduces a second job which publishes on new releases
- Publishing _will not work anymore_ unless trusted publishers are set up

### DIFF
--- a/.github/workflows/draft-pdf.yml
+++ b/.github/workflows/draft-pdf.yml
@@ -14,7 +14,7 @@ jobs:
           # This should be the path to the paper within your repo.
           paper-path: docs/joss/paper.md
       - name: Upload
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v4
         with:
           name: paper
           # This is the output path where Pandoc will write the compiled
--- a/.github/workflows/python_publish.yml
+++ b/.github/workflows/python_publish.yml
@@ -24,7 +24,7 @@ jobs:
         run: |
           python -m twine check dist/*
       - name: Store the distribution packages
-        uses: actions/upload-artifact@v3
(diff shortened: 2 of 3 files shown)


LiamPattinson:
While I'm at it, just updated the `upload/download-artifact` actions as the old versions are due to be deprecated within the year.

bpatel2107:
I've followed the instructions and added in a trusted publisher (_I think_)

ZedThree (approved):

ZedThree:
@bpatel2107 Should we get this in and then release 0.7.0?

bpatel2107:
There are quite a few smaller issues I am working on that I would like to get in. We can of course do another release as we have them quite far apart as is. Maybe in a week or two we can do 0.7.1?

ZedThree:
More frequent, smaller releases is fine. It's only a few button clicks to make a new one :)

bpatel2107:
I just want #324 merged and then we can publish as that is quite important. As soon as that is merged (should be done now) I'll make a new release.

PR #341, opened 2024-05-15
Bugfix/eigenfunction amp

bpatel2107:
Eigenfunctions were not always being normalised so I've added a test ensuring they are. 
Added in reader for the eigenfunction in GS2 when the fields are not loaded and a measure of the total field needed to normalise the fluxes as a function of time.


ZedThree (approved):
I can't comment on the physics side, but LGTM. Looks like there's just enough variation between the codes to make pulling out a common helper function not worth it?

bpatel2107:
We could probably pull it out and choose to do this in the initialisation of `GKOutput` itself, i.e. check if `Eigenfunctions` is defined and ensure it is re-normalised there

bpatel2107:
Turns out the order of this is important as we normalise the eigenfunctions to the field amplitude ( root of integral of the 3 fields squared). The issue us that with different code normalisations the amplitude of the fields relative to each other changes which means you need to map the eigenfunctions to the correct units and _then_ calculate the field amplitude

PR #343, opened 2024-05-16
Add in reader for nrg in h5 format

bpatel2107:
Adds reader for GENE `nrg` files in h5 format needed for GENE v3.0


bpatel2107:
Small issue with cases where the fluxes are spit out less often than the fields (we assume the other way around) but the h5 reader works as is. Will raise a separate issue

PR #345, opened 2024-05-17
Unpin scipy

ZedThree:
Apparently causes problems with Python 3.12

Also:
- test on Python 3.12
- bump CI actions


ZedThree:
Err, failing on a `pint` conversion error:

```
tests/databases/test_imas.py:200: in <module>
    direct_pyro = Pyro(gk_file=gkw_template, gk_code="GKW")
/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/pyrokinetics/pyro.py:177: in __init__
    self.read_gk_file(gk_file, gk_code, norms=self.norms)
/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/pyrokinetics/pyro.py:909: in read_gk_file
    self.numerics = self.gk_input.get_numerics()
/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/pyrokinetics/gk_code/gkw.py:4[25](https://github.com/pyro-kinetics/pyrokinetics/actions/runs/9130101234/job/25106203835#step:5:26): in get_numerics
    metric_terms = MetricTerms(local_geometry, ntheta=metric_ntheta)
/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/pyrokinetics/local_geometry/metric.py:185: in __init__
    self.set_field_aligned_covariant_metric()
/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/pyrokinetics/local_geometry/metric.py:683: in set_field_aligned_covariant_metric
    g_r_r + (self.dalpha_dr**2) * g_zeta_zeta
/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/pyrokinetics/local_geometry/metric.py:552: in dalpha_dr
    initial_units = self.d2alpha_drdtheta.units
/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/pyrokinetics/local_geometry/metric.py:5[26](https://github.com/pyro-kinetics/pyrokinetics/actions/runs/9130101234/job/25106203835#step:5:27): in d2alpha_drdtheta
    term1 = self.dB_zeta_dr * self.Jacobian * gcont_zeta_zeta / self.dpsidr
/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/pyrokinetics/local_geometry/metric.py:371: in dB_zeta_dr
    return (self.B_zeta / H) * (term1 + term2 + term3 + term4)
/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/pint/facets/plain/quantity.py:841: in __add__
    return self._add_sub(other, operator.add)
/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/pint/facets/plain/quantity.py:103: in wrapped
    return f(self, *args, **kwargs)
/opt/hostedtoolcache/Python/3.9.19/x64/lib/python3.9/site-packages/pint/facets/plain/quantity.py:739: in _add_sub
    raise DimensionalityError(self._units, "dimensionless")
E   pint.errors.DimensionalityError: Cannot convert from '1 / lref_major_radius' to 'dimensionless'
```

bpatel2107:
This is probably cos of the integrals in scipy. I think before `scipy.integrate.cumilative_trapezoid` used to drop any units but maybe now the units persist. Before I just saved the units and re-applied them when needed but we should probably test if units are kept or not but maybe we need to be more careful

bpatel2107:
Actually its the opposite with `trapezoid`, see m.w.e

With `scipy==1.10.1`
```python
import numpy as np
from scipy import integrate
from pint import UnitRegistry

units = UnitRegistry()
a = np.linspace(0, 10, 11) * units.meter
integral = integrate.trapezoid(a)
print(integral)
```

` 50.0 meter`

But with `scipy==1.13.0` we get

`50.0`

Odd that it went in opposite direct than expected

LiamPattinson:
I had a look into it, and it appears that the old SciPy integration routines were just thin wrappers to NumPy functions, so Pint quantities were handled naturally. The newer versions try to solve things in a more clever way, but rather than just passing the inputs directly to NumPy functions without checking them, they now call `np.asarray` or `np.asanyarray`, and this strips units. We could fix this ourselves by writing small wrapper functions using the `@ureg.wraps` decorator.

LiamPattinson:
I managed to fix the immediate problem, and a few other issues while I was at it:

- The imas tests were creating a bunch of local files using module-level code, and it was running before any tests even started. Moved these parts into fixtures and switched to using `tmp_path`.
- Fixed a SciPy warning for `integrate.simpson`, since it now requires the second argument to be explicitly named.
- Replaced some `assert np.allclose` calls to `np.testing.assert_allclose`, as it gives better information when things go wrong.

Unfortunately, it's now failing in the Fourier tests again. I'm also seeing over 12500 warnings, many of which of `UnitStrippedWarnings` caused by SciPy functions. This might actually be causing errors, as if a unit is stripped and then we re-normalise, it'll call `val * unit` instead of `val.to(unit)` and likely give us the wrong answer. I'll try to go through the project and fix up all the SciPy function calls early next week. If that doesn't fix the errors I'm seeing, we'll need to update the golden answer tests again (though given how often they fail, I think it would be preferable to replace them if we can).

LiamPattinson:
I think the Fourier tests failing was actually my fault, but I got them working again. I've reduced the number of warnings from over 12500 to less than 2000, and the majority of those are coming from IDSPy (unsure if that's something we can fix from our end). There are still a lot more avoidable warnings than I'm personally comfortable with, but it's a lot better than nothing. 

I'm happy for this to be reviewed -- @ZedThree did you want to add or change anything? Sorry for taking over the branch, in hindsight I should have raised a new PR into this one.

--

Edit: Nevermind... tests were passing on my machine, but no such luck with the CI.

LiamPattinson:
Fixed the golden answer tests, and then realised there were further problems with Python 3.12:

- `jetto-tools` depends on `pkg_resources` which was removed in 3.12. You can still access it by installing `setuptools`, but it warns that `pkg_resources` is a deprecated API. I've added `setuptools` for now, but somebody who knows the `jetto-tools` devs should get in touch and let them know about this issue.
- Issues with the plugin system. Pre-Python 3.10, `importlib.metadata.entry_points` returned a dict-like object, and you would query that to find entry point groups: `entry_points()[group_name]`. In 3.10 and 3.11, that changed to a function interface, `entry_points(group=group_name)`, but if you passed no args it would return a dict-like object for backwards-compatibility. That seems to have also been removed in Python 3.12, so I had to put in a runtime version check to keep things working on 3.9 and 3.12.

bpatel2107 (approved):
Nice! Thanks for tidying up some of the testing in the meantime!

bpatel2107:
@fcasson FYI about `jetto_tools` and python 3.12

> jetto-tools depends on pkg_resources which was removed in 3.12. You can still access it by installing setuptools, but it warns that pkg_resources is a deprecated API. I've added setuptools for now, but somebody who knows the jetto-tools devs should get in touch and let them know about this issue.

Happy to merge whenever. I think quite a few `UnitStrippedWarnings` are still there so they could be contributing to quite a few of the warnings. I few strategic `.magnitude` might drop that further. Is there a way to force those to raise as error so we can find where to put them?

LiamPattinson:
I hadn't thought of that, but that's actually a much better way to track down warnings. You can run:

```bash
$ pytest [path/to/test_file.py] -W error::pint.UnitStrippedWarning
```

The syntax for the warning filter is `action:message:subclass:module:linenumber`, so `error::pint.UnitStrippedWarning` will catch only `UnitStrippedwarning` (though with any message, in any module, on any linenumber) and upgrade it to an exception. The advantage of this is that you get a full stack trace telling you exactly where the warning was raised. I'd just been trying to figure it all out manually.

Should we hunt down those warnings in a separate PR, or would you prefer to do so here?

bpatel2107:
May as well do it here

LiamPattinson:
I think that should be all of them sorted, and I've added `error::pint.UnitStrippedWarning` to the project-wide warnings filter in `pyproject.toml`, so hopefully they won't creep back in again. There are now about 1800 errors, so that's another 200 down. I'm also seeing 1583 warnings of this form:

```
tests/databases/test_imas.py: 747 warnings
  /home/username/workspace/pyrokinetics/venv/lib/python3.12/site-packages/idspy_toolkit/converter.py:206: DeprecationWarning: The truth value of an empty array is ambiguous. Returning False, but in future this will result in an error. Use `array.size > 0` to check that an array is not empty.
    if field_value[()] in default_vals_list:
```

I don't know if we're using `idspy_toolkit` incorrectly, or if this is a problem that's out of our control. Solving that would be a great boost, but I think it's out of the scope of this PR.

bpatel2107:
@ycamenen do you know anything about these warnings?

bpatel2107:
I found the issue in `idspy-toolkit` and it essentially comes down to us setting things in the IDS as numpy `float` rather than a base python `float`.  We could set everything with type `float` and `list` to avoid this.

The fix on their end is quite simple so might suggest a tweak to their code.

bpatel2107:
Making a PR on `idspy-toolkit` so will merge this now

fcasson:
> @fcasson FYI about `jetto_tools` and python 3.12

Please try jetto-tools 2.1.0, all should be good now.

PR #346, opened 2024-05-20
Update bunit calculation to use analytic solution

bpatel2107:
Currently we are double counting a point in the loop integral `local_geometry.bunit_over_b0` but the way we do the integral with a trapezoidal style rule isn't great as we could do an analytic version.

Currently we have 
$\frac{B_{unit}}{B_0} = \frac{R_0}{2\pi r} \oint \frac{a}{R} \frac{dL_N}{\nabla r}$

where we just look at the point to point distance when calculating $dL_N$. However we have `dRdtheta` and `dZdtheta` so we can actually calculate $dL_N$ analytically. See minimal working example for a unit circle's diameter below

```python
import numpy as np

theta = np.linspace(0, 2 * np.pi, 256, endpoint=False)
dtheta = theta[1] - theta[0]

R = np.cos(theta)
Z = np.sin(theta)

dR = (np.roll(R, 1) - np.roll(R, -1)) / 2.0
dZ = (np.roll(Z, 1) - np.roll(Z, -1)) / 2.0
dL = np.sqrt(dR**2 + dZ**2)

integral = np.sum(dL) / (2 * np.pi)

print("Crude dL calculator:",  integral)

dtheta = np.diff(theta, append=theta[0] + 2 * np.pi)

dRdtheta = -np.sin(theta)
dZdtheta = np.cos(theta)

dL = np.sqrt(dRdtheta**2 + dZdtheta**2) * dtheta

integral = np.sum(dL) / (2 * np.pi)

print("Analytic dL calculator:", integral)
```

Gets us
```
Crude dL calculator: 0.9998996042161418
Analytic dL calculator: 1.0
```

Using the following will get us to floating point error with ntheta ~ 32 or so. 

```python

# Equilibrium file
eq_file = template_dir / "test.geqdsk"

# Kinetics data file
kinetics_file = template_dir / "jetto.jsp"

# Load up pyro object
pyro = Pyro(
    eq_file=eq_file,
    kinetics_file=kinetics_file,
    kinetics_type="JETTO",
    kinetics_kwargs={"time": 550},
)

# Generate local parameters at psi_n=0.5
pyro.load_local(psi_n=0.5, local_geometry="Miller")

nthetas = [4, 8, 16, 32, 64, 128, 256, 512]
for ntheta in nthetas:
    bunit_over_b0 = pyro.local_geometry.get_bunit_over_b0(ntheta)
    print(f"Bunit/B0 with ntheta = {ntheta:3d}: {bunit_over_b0}")
```

Gets us
```
Bunit/B0 with ntheta =   4: 3.521249858800794 dimensionless
Bunit/B0 with ntheta =   8: 3.5483625019362806 dimensionless
Bunit/B0 with ntheta =  16: 3.548375666161369 dimensionless
Bunit/B0 with ntheta =  32: 3.548375666156663 dimensionless
Bunit/B0 with ntheta =  64: 3.5483756661566623 dimensionless
Bunit/B0 with ntheta = 128: 3.5483756661566623 dimensionless
Bunit/B0 with ntheta = 256: 3.5483756661566623 dimensionless
Bunit/B0 with ntheta = 512: 3.548375666156663 dimensionless
```

@dake0795 @mgiacomin This does impact conversions between CGYRO<-> GS2/GENE slightly as Bunit/B0 goes from 3.54805277856171 to 3.548375666156663 (0.01% difference) for a specific case


src/pyrokinetics/local_geometry/local_geometry.py
```diff
@@ -565,16 +552,14 @@ def get_f_psi(self):
         """
```
d7919, line 566: Typo in doc string a few lines above -- "fractor" rather than factor.

src/pyrokinetics/local_geometry/local_geometry.py
```diff
@@ -528,25 +527,13 @@ def get_bunit_over_b0(self, ntheta: int = 256):
 
         """
 
-        theta = np.linspace(0, 2 * pi, ntheta)
+        theta = np.linspace(0, 2 * pi, ntheta, endpoint=False)
+        dtheta = np.diff(theta, append=theta[0] + 2 * np.pi)
 
         R, Z = self.get_flux_surface(theta=theta)
+        dRdtheta, dRdr, dZdtheta, dZdr = self.get_RZ_derivatives(theta)
```
d7919: In f_psi we get dRdtheta and dZdtheta as properties of self, is there a reason we don't do that here? Calculating dL appears in a few routines -- can we pull it out into a helper?

bpatel2107: This was to ensure that we always use the same `ntheta` when calculating the `Bunit/B0` as they `ntheta` could vary depending on the resolution used by the code. With the previous method this caused more variability but now it is less of a concern.

It was important to get this consistent regardless of resolution as `Bunit/B0` was used to convert between simulations so should ideally be resolution independent.

`get_f_psi` should probably use the simulation value of `ntheta` as that is what the codes would use. Though a similar argument could be made for `Bunit/B0` I think to get this right across different resolutions is less important as this isn't used for conversions.

src/pyrokinetics/local_geometry/local_geometry.py
```diff
@@ -528,25 +527,13 @@ def get_bunit_over_b0(self, ntheta: int = 256):
...
-        dZ = (np.roll(Z, 1) - np.roll(Z, -1)) / 2.0 * l_units
-
-        R *= l_units
-        Z *= l_units
-
-        dL = np.sqrt(dR**2 + dZ**2)
+        dL = np.sqrt(dRdtheta**2 + dZdtheta**2) * dtheta
 
         R_grad_r = R * self.get_grad_r(theta)
         integral = np.sum(dL / R_grad_r)
```
d7919: I suppose if one really wanted to get the "correct" result we could use something like `quad` to integrate a `def dL(theta): dRdtheta, _, dZdtheta, _ = self.get_RZ_derivatives(theta) ; return np.sqrt(dRdtheta**2 + dZdtheta**2)` type function?

d7919: This might at least allow one to measure the error and report if it looks like we're under-resolved?

bpatel2107: Quick gander at it using scipy gets you `3.548375666156662 dimensionless`. I guess this is better than specifying ntheta

src/pyrokinetics/local_geometry/local_geometry.py
bpatel2107: Pulled out `dLdtheta`

bpatel2107:
@d7919 I've switched to using `scipy.integrate.quad` which works quite nicely so we no longer need to worry about the resolution of `theta`

bpatel2107:
Pretty happy with the agreement now

image

As you increase `ntheta_geometry` in GS2 the difference between the codes plots scales like `1/ntheta_geometry`
image

d7919:
> Pretty happy with the agreement now
> 
> image
> 
> As you increase `ntheta_geometry` in GS2 the difference between the codes plots scales like `1/ntheta_geometry` image

Nice! It looks like an order of magnitude increase in ntheta_geometry improves the error by two orders. If the gradient is -2 then this is close to second order convergence?

bpatel2107:
Yeah I think you're right, I can't count the axes ticks very well, looks like 5 orders down, ~2.5 across so that checks out

bpatel2107:
I'll merge this now!

PR #352, opened 2024-06-02
correct nperiod calculation from nx0

arkabokshi:
`nx0 = 2*nperiod - 1`,  but we specify `nx0` which can be odd / even.  How do we correctly handle this?


arkabokshi:
This seems to handle the odd and even `nkx` "well", in that the mode is centered at `theta=0` as expected for this situation.  However for even `nkx`, the first `2pi` segment seems to be zero suggesting an indexing issue somewhere (for odd `nkx` this is fine).

bpatel2107:
Sorry this seems to have slipped through the cracks. This seems to be failing tests because the ballooning angle grid is not matching the golden answers. Do you know why?

bpatel2107:
@arkabokshi  is this still an issue?

bpatel2107:
Will close unless issue is raised again

PR #357, opened 2024-06-03
Bugfix: unit conversions using simulation units

LiamPattinson:
I've been experimenting with ways to keep units consistent across many simulations (Issue https://github.com/pyro-kinetics/pyrokinetics/issues/337) while also removing the need to repeatedly rebuild the pint units cache, which is significantly slowing down our tests (Issue https://github.com/pyro-kinetics/pyrokinetics/issues/326). I should have something to show for it shortly, but it's definitely something that should wait until after our next release.

In the process of this work, I found a bug in the `LocalGeometry` normalisation function:

```
>>> import pyrokinetics as pk
>>> pyro = pk.Pyro(gk_file=pk.gk_templates["GS2"])
>>> pyro.local_geometry
<class 'pyrokinetics.local_geometry.miller.LocalGeometryMiller'>(
type  = Miller,
psi_n = 0.5 dimensionless
rho = 0.5 lref_minor_radius
Rmaj = 3.0 lref_minor_radius
Z0 = 0.0 lref_minor_radius
a_minor = 1.0 lref_minor_radius
Fpsi = 0.0 bref_B0 * lref_minor_radius
B0 = 14.142135623730951 bref_B0
q = 2.0 dimensionless
shat = 1.0 dimensionless
beta_prime = -0.04 bref_B0 ** 2 / lref_minor_radius
dpsidr = 0.253546276418555 bref_B0 * lref_minor_radius
bt_ccw = 1 dimensionless
ip_ccw = 1 dimensionless
kappa = 1.0 dimensionless
s_kappa = 0.0 dimensionless
delta = 0.0 dimensionless
s_delta = 0.0 dimensionless
shift = 0.0 dimensionless
dZ0dr = 0.0 dimensionless
bunit_over_b0 = 1.01418510567422 dimensionless
>>> pyro.local_geometry.normalise(pyro.norms.gene)
>>> pyro.local_geometry
<class 'pyrokinetics.local_geometry.miller.LocalGeometryMiller'>(
type  = Miller,
psi_n = 0.5 dimensionless
rho = nan lref_major_radius
Rmaj = nan lref_major_radius
Z0 = nan lref_major_radius
a_minor = nan lref_major_radius
Fpsi = nan bref_B0 * lref_major_radius
B0 = 14.142135623730951 bref_B0
q = 2.0 dimensionless
shat = 1.0 dimensionless
beta_prime = nan bref_B0 ** 2 / lref_major_radius
dpsidr = nan bref_B0 * lref_major_radius
[...]

PR #358, opened 2024-06-10
Bugfix: Pin `idspy-dictionaries` version

LiamPattinson:
In PR https://github.com/pyro-kinetics/pyrokinetics/pull/357, all tests passed following the final push (https://github.com/pyro-kinetics/pyrokinetics/actions/runs/9372068998/job/25802746935). They then failed during the merge (https://github.com/pyro-kinetics/pyrokinetics/actions/runs/9415332642/job/25936125182), as `idspy-dictionaries` released a new version which is incompatible with our existing test data.

I'm not sure how to fix this one, but given our version of `idspy-toolkit` is pinned, the easiest solution was to pin `idspy-dictionaries` to the previous version. Perhaps we could raise an issue and come up with a longer term solution later?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ dependencies = [
     "pint-xarray ~= 0.3",
     "contourpy ~= 1.0",
     "xrft >= 1.0.0",
-    "idspy-dictionaries >= 34000.2.0",
+    "idspy-dictionaries == 34000.2.0",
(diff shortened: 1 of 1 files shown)


bpatel2107:
I've spoken to @ycamenen  and for now it is best to pin `idspy-dictionaries` until a new version of `idspy-toolkit` is released

bpatel2107 (approved):

Issue #337, opened 2024-05-10
Inconsistent units across multiple simulations


LiamPattinson:
In Issue https://github.com/pyro-kinetics/pyrokinetics/issues/326, we found that calls to `_build_cache` in functions such as `SimulationNormalisation.set_lref` were imposing a significant performance penalty. It's necessary to call this as these functions redefine units, and without `_build_cache`, pint won't recognise the change and will continue using the old units definitions.

I've found another reason beyond performance concerns to rethink this approach:

```python
def test_pyro_consistent_units():
    """Test that after loading a new geometry, old values are still valid."""

    # Create local geometry from G-EQDSK file
    pyro = Pyro(gk_file=gk_templates["GS2"], gk_type="GS2")
    pyro.load_global_eq(eq_templates["GEQDSK"], eq_type="GEQDSK")
    pyro.load_local_geometry(psi_n=0.5)

    # Take reference copies of some quantity that depends on lref
    R = pyro.local_geometry.R
    R_copy_1 = R.copy()
    R_meter_1 = R.to("meter")

    # Load in a different equilibrium file and generate a new geometry
    pyro.load_global_eq(eq_templates["TRANSP"], eq_type="TRANSP")
    pyro.load_local_geometry(psi_n=0.95)

    # Make new reference copies of the _old_ quantity
    # This should not have changed!
    R_copy_2 = R.copy()
    R_meter_2 = R.to("meter")

    # Test that our quantities are unchanged
    np.testing.assert_array_equal(R_copy_1.magnitude, R_copy_2.magnitude)
    assert R_copy_1.units == R_copy_2.units
    np.testing.assert_array_equal(R_meter_1.magnitude, R_meter_2.magnitude)
```

What we find is that the first two assertions pass, as the numbers and unit name of `R` haven't been modified. However, the third assertion fails, as the definition of `lref_minor_radius0000` (or whatever it gets called) has changed behind the scenes, and therefore `R` has been invalidated. I'm not sure if this is currently causing any internal issues with Pyrokinetics, but it is a surprising and unintuitive behaviour.

--- 

To solve this, we could modify functions such as `Pyro.load_*` so that they always generate a new `norms` object, as after calling those functions we're always looking at a brand new simulation anyway. We'd also have to modify `LocalGeometry.from_global_eq` etc to do this.

[...]

PR #359, opened 2024-06-11
Feature/gene trpeps geometry file

bpatel2107:
Adds ability to read in `trpeps` from the geometry output file (`miller.dat`, `circular.dat`, `tracer_efit.dat`). 

It seems that when using `tracer_efit` geometry `trpeps` isn't written in the `parameters.dat` file (this seems like a bug tbh) so it needs to be read in from `magn_geometry` file. This adds that in.

Note this currently looks for files with the same extension as the input file so if you do 
`pyro = Pyro(gk_file="parameters")`

It would look for a file called `tracer_efit`, which won't exist.

If you do `pyro = Pyro(gk_file="parameters.dat")`, it will look for `tracer_efit.dat`. Basically this only really works if you specify the extension. We could search for the geometry file so it would work for just `parameters` if need be, so let me know if anyone would prefer that.

### DIFF
--- a/docs/examples/example_imas.py
+++ b/docs/examples/example_imas.py
@@ -123,7 +123,7 @@ def compare_pyro_run_nl(og_pyro, new_pyro, code):
 reference_values = {
     "tref_electron": 1000.0 * units.eV,
     "nref_electron": 1e19 * units.meter ** -3,
-    "lref_minor_radius": 1.5 * units.meter,
+    "lref_major_radius": 3.0 * units.meter,
     "bref_B0": 2.0 * units.tesla,
 }
 
--- a/docs/examples/example_imas.py
+++ b/docs/examples/example_imas.py
@@ -122,7 +122,7 @@ def compare_pyro_run_nl(og_pyro, new_pyro, code):
 
 reference_values = {
     "tref_electron": 1000.0 * units.eV,
-    "nref_electron": 1e19 * units.meter ** -3,
+    "nref_electron": 1e19 * units.meter**-3,
     "lref_major_radius": 3.0 * units.meter,
     "bref_B0": 2.0 * units.tesla,
 }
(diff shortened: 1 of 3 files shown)


venkiteshayyar:
Facing a couple of issues : 
1. In the following line : [https://github.com/pyro-kinetics/pyrokinetics/blob/feature/gene_trpeps_geometry_file/src/pyrokinetics/gk_code/gene.py#L254](line) 
using a file like `parameters_0002` doesn't convert to `tracer_efit_0002`. it drops the _0002

Something like this should fix it
```
sfx = original_filename.stem.split('_')[-1]
(original_filename.parent / geometry_type).with_suffix('_%s'%(sfx))
```


2. Proceeding forward (after renaming the file) , when using 
```
gkdict = pyro_to_imas_mapping(
        pyro,
        comment=f"Testing IMAS GENE",
        ids=gkdict
    )
```

the code fails as : 

Cannot convert from 'lref_minor_radius_parameters_00020000 * rhoref_pyro_parameters_00020000 * tref_electron_parameters_00020000 / elementary_charge' ([length] ** 4 * [mass] / [current] / [time] ** 3) to 'rhoref_gs2_parameters_00020000 * tref_electron_parameters_00020000 / elementary_charge / lref_major_radius_parameters_00020000' ([length] ** 2 * [mass] / [current] / [lref] / [time] ** 3)

bpatel2107:
I think I've fixed the issue with the `_0000` style files. Can you run again and see?

The normalisation issue is odd as it should be able to convert this. I wonder if it is due to the values of `major_R` and `minor_r` though I don't have any issues converting the output. If you're still having issues then I would need check the input file.

venkiteshayyar:
First fix works well ! 
Second problem persists and shows up here : 
In this [line](https://github.com/pyro-kinetics/pyrokinetics/blob/unstable/src/pyrokinetics/units.py#L118), `value` is not set 

This gives the error in line 121

bpatel2107:
I think this is related to #337. Will investigate

bpatel2107:
Got confirmation this works, will merge now

PR #362, opened 2024-06-14
Add check for box_size when no kx is set with a test.

bpatel2107:
TGLF doesn't set a kx grid which can break calculations of `BOX_SIZE` or `jtwist`. Here we check if `kx==0` and set the radial box multiplier to 1 in these cases.

Added a test to check this.


hgd511 (approved):
Looks good to me.

PR #364, opened 2024-06-17
Direct matching of linear GKOutput between codes

bpatel2107:
This is one I've been waiting to do for a while.

We now have some benchmark cases where we get very good agreement between CGYRO, GENE and GS2 outputs. In this PR we directly compare the outputs and get a good agreement for the eigenvalues, eigenfunctions and flux ratios.

### Things changed for each code
**GENE:**
- Tweak definition of beta to always use `ne` 
- Map from field aligned coordinate `z` to geometric angle `theta` (IMAS standard`)

**CGYRO**
- Tweak definition of beta to always use `ne * Te`
- Undo CGYRO's internal flux normalisation via $\int \phi^2 w_\theta d\theta$ to allow for our own normalisation via the IMAS defn.
- Map output `theta` grid back to geometric `theta` grid when `CONSTANT_STREAM_FLAG=1`

**GS2**
 - Map back to geometric `theta` when `equal_arc = True`
 - Scale linear fluxes by `2 * nperiod - 1` (seems to be need due to way of defining flux surface averages/integrals
 
 ### Electromagnetic GA-STD/Cyclone base case style run
Below I plot several outputs from each code for a range of `ky` from ~0.1 --> 10. Note that all the values are calculated independently from each other. The agreement between the codes is quite excellent below. The CGYRO eigenfunction does tail off faster than the others here.

The agreement between the eigenfunctions persist at lower ky and below I show the ratio of A|| and phi for each ky.

If I plot the ratio of the fluxes to the total flux heat flux (summed over fields and species) for each code, we see they all agree really well.

**However**, the absolute fluxes seem to be out. The definition here aims to match the IMAS definition of the linear fluxes. These are important to get right for things like saturation rules and such.

_IMAS definitions:_

Plotting the ratio of each codes fluxes to the CGYRO flux (note I am not saying CGYRO is correct. I chose it because was in the middle of the other two) you can see there is a constant factor 2 going is in opposite directions. Note at $k_y\rho_s > 1$, $Q_i \sim 0.0$ so we're comparing 0s there hence the poor agreement.

I guess we need to figure out what the definitions of fluxes are different for the code outputs and how they map to the IMAS defn.

 ### Electromagnetic highly shaped plasma
I perform a similar study for a more complicated plasma shape. I won't go into detail but overall we see very similar behavior in general; good agreement in most outputs with factor ~2 differences between the absolute fluxes.
 

The agreement is a bit more wobbly, especially at lower ky, but that is too be expected for more complicated plasmas and likely detailed resolution scans there would be required.

_side note_: We get good matches when the codes fiddle with the theta grids, shown below.

### Summary
We are in a great position right now to have a multi-code direct comparison with all the units and normalisations automatically handled. 

There is one point on the absolute linear fluxes but it seems like there are constant factors of 2 missing that I have yet to understand so any input on here would be particularly useful.

The tests will fail but that is due to the need for new golden answers. I will wait for this flux issue to be handled first. It would be nice to get Stella and GKW on here too to confirm this for them. 

I should stress that the numerical set up for each code is difficult to match given the different velocity spaces here but I have done my best. I've attached the input files used for the GA-STD/CBC style case.
[GA-STD.zip](https://github.com/user-attachments/files/15870220/GA-STD.zip)


bpatel2107:
Few bits of input would be useful here

@mrhardman does STELLA also have an similar `equal_arc` option as that may need to be implemented here. If you would be happy then I could include STELLA on these plots.

@ycamenen I would like to get GKW on here too. I will give a go at compiling/running but may come to you if I fail. Does the direct GKW output flux match the IMAS definition. It would help to narrow down which codes are out by what factor.

d7919:
Nice work! Just a quick question -- on the eigenfunction plots is the x-axis label "theta" actually "theta/pi" (at least in some cases)?

bpatel2107:
> Nice work! Just a quick question -- on the eigenfunction plots is the x-axis label "theta" actually "theta/pi" (at least in some cases)?

Yes sorry I will update that to avoid confusion

mrhardman:
@bpatel2107

I would be happy for `stella` to be included - just keep us updated with the comparison if you need any further assistance.

Yes, I believe `stella` has and `equal_arc` style option. I think that the input parameter that you need is `zed_equal_arc` in the namelist here: https://github.com/stellaGK/stella/blob/6dab824089682d1b046e31d069bf0324a6e73ffe/zgrid.f90#L91-L93

@GeorgiaActon @mabarnes might lend a comment here.

bpatel2107:
Here are the input files for the EM shaped case
[Shaped_plasma.zip](https://github.com/user-attachments/files/15871992/Shaped_plasma.zip)

bpatel2107:
Managed to get GKW running. Some good news and some bad news.


**The good**
Maybe the fluxes are a little out but pretty good overall.








### The bad
GKW is about 5 times smaller than CGYRO


Really not sure what the issue is here now...

d7919:
Just a quick comment in case it helps others -- the reason the Q_i ratio plots don't look like flat lines is because the ky range includes electron scales where Q_i becomes ~0 so the ratio is not really numerically sensible.

src/pyrokinetics/gk_code/gs2.py
```diff
@@ -1297,9 +1297,20 @@ def _get_fluxes(
...
             fluxes[iflux, ifield, ...] = flux
 
+        if gk_input.is_linear():
+            jacob = raw_data["jacob"].data
+            grho = raw_data["grho"].data
+            theta = raw_data["theta"].data
+            theta_append = 2 * theta[-1] - theta[-2]
+            dtheta = np.diff(theta, append=theta_append)
+
+            flux_norm = np.sum(jacob * dtheta) / np.sum(jacob * dtheta * grho)
```
d7919: Note internally GS2 tends to use the trapezium rule rather than the left rectangular rule for integration. I'm sure it doesn't really matter here but just thought I'd mention it in case it explains any small differences in benchmarking etc.

bpatel2107:
So after a lot of time this is hopefully done for now!

The main issue with the fluxes was the normalisation of the Maxwellian used in the codes. Nonlinear this isn't an issue as the flux is basically just

$\Gamma^{NL} \sim g \chi $

where $\hat{\chi}$ comes is a generalised field and $\hat{g}$ is our non-adiabatic distribution. We also have

$\Gamma^{NL}  \sim g \chi$

The linear fluxes (weights really) are given

$\Gamma^{L} \sim \frac{\Gamma^{NL}}{\chi^2} \sim \frac{ g \chi}{\chi^2}$

Therefore you are effectively left with a ratio of $g/\chi$ and there is a sneaky Maxwellian factor in that ratio. Some codes have factors of $2\pi^{3/2}$. GS2 was also missing a grad_r factor.

Below is a high beta shaped case (different to earlier as this case is much more unstable to we're not worried about issues being close to stability)

image
image
image
image


Note Qi/Qe ~ 0 at higher ky so its not surprising they don't quite agree.

I don't have the ability to run GX and STELLA is going through some changes which is making it a tad more difficult to check but  I think it is fine. Input files are below

[input_files.zip](https://github.com/user-attachments/files/19009184/input_files.zip)

bpatel2107:
All the testing now works so I am keen to get this merged and put out a new release. Let me know any comments asap. (Need to prep for OSSFFE)

bpatel2107:
Finally done now! I think this is worth merging now and dealing with any potential fall out later. 

Thanks to @physycola we got GX on there and working too!

The input files can be found earlier in the PR.

Cyclone base case:

image
image
image
image
image


High shaped high beta case:

image
image
image
image
image

PR #365, opened 2024-06-18
Add method to enforce quasineutrality in LocalSpecies

bpatel2107:
Add in method to `LocalSpecies` that will enforce quasi-neutrality by modifying a single species' density and density gradient. I've added in tests to check that this actually works.

### DIFF
--- a/src/pyrokinetics/local_species.py
+++ b/src/pyrokinetics/local_species.py
@@ -172,6 +172,48 @@ class LocalSpecies(CleverDict):
             return False
         return True
 
+    def enforce_quasineutrality(self, modify_species: str) -> None:
+        """
+        Enforces quasineutrality by adjusting the density and gradient of one
+        species
+
+        Parameters
+        ----------
+        modify_species: str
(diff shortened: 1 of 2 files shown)


LiamPattinson (approved):
Looks great! The only change I can think of is maybe having `"electron"` as the default value for `modify_species`, as that's the one species we can always rely on being present, and I imagine it's more likely that people would want to keep their D/T/impurity ratios unchanged than to keep the electron density constant (although I'm totally open to being wrong about that).

bpatel2107:
Actually people generally want to keep the electron density/gradient fixed as that is what we generally know best from experiments. Usually people tend to adjust the impurities. Their larger Z means you only need small changes to their density/gradient to get a quasi-neutral plasma, which tends to have the smallest overall impact to the simulations.

It of course depends on what they want but I think having a default is dangerous as people should really make an active decision about which to change.

LiamPattinson:
Fair enough, in that case I retract my recommendation, and I'd say it's good to merge!

PR #366, opened 2024-06-18
Remove extra species namelist from inputs when removed

bpatel2107:
When removing a species for any reason, input files would not be modified to reflect this and the removed species would end up persisting when writing out files. Here we check to make sure the total number of species matches the number in `LocalSpecies` and remove additional ones if need be. Added some tests for each code too.


LiamPattinson (approved):
Sorry for the delay in checking this, it all looks good to me.

PR #370, opened 2024-06-20
Enhancements/gene fourier reading

bpatel2107:
Add in `miller_general` to `gene.py` and tweak some small issues it was having. I think there was a small bug in GENE cases where `major_R=1.0` and the setting of `Lref` but that should be fixed now.


bpatel2107:
Had manual confirmation that this works, will merge now.

PR #371, opened 2024-06-21
Set default norm convention when reading IDS GKOutput and add test

bpatel2107:
The `default_convention` was not being set when reading an IDS in which mean you couldn't write a new one out if the conversion would fail.

For example when reading a GKW IDS, no information about `R_major/a_minor` is stored anywhere meaning you can't convert between say `gkw` and `pyrokinetics` Normalisation Conventions. This is normally fine but when writing out an IDS, we convert the `GKOutput` to the `imas` convention and convert it back to the `default_convention` As this was not being set it defaulted `pyrokinetics` which would fail at converting.

This fixes that and adds a test.


hgd511:
Looks good to me, generating the IDS from pyro seems to work now with this change.

hgd511 (approved):

Issue #369, opened 2024-06-19
NumPy 2.0 Compatibility


LiamPattinson:
[NumPy 2.0](https://numpy.org/doc/stable/release/2.0.0-notes.html) was released a couple of days ago :tada:

You can test compatibility by installing into a fresh virtual environment as so:

```bash
$ python -m venv numpy2venv
$ source numpy2venv/bin/activate
$ pip install -e .[tests]
$ pip install --upgrade numpy
$ pytest
```

We're mostly compatible, with just two issues:

- We need to replace all instances of `np.asfarray(x)` with `np.asarray(x, dtype=float)` -- a very easy fix. The vim/sed command `:%s/asfarray(\(.*\))/asarray(\1, dtype=float)/g` will do it automatically.
- `test_poincare` is failing, which compares an array against golden answer data stored in a `.npy` file. The arrays are almost identical for the first 22,000 elements, but start disagreeing in the last 8,000. No idea what's going on there.

For now, installing Pyrokinetics will only grab NumPy versions up to 1.26, as some of our dependencies require `numpy == 1.x` (`pyloidal` is one of them, but I can fix that very easily). If somebody can figure out what's causing `test_poincare` to fail, we'll be ready to upgrade.

PR #372, opened 2024-06-21
Numpy 2.0 Compatibility

LiamPattinson:
Solves Issue https://github.com/pyro-kinetics/pyrokinetics/issues/369, following the [Numpy 2.0 migration guide](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html).

`test_poincare` was failing in NumPy 2.0 due to a change in NumPy type promotion rules. In NumPy 1.26, CGYRO output files were being read using `np.fromfile` with `dtype=float32`, and this precision was propagating throughout the code. In NumPy 2.0, this data would be promoted to `float64` at some point during the computation, and this revealed that floating point errors had been accumulating to produce a 10% difference in the final results. The failing test could be reproduced in NumPy 1.26 by casting the CGYRO output data to `float64` immediately after reading from file.

I fixed this by immediately promoting the data read from file to `float64`, and regenerating the golden answer data. NumPy 1.26 and NumPy 2.0 now agree and pass all tests.


bpatel2107 (approved):
Looks good!

LiamPattinson:
I also made similar precision fixes to GKW while I was at it, which doesn't seem to have affected any tests.

ZedThree (approved):

bpatel2107:
I imagine upgrading to numpy in the `pyproject.toml` may take a while though

ZedThree:
Do we need to worry about `float` vs `np.float64`?

ZedThree:
> I imagine upgrading to numpy in the `pyproject.toml` may take a while though

There's a spec for suggested minimum versions of the scientific ecosystem: [SPEC0](https://scientific-python.org/specs/spec-0000/).

According to that, we should be targeting numpy 1.24 for now

LiamPattinson:
> Do we need to worry about float vs np.float64?

I think `dtype=float` and `dtype=np.float64` are functionally identical in most functions, but the type promotion rules seem to change depending on whether you're looking at scalars or arrays in 2.0:

```python
# NumPy 1.26
>>> (np.float32(5.0) + 2.0).dtype
dtype('float64')
>>> (np.float32(5.0) + np.float64(2.0)).dtype
dtype('float64')
>>> (np.array([5.0], dtype=np.float32) + np.array([2.0])).dtype
dtype('float64')
>>> (np.array([5.0], dtype=np.float32) + np.array([2.0], dtype=np.float64)).dtype
dtype('float64')

# Numpy 2.0
>>> (np.float32(5.0) + 2.0).dtype
dtype('float32')
>>> (np.float32(5.0) + np.float64(2.0)).dtype
dtype('float64')
>>> (np.array([5.0], dtype=np.float32) + np.array([2.0])).dtype
dtype('float64')
>>> (np.array([5.0], dtype=np.float32) + np.array([2.0], dtype=np.float64)).dtype
dtype('float64')
```

I'm not sure why they changed it to work like this, and I'm not sure how this resulted in reduced precision in 1.26 but higher precision in 2.0. It's all a bit confusing for me.

LiamPattinson:
> I imagine upgrading to numpy in the pyproject.toml may take a while though

We have it set to `>= 1.20.3` currently, so the only thing stopping us from using 2.0 now is stricter requirements on our dependencies. `pyloidal` has `numpy ~= 1.24` (i.e `numpy >= 1.24; < 2`), but I can fix that whenever.

bpatel2107:
This seems to have broken #383 so I think this needs merging, is this good to go?

LiamPattinson:
Sorry about that, I'm currently having some minor issues with the tests on my machine and will have to get that fixed first. It might be worth raising a quick PR to fix pyloidal on 0.2.0 until I get it sorted.

bpatel2107:
It doesn't seem to be failing due to pyloidal and is using 0.2.1. The issue is in the use of `np.asfarray` which is still in unstable. I think merging this would fix that?

LiamPattinson:
Pyloidal was keeping the NumPy version below 2.0 as it had stricter versioning than Pyrokinetics. I suppose the better solution would actually be to set `numpy ~= 1.24` in `pyproject.toml`.

The issue I'm having right now is that the tests are failing on my machine when I try merging `unstable` into this branch, but I can't figure out why it's happening. It might be that a completely different dependency has also started breaking on NumPy 2.0.

bpatel2107:
I'll pin numpy in that PR for now. Can't see any reasons why it would fail. Maybe merge unstable into this and push to trigger the tests here.

LiamPattinson:
Never mind, I think I 'fixed' the problem. For some reason a bunch of IMAS tests fail when you're not using the standard Python `pip` and `venv` tools (I just started using [`uv`](https://docs.astral.sh/uv/)). I have no idea what could be causing that to happen though.

LiamPattinson:
Okay, I think this should be good to go now.

PR #373, opened 2024-07-03
Switch TGLF to use MillerTurnbull

bpatel2107:
TGLF uses `MillerTurnbull` when including squareness, **NOT** MXH which is what I had originally thought (silly me). 
This reverts it back to `MillerTurnbull`. It was this this originally so I must have been confused when changing it to MXH.


hgd511 (approved):
Yep, LGTM. Nice one on remembering the s_delta difference.

PR #374, opened 2024-07-05
Update to loop over all wavenumber/eigenmode and get fluxes/field to …

bpatel2107:
…return dimensions directly

IDS have the option of storing multiple `kx`/`ky` and multiple eigenmodes for a given wavenumber. Before we just loaded in the first `kx`/`ky` and mode but now we support loading them all in at once.

This is particularly important for TGLF which has multiple mode.


bpatel2107:
This is done. @ycamenen  I should note that currently for "nonlinear" runs, TGLF calculates a field intensity for each mode but the IMAS data dictionary only allow for a single mode to be stored so some info is lost

hgd511 (approved):
Yep, looks good. Nice to have the architecture for multiple modes there, as it helps us should we ever decide to allow for more than one mode in GK codes in the future.

PR #379, opened 2024-08-15
Bugfix/gk output field coord

### LINKED ISSUE #378: Shape mismatch when pyrokinetics reads CGYRO output files when some fields haven't been saved
We're facing an issue with pyrokinetics' reading of output files for non-linear CGYRO runs due to shape mismatch
 
We have many runs where only the file **bin.cgyro.kxky_phi** is written, while **bin.cgyro.kxky_apar** and **bin.cgyro.kxky_bpar** are not. But the flux file `bin.cgyro.ky_flux` seems to have the information of all the fields.
The mismatch is seen at this line :
https://github.com/pyro-kinetics/pyrokinetics/blob/unstable/src/pyrokinetics/gk_code/gk_output.py#L600
 
```
data_vars['particle'][0][0],data_vars['particle'][1][0].shape
('field', (2, 8, 330))
dataset_coords['field'][0],dataset_coords['field'][1].shape
('field', (1,))
```
 
Hence the failure at the line `pyro.load_gk_output()`
 
I've looked at the git issue here : https://github.com/pyro-kinetics/pyrokinetics/issues/333
and tried the suggestion :
`pyro.load_gk_output(load_fields=False)
` 
However, that seems to be causing another issue:
`"No variable named 'phi'. Variables on the dataset include ['particle', 'heat', 'momentum', 'species', 'flux', ..., 'kx', 'ky', 'energy', 'pitch', 'theta']"`
due to the line
https://github.com/pyro-kinetics/pyrokinetics/blob/unstable/src/pyrokinetics/databases/imas.py#L710
 
This happens at
```
gkdict = pyro_to_imas_mapping(
            pyro,
            comment=f"Testing IMAS %s"%(gkcode),
            ids=gkdict
        )
```
 

For the future, we can insist on saving all the files, but to deal with existing runs one would have to re-run to generate the missing apar and bpar files. 

Is there a way so that we can use existing runs with these missing field fields, without having the users to re-run? 
Perhaps bypassing the comparison ?

bpatel2107:
Currently pyro uses the `Field` object to determine the coordinate `field` in the `GKOutput` object. However, is the case where there is no electromagnetic field output but there is electromagnetic flux output (the default in CGYRO...) this breaks the code. 

Here we directly take the `field` coordinate from the `Coords` object which is determined from the input file. Fixes #378

### DIFF
--- a/src/pyrokinetics/databases/imas.py
+++ b/src/pyrokinetics/databases/imas.py
@@ -707,10 +707,12 @@ def get_nonlinear_fields(gk_output: GKOutput):
 
     for field in gk_output["field"].data:
         field_name = imas_pyro_field_names[field]
-        field_data_norm = gk_output[field]
 
-        # Normalised
-        fields[f"{field_name}_perturbed_norm"] = field_data_norm.data.m
+        if field in gk_output:
+            field_data_norm = gk_output[field]
+
+            # Normalised
+            fields[f"{field_name}_perturbed_norm"] = field_data_norm.data.m
 
     if field_data_norm.ndim == 4:
         fields = gkids.GyrokineticsFieldsNl4D(**fields)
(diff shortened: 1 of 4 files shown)


bpatel2107:
@venkiteshayyar can you confirm if this works for you? You can leave `load_fields=True` here

PR #380, opened 2024-08-19
Add check for new position of rhostar in cgyro output

bpatel2107:
Format of `out.cgyro.equilibrium` file has changed recently so this checks and tweaks the position to read in `rho_star`. This does not break backwards compatibility.


bpatel2107:
@venkiteshayyar have a go with this now

venkiteshayyar:
I can confirm that the Nans have gone away.

PR #382, opened 2024-08-28
GENE IMAS fixes

### LINKED ISSUE #381: Problems with GENE when only ExBrate is specified in external_contr
@bpatel2107 
We're facing an issue rendering the input parameter file with pyrokinetics.
The data is for a GENE non-linear run

The parameter file has the content 
```
 &external_contr
 ExBrate    =  -0.67461610
```
Hence, the dictionary gets only the value for 'exbrate' in
[Line 461](https://github.com/pyro-kinetics/pyrokinetics/blob/unstable/src/pyrokinetics/gk_code/gene.py#L461) in gene.py

This causes a key not found error in [line 494](https://github.com/pyro-kinetics/pyrokinetics/blob/unstable/src/pyrokinetics/gk_code/gene.py#L494)
 for `omega0_tor`

Speaking to GENE experts, I gather that specifying just one of these parameters is often part of the workflow.

I tried a simple, hacky fix like this. 
```
        for key in ['exbrate','omega0_tor','pfsrate']: 
            if key not in external_contr.keys():
                external_contr[key]=0.0
```
But it's causing further issues later, so I wonder if more changes are required.

bpatel2107:
Adds default to `omega0_tor` in GENE, fixes #381 

EDIT:
Also for IMAS data dictionaries, checks if nonlinear field are loaded and adds to data dictionary if available.

EDIT:
Now we carefully check the dimensions of the nonlinear fluxes and populate the ids entries that we can. Similarly when loading in the ids data we see what is available and add it accordingly. 

This does have a caveat that currently the IDS does not store the fluxes with dimensions  `Fluxes(field, species, ky, time)`. It stores either `Fluxes(field, species, time)` or `Fluxes(field, species, ky)` so we can't have both. Right now pyro chooses to load the former though that is up for debate. **Note that this does change the behaviour when loading an IDS in. Before we loaded in  `Fluxes(field, species, ky)` first**. This does not affect the writing out of data.

Ideally what we want is to do is store the full 4D flux in the IDS. @ycamenen would it be possible to add in something like `fluxes_3d_k_x_sum` as that is a common output for a lot of codes.

### DIFF
--- a/src/pyrokinetics/databases/imas.py
+++ b/src/pyrokinetics/databases/imas.py
@@ -751,6 +751,11 @@ def get_nonlinear_fluxes(gk_output: GKOutput, time_interval: [float, float]):
     fluxes = {}
     fluxes_2d_k_x_sum = {}
     fluxes_2d_k_x_k_y_sum = {}
+    fluxes_1d = {}
+
+    flux_k_x_sum_check = True
+    flux_k_x_k_y_sum_check = True
+    flux_sum_check = True
 
     if "time" in gk_output.dims:
         min_time = gk_output.time[-1].data * time_interval[0]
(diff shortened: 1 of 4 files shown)


bpatel2107:
@vayyar I've updated the code to handle the NL problem you were having

PR #383, opened 2024-09-10
IMAS Equilibrium/Kinetics Reader

bpatel2107:
Here we add support for the IMAS data dictionaries `equilibrium` and `core_profiles` which map to our `Equilibrium` and `Kinetic objects respectively. This essentially makes up fully IMAS compliant. (One could even imagine having an `EquilibriumWriter` and `KineticsWriter` though this is outside the scope of this work anyway)

I have included some basic tests on the readers and included a simple plot method for the `Kinetics` object. `Kinetics` objects requires the mapping from $\psi_n$ to $r/a$ which is not always stored in the output file (pFile, IMAS) so this info must be supplied via the `Equilibrium` object. Before we would pass the file name of the equilibrium but now I think it makes more sense to use an actual `Equilibrium` object.

- [x] Basic reader tests
- [x] Detailed reader tests checking actual values/units
- [x] Include rotation from `core_profiles`


bpatel2107:
This is good to go now! 

It does currently use `h5py` so we are limited to only read in HDF5 style formats. I _think_ that `idspy_toolkit` doesn't support `Equilibrium` or `CoreProfiles` just yet though @ycamenen could correct me if I am wrong and this would be trivial to switch over.

I believe the format of these IDS are relatively fixed but if there are major version changes in IMAS we can cross that bridge when we come to it.

ZedThree:
What file formats do IMAS files typically use? I believe they might also use JSON?

One option is to have an internal routine for doing all the parsing, and pass in a `dict` -- then it would be trivial to use either `h5py` or `json` to initially read the file (and later writing the file, if that's eventually required).

On the IMAS versions, presumably that's stored in the file, so we could have an explicit check for compatible schema?

src/pyrokinetics/equilibrium/imas.py
```diff
@@ -0,0 +1,237 @@
...
+        """
+        # Define some units to use later
+        psi_units = units.weber / units.radian
+        F_units = units.meter * units.tesla
+
+        if time_index is not None and time is not None:
+            raise RuntimeError("Cannot set both 'time' and 'time_index'")
+
+        with h5py.File(filename, "r") as raw_file:
+            data = raw_file["equilibrium"]
```
ZedThree, line 91: Could un-indent after these two lines?

src/pyrokinetics/kinetics/imas.py
```diff
@@ -0,0 +1,179 @@
...
+from ..equilibrium import Equilibrium
+from ..file_utils import FileReader
+from ..species import Species
+from ..typing import PathLike
+from ..units import UnitSpline
+from ..units import ureg as units
+from .kinetics import Kinetics
+
+
+class KineticsReaderIMAS(FileReader, file_type="IMAS", reads=Kinetics):
```
ZedThree, line 17: Needs a docstring too?

src/pyrokinetics/kinetics/imas.py
```diff
@@ -0,0 +1,179 @@
...
+    def read_from_file(
+        self,
+        filename: PathLike,
+        time_index: int = -1,
+        time: float = None,
+        eq: Equilibrium = None,
+    ) -> Kinetics:
+        """
+        Reads in IMAS profiles NetCDF file
```
ZedThree:
```suggestion
        Reads in IMAS profiles HDF5 file
```

src/pyrokinetics/kinetics/imas.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+
+        if eq is None:
+            raise ValueError(
+                dedent(
+                    f"""\
+                    {self.__class__.__name__} must be provided with an Equilibrium object via
+                    the keyword argument 'eq'. Please load an Equilibrium.
+                    """
+                )
```
ZedThree: Instead of `dedent`, just split the string over two lines:

```suggestion
                f"{self.__class__.__name__} must be provided with an Equilibrium object via"
                "the keyword argument 'eq'. Please load an Equilibrium."
```

src/pyrokinetics/kinetics/imas.py
```diff
@@ -0,0 +1,179 @@
...
+            )
+
+        if time_index is not None and time is not None:
+            raise RuntimeError("Cannot set both 'time' and 'time_index'")
+
+        with h5py.File(filename, "r") as raw_file:
+
+            data = raw_file["core_profiles"]
```
ZedThree:
```suggestion
        with h5py.File(filename, "r") as raw_file:
            data = raw_file["core_profiles"]
```

And un-indent after this?

src/pyrokinetics/equilibrium/imas.py
```diff
@@ -0,0 +1,237 @@
...
+        )
+
+    def verify_file_type(self, filename: PathLike) -> None:
+        """Quickly verify that we're looking at a IMAS file without processing"""
+        # Try opening the IMAS file using freeqdsk.geqdsk
+        filename = Path(filename)
+        if not filename.is_file():
+            raise FileNotFoundError(filename)
+        try:
+            raw_data = h5py.File(filename, "r")
```
ZedThree: I'm a little worried this technically leaves the file open. Maybe this instead?

```suggestion
            with h5py.File(filename, "r") as f:
                keys = f.keys()
```

bpatel2107:
The IMAS version _should_ be stored but is not always as far as I can tell.

So far I have only seen HDF5 formats but I think they might be moving away from that. Whenever IMASpy comes out we should be able to use that to parse any format. Until then we could imagine trying to convert to a dict. I haven't got any example of JSON files though so it is hard to code it up.

src/pyrokinetics/equilibrium/imas.py
```diff
@@ -0,0 +1,237 @@
...
+        except Exception as exc:
+            raise ValueError("Couldn't read IMAS file. Is the format correct?") from exc
+        # Check that the correct variables exist
+        if "equilibrium" not in list(raw_data.keys()):
```
ZedThree: Is there some more specific IMAS metadata we could check as well/instead?

src/pyrokinetics/pyro.py
```diff
@@ -200,6 +200,12 @@ def __init__(
         # Load global kinetics file if it exists
         if kinetics_kwargs is None:
             kinetics_kwargs = {}
+        if kinetics_type == "IMAS":
+            if not hasattr(self, "eq"):
+                raise ValueError(
+                    "Loading a Kinetics object from IMAS requires an Equilibrium to be loaded"
+                )
+            kinetics_kwargs["eq"] = self.eq
```
ZedThree: This feels like a bit of a smell to me, but I might need to think about it a bit more. The two things I'm not keen on:

1. can we be more specific than needing the whole equilibrium? Maybe just pass `psi` (or whatever is needed)? Or maybe this is feedback for upstream, that it would be better for this information to be stored in the file? Could this be something we _always_ pass?
2. hardcoding "IMAS" here -- what if we have other formats that need this information? Maybe this can be a property of the kinetics types so we can more generically check this?

Basically, this feels necessary but inelegant -- what are our constraints and options here?

ZedThree (commented):
Looks good to me

src/pyrokinetics/equilibrium/imas.py
bpatel2107: Yes I think you're right

src/pyrokinetics/equilibrium/imas.py
bpatel2107: The IMAS data is one layer below but we could add it in if need be?

src/pyrokinetics/pyro.py
bpatel2107: I think this could be removed entirely as this will get caught later on (Line ~1560) by the error of not having a `Equilibrium`. Other formats like `pFile` needs an equilibrium too and that gets caught without issue.

src/pyrokinetics/equilibrium/imas.py
bpatel2107: Seems like I can't do that as the file and datatype get closed...

Here is what I get in the indent and outside the indent

`<HDF5 group "/equilibrium" (113 members)>`
`<Closed HDF5 group>`

The second doesn't allow me to read any data

ZedThree: Ah right, of course, sorry! Yeah, you'd need to read all the data up-front, so maybe not then

src/pyrokinetics/pyro.py
bpatel2107: Removed

src/pyrokinetics/equilibrium/imas.py
bpatel2107: I've added in a test going a layer deeper

src/pyrokinetics/kinetics/imas.py
bpatel2107: Done

src/pyrokinetics/kinetics/kinetics.py
```diff
@@ -1,5 +1,7 @@
 from typing import List, Optional
 
+import matplotlib.pyplot as plt
```
LiamPattinson: `matplotlib` is imported within the function that uses it, so we don't need it here. Normally I'd prefer to keep all imports at the top of the file, but I recall running into issues earlier with Pyrokinetics taking several seconds to import, and Matplotlib was one of the worst culprits.

LiamPattinson (approved):
I had a quick look through and couldn't find any major issues. LGTM!

src/pyrokinetics/kinetics/kinetics.py
bpatel2107: I've added it under a `if` statement as it is needed for some type checking but it seems to be failing the tests now. I've done that same thing as `equilibrium` so I'm a tad confused

bpatel2107: Figured it out

PR #389, opened 2024-10-23
Routines to compute field line displacement and correlation length

mgiacomin:
This PR adds a couple of routines to the `Diagnostics`, which can be used to evaluate the radial displacement of magnetic field lines after half poloidal turn and the radial correlation length of magnetic perturbations. These quantities are linked to the nonzonal transition described in _Pueschel et al. Phys. Rev. Lett. 110, 155005 (2013)_ .

@bpatel2107 I don't know while tests failed in the last commit where I simply added an example. Note that tests passed in the commit before and the example is not covered by any test. 

@dake0795 Did you try these new routines? Are they working smoothly?


src/pyrokinetics/diagnostics/__init__.py
```diff
@@ -391,6 +391,227 @@ def ideal_ballooning_solver(self, theta0: float = 0.0):
...
+
+        import pint_xarray  # noqa
+
+        if self.pyro.gk_output is None:
+            raise RuntimeError(
+                "Diagnostics: Please load gk output files (Pyro.load_gk_output)"
+                " before using any diagnostic"
+            )
+
+        if self.pyro.gk_code not in ["CGYRO", "GS2", "GENE"]:
```
bpatel2107: Why is it only available for these codes? The 4D fields are spit out for other codes too (STELLA/GKW)

mgiacomin: Yes, that's legacy. I will took that out.

bpatel2107:
@mgiacomin the tests are failing due to an issue in the Freeqdsk library. Try merging unstable into this branch and then pushing again

Overall this looks nice! I think it all works but it may be worth looking at if we can speed this up if this is quite slow. Haven't tried it myself yet. Have you benchmarked it against a "known" case?

dake0795:
Haven't looked this myself yet unfortunately. 

I managed to talk to MJ Pueschel at APS and he is planning to have a look at a STEP simulation (/marconi_scratch/userexternal/dkennedy/contribs/MJ/) with his own NZT diagnostics. We could benchmark against this.

mgiacomin:
After merging `unstable`, tests are successful, thanks!

I have only tested these routines on STEP data. As Dan just wrote, he will carry out a proper benchmark with MJ Pueschel.

bpatel2107:
Will wait for @dake0795 to check but happy to merge and then optimise later as needed.

PR #390, opened 2024-10-23
Enhancements/add gk output data

bpatel2107:
Lets you programmatically add data to the `GKOutput` object if you it is not a default output. All that is needed is the raw data, units and dimensions of the data. This does require the output to use existing dimensions but it would be easy to add new dimensions though I can't see any situation where we would need that.

For an example I add a Reynolds and Maxwell stress from GENE


bpatel2107:
@dake0795 can you check if this works for your stress calculation in GENE?

PR #392, opened 2024-11-05
changed np.isclose() to np.allclose() for calculating normed gradients

rijudutta:


bpatel2107:
This doesn't quite avoid the issue with the divide by 0 issue that could come up. If you imagine you ask for the temperature gradient (`1/T dT/dr`) at a few points and one of those points is 0, so `T = [2.0, 1.0, 0.0]` then this would not be caught by the `allclose` but would still result in a divide by 0. Also the thing you are returning in the if statement would have a different shape. i.e. its a gradient at one point rather than the `n` points you requested.

Probably worth being a bit more clever about replacing the 0s here to avoid issues going ahead.

Might be worth changing line 52 on the check for charges too?

rijudutta:
Hm, that makes sense, I'll take a look at it.

Best,
Riju

d7919:
Would something like

```python
return np.where(np.isclose(field_value, 0.0), 0.0 / units.lref_minor_radius, (-1.0 / field_value) * (gradient / self.grad_rho(psi_n)))
```

do it?

PR #393, opened 2024-11-18
Add in deuterium mass if not in Kinetics object - Fast and Furious edition

bpatel2107:
When setting up a reference values from a simulation (like TRANSP), pyro currently assumes that there will be Deuterium in the plasma. This are cases with Tritium/Hydrogen only plasmas so we add a check to ensure that we can handle these cases.

Add a simple check so when deuterium is not located in `Kinetics` object, we determine its mass from tritium/hydrogen and then add it to the physical reference values (note it is already in the simulation reference values by default).

EDIT

A few small bugs were addressed here and this is now basically about something else...
- Hydrogen species were not caught here in TRANSP and now they are
- Helium-3 had the wrong key name so that is now fixed
- A bigger change is that now we add in fast ion species in TRANSP which were not actually included before @juanruizruiz this may impact some of your runs...

The treatment of fast ions is probably best left as a separate PR but to summarise currently most codes treat them as a hot maxwellian. The issue here is that choice of how you define the temperature here is not well defined.

One could (and we currently) do in TRANSP
$T_{fast} = p_{fast} / n_{fast}$

here $p_{fast}$ which goes into the Grad-Shafranov solver in TRANSP is defined as (maybe this is a TRANSP specific thing...)

$p_{fast} = U_{\perp} + U_{||}/2$

where $U$ is the energy density in the perpendicular/parallel direction.

Doing so matches the "fast ion pressure" output from TRANSP
image

But the overall energy content doesn't match as you need, see below

$U_{fast} =  U_{\perp} + U_{||}$
image

I think the issue boils do to the choice of distribution function. Technically with maxwellian distributions we should have

$p = \frac{2}{3}  U$

and we do this for JETTO

https://github.com/pyro-kinetics/pyrokinetics/blob/765d0f1032470937926431847a9f87efefcdcd41/src/pyrokinetics/kinetics/jetto.py#L101-L109

which is technically different

JETTO:
$T_{fast} = \frac{2}{3} U_{fast} / n_{fast} = \frac{2}{3} (U_\perp + U_{||}) / n_{fast}$

TRANSP:
$T_{fast} = (U_\perp + U_{||}/2) / n_{fast}$

Now I'm not 100% sure what the right answer is or if there is even a right answer... Any thoughts on this would be useful!

### DIFF
--- a/src/pyrokinetics/kinetics/transp.py
+++ b/src/pyrokinetics/kinetics/transp.py
@@ -166,16 +166,13 @@ class KineticsReaderTRANSP(FileReader, file_type="TRANSP", reads=Kinetics):
                 if split_name[-1] == "SINGL":
                     name = "impurity"
                     impurity_charge = int(kinetics_data["XZIMP"][time_index].data)
-                    impurity_charge_func = (
-                        UnitSpline(
-                            psi_n,
-                            impurity_charge
-                            * unit_charge_array
-                            * units.elementary_charge,
-                        ),
+                    impurity_charge_func = UnitSpline(
+                        psi_n,
+                        impurity_charge * unit_charge_array * units.elementary_charge,
                     )
+
                     impurity_dens_data = (
-                        kinetics_data[f"NIMP"][time_index, :].data * units.cm**-3
+                        kinetics_data["NIMP"][time_index, :].data * 1e6 * units.m**-3
                     )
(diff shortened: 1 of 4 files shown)


fitz-culham:
image

bpatel2107:
@hgd511 Are you happy to merge? We can leave the rotation pressure as a separate PR

PR #395, opened 2024-12-17
Bicoherence diagnostics

bpatel2107:
Here we add in an auto-bicoherence/cross-bicoherence diagnostic for nonlinear simulation outputs which gives information about the level of mode coupling between two different modes.

$$b^2(X) = \frac{|B(k_{x1}, k_{y1},k_{x2}, k_{y2})|^{2}}{\langle
    |X(k_{x1}, k_{y1})X(k_{x2}, k_{y2})|\rangle^2 \langle|{X(k_{x3}, k_{y3})
    \rangle|^2}}$$

where $B$, the bispectrum is given by

$$
    B(k_{x1}, k_{y1},k_{x2}, k_{y2}) = \langle X(k_{x1}, k_{y1})X(k_{x2}, k_{y2})
    \overline{X(k_{x3}, k_{y3})}\rangle
$$

is the bispectrum
with $k_{x3} = k_{x1} + k_{x2}$, $k_{y3} = k_{y1} + k_{y2}$, 
$\overline{X}$ denoted a complex conjugate 
$\langle\rangle$ denoting an average over many realisations.

If the modes at $(k_{x1},k_{y1})$, $(k_{x2},k_{y2})$ and $(k_{x3},k_{y3})$ are not  coupled then that complex triple product will have random phases so will average to zero over many realisations, but if there is strong coupling then this average will have a non-zero value. This can then be normalised to the total amplitude of the fluctuations such that $b^2$ runs between 0 (no phase coupling) and 1 (entirely phase coupled). It should be noted that bicoherence cannot determine the direction of energy transfer but only if modes are coupled.

We currently calculation this and the cross-bicoherence. Right now we have a bit of redundancy as the bicoherence is symmetric when swapping $k_{x1}$/$k_{x2}$ and $k_{y1}$/$k_{y2}$ so we could hit memory issues.

We currently return a xarray Dataset with the bicoherence and phase data which has dimensions (`kx1`, `kx2`, `ky1`, `ky2`).

Plotting this is tricky but can be informative. Here is an example output for $phi$ for a GA-STD for the lowest `ky1=0.05`, `kx1=0.0` which has strong coupling with `ky2=0.05`, `kx2=0.05` suggesting that the mode at `ky3=0.1`, `kx3=0.05` is coupled. 
Similarly there is coupling at `ky2=0.0`, `kx2=-0.05` indicating there is coupling to the zonal field probably to higher `kx`

image

### DIFF
--- a/docs/examples/bicoherence.rst
+++ b/docs/examples/bicoherence.rst
@@ -113,7 +113,8 @@ like for :math:`\delta \phi`, :math:`\delta n_e`, :math:`\delta v_e` by doing
     data = diagnostics.cross_bicoherence(phi, density, velocity)
 
 
- If a single is not stationary (like during the linear phase of
- a nonlinear simulation), it is possible to examine the bicoherence
- of the phase of the fluctuations :math:`\hat{X} = X / |X|`  by setting
+
+If a single is not stationary (like during the linear phase of
+a nonlinear simulation), it is possible to examine the bicoherence
+of the phase of the fluctuations :math:`\hat{X} = X / |X|`  by setting
 `stationary=False` as a kwarg.
--- a/docs/examples/bicoherence.rst
+++ b/docs/examples/bicoherence.rst
@@ -33,11 +33,12 @@ coupled). It should be noted that bicoherence cannot determine the direction of 
 transfer but only if modes are coupled.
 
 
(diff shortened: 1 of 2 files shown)


bpatel2107:
`idspy_toolkit` does not seem compatible with `numpy=2.2.0` so the tests are failing. Works on 2.1 so could pin there until it is fixed

PR #396, opened 2024-12-17
Pin numpy to <2.1 and remove some warnings with numpy>=2.0

bpatel2107:
Tests in #395 failing with python>3.9 due to deprecation warning becoming an error in `numpy=2.2.0` inside `idspy_toolkit`.

```
if field_value[()] in default_vals_list:
ValueError: The truth value of an empty array is ambiguous. Use `array.size > 0` to check that an array is not empty.
```

Could maybe get this fixed in `idspy_toolkit` but with `imaspy` being made public and hopefully on PyPi soon its not worth doing too much to fix this.

Also fixed a bunch of small things causing other deprecation/unnecessary warnings but currently the above is causing over 1600 warnings...

### DIFF
--- a/.environment.yaml
+++ b/.environment.yaml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - python=3.9
+  - python=3.10
   - cmake
   - numpy
   - compilers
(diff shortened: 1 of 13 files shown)


ZedThree (approved):
The [SPEC-0](https://scientific-python.org/specs/spec-0000/) minimum supported versions suggests we can safely drop support for Python 3.9

bpatel2107:
I've dropped support for `python=3.9` here and tweaked some uses of `numpy.trapz` which is being deprecated in `numpy>2.3` so may as well jump the gun here.

PR #398, opened 2025-01-15
Start to add MXH treatment for GS2

d7919:
Attempting to add support for MXH with GS2. @ZedThree added this into GS2 itself since version 8.2.0 and is perhaps better placed to address what the pyro changes should actually look like but I thought I'd get the ball rolling.


d7919:
I'm also trying to tweak the mxh inputs for the next release of GS2 through https://bitbucket.org/gyrokinetics/gs2/pull-requests/1075 so we might want to just introduce support from GS2 8.3?

bpatel2107:
Nice! I'm guessing this will pass if you merge unstable back into this having merge #401

src/pyrokinetics/gk_code/gs2.py
```diff
@@ -67,6 +82,15 @@ class GKInputGS2(GKInput, FileReader, file_type="GS2", reads=GKInput):
         "beta_prime": 0.0,
     }
 
+    pyro_gs2_mxh_defaults = {
+        **pyro_gs2_miller_defaults,
+        "c0": 0.0,
```
bpatel2107: Why is there a `c0`? Is that not the first value in `cn`?

src/pyrokinetics/gk_code/gs2.py
```diff
@@ -688,11 +760,20 @@ def set(
...
+            self.data["theta_grid_parameters"]["geoType"] = 4
+
+            # Assign MXH values to input file
+            for key, val in self.pyro_gs2_mxh.items():
+                self.data[val[0]][val[1]] = local_geometry[key]
 
-        # Assign Miller values to input file
-        for key, val in self.pyro_gs2_miller.items():
-            self.data[val[0]][val[1]] = local_geometry[key]
+            # Do we need to deal with rho factor in derivatives here?
```
bpatel2107, line 834: Maybe? How are they defined in GS2, in pyro they are just $\frac{\partial c_n}{\partial \rho}$. No factor of $\rho$

src/pyrokinetics/gk_code/gs2.py
```diff
@@ -62,6 +64,15 @@ class GKInputGS2(GKInput, FileReader, file_type="GS2", reads=GKInput):
         "beta_prime": ["theta_grid_eik_knobs", "beta_prime_input"],
     }
 
+    pyro_gs2_mxh = {
+        **pyro_gs2_miller,
+        "c0": ["theta_grid_parameters", "c0_mxh"],
```
d7919: GS2 no longer has a separate c0 input so we should remove this and similar later variables.

src/pyrokinetics/gk_code/gs2.py
```diff
@@ -251,6 +273,52 @@ def get_local_geometry_miller(self) -> LocalGeometryMiller:
...
+
+        if self.data["theta_grid_eik_knobs"]["bishop"] != 4:
+            raise RuntimeError(
+                "Pyrokinetics requires GS2 input files to use "
+                "theta_grid_eik_knobs.bishop = 4"
+            )
+        if self.data["theta_grid_eik_knobs"]["irho"] != 2:
+            raise RuntimeError(
+                "Pyrokinetics requires GS2 input files to use "
+                "theta_grid_eik_knobs.bishop = 2"
```
d7919: Typo in message `bishop` --> `irho`

src/pyrokinetics/gk_code/gs2.py
```diff
@@ -72,6 +83,15 @@ class GKInputGS2(GKInput, FileReader, file_type="GS2", reads=GKInput):
         "beta_prime": 0.0,
     }
 
+    pyro_gs2_mxh_defaults = {
+        **pyro_gs2_miller_defaults,
+        "cn": [0.0, 0.0, 0.0, 0.0, 0.0],
+        "sn": [0.0, 0.0, 0.0, 0.0, 0.0],
+        "dcndr": [0.0, 0.0, 0.0, 0.0, 0.0],
+        "dsndr": [0.0, 0.0, 0.0, 0.0, 0.0],
+        "n_mxh": 5,
```
d7919: I don't know if we actually want to have the number of modes in this data set or if we should automatically determine it from the size of the cn/sn etc. lists.

d7919:
I think this might be ready for review now, although I've not tested this in any way

src/pyrokinetics/gk_code/gs2.py
bpatel2107: At the moment the defaults aren't massively important as they will likely get overwritten. As long as we set `n_moments` appropriately when creating the `LocalGeometryMXH` object it should be fine.

src/pyrokinetics/gk_code/gs2.py
```diff
@@ -251,6 +273,52 @@ def get_local_geometry_miller(self) -> LocalGeometryMiller:
...
+            )
+
+        mxh_data = default_mxh_inputs()
+
+        for (pyro_key, (gs2_param, gs2_key)), gs2_default in zip(
+            self.pyro_gs2_mxh.items(), self.pyro_gs2_mxh_defaults.values()
+        ):
+            mxh_data[pyro_key] = self.data[gs2_param].get(gs2_key, gs2_default)
+
+        # Do we need to scale derivatives by rho?
```
bpatel2107, line 308: The definition in pyro is 

`pyro.local_geometry.cn =` $\partial c_n/\partial r$

If the GS2 input is actually $r \partial c_n/\partial r$ then it needs rescaling.

bpatel2107:
This all looks mostly good. We have some roundtrip testing here

https://github.com/pyro-kinetics/pyrokinetics/blob/unstable/tests/test_roundtrip.py#L439

We could add GS2 to here. We would also need to tweak this file to allow for GS2 input file when using MXH

https://github.com/pyro-kinetics/pyrokinetics/blob/unstable/docs/examples/example_JETTO.py

@d7919 Would you be happy to add those in?

d7919:
@bpatel2107 I should hopefully have added what you've requested now.

bpatel2107:
I think this fails for MXH, we need to add in MXH here. Otherwise it looks good

https://github.com/pyro-kinetics/pyrokinetics/blob/b57208ab7fcdc392856e4b00c7764632b49d11b5/src/pyrokinetics/gk_code/gs2.py#L793-L797

bpatel2107 (approved):
LGTm, was a long road though

d7919:
> LGTm, was a long road though

Sorry about that! Usually my stupid broken commits are hidden locally :p

PR #399, opened 2025-01-16
Minor/find gs2 species type without reading the input data

d7919:
Try to fix a simple TODO and avoid inspecting input parameters to determine GS2 species types.


src/pyrokinetics/gk_code/gs2.py
```diff
@@ -1215,16 +1217,6 @@ def _get_fluxes(
...
 
-        # Get species names from input file
-        species = []
-        ion_num = 0
-        for idx in range(gk_input.data["species_knobs"]["nspec"]):
-            if gk_input.data[f"species_parameters_{idx+1}"]["z"] == -1:
-                species.append("electron")
-            else:
-                ion_num += 1
-                species.append(f"ion{ion_num}")
```
bpatel2107, line 1226: Don't we still need the ion names?

d7919: Actually, this species list and ion_num aren't used here, we get them from coords that is passed in.

bpatel2107 (approved):
LGTM

PR #402, opened 2025-01-21
Saturation rules - KBM MG rule

bpatel2107:
In this PR I add in the saturation rule from https://arxiv.org/abs/2404.17453 designed for KBM turbulence.

The model has two inputs ($Q_0$ and $\alpha$) and takes in a `PyroScan` object that has the dimensions ($k_y$ and $\theta_0$).  Note I have added the ability to load a `PyroScan` object by specifying a dictionary that maps the different parameter values to an input file so it is easier to load in pre-existing scans.

This should be extensible to any number of dimensions but this particular saturation rule requires those two dimensions.

I've written it in such a way that it should be straightforward to add in other saturation rules (TGLF sat for example). It is current imported from `diagnostics` but maybe it deserves its own folder...

@tomneiser this is what I was thinking a similar set up could be for TGLF. Could pass a `GKInputTGLF` file along with a `PyroScan` to get the fluxes.

I've included an example script and found that it matches T3D simulation results so pretty happy with it as is. Had to do quite a bit of fiddling with the outputs as it was trained using the default pyro units but there were subtleties in the unit conversion with ratios/definitions of the fields.

If the rule gets updated in the future we can likely adjust this and simplify the whole conversions.

```python
from pyrokinetics import Pyro, PyroScan
from pyrokinetics.diagnostics.saturation_rules import SaturationRules
import numpy as np

# Choose convention
output_convention = "pyrokinetics"

# Base file
base_dir = "/common/CSD3/ir-giac2/t3d/resolution/nth128/restart6/outputs"
base_file = "r3-ky0-th00-0.in"

pyro = Pyro(gk_file=f"{base_dir}/{base_file}")

base_ky = pyro.numerics.ky.to(pyro.norms.pyrokinetics).m / 2

# Set up ky and theta0 grid
param_1 = "ky"
param_2 = "th0"

kys = np.array([2, 3, 5, 10, 20, 30, 40, 50, 70, 100, 120, 140]) * base_ky
th0s = np.array([0, 0.1, 0.2, 0.4, 1.2, 3.14])

values_1 = kys
values_2 = th0s

# Create dictionary mapping input files to parameters in scan
runfile_dict = {}
for iky, ky in enumerate(kys):
    for ith0, th0 in enumerate(th0s):
        runfile_dict[
            (f"{param_1}_{values_1[iky]}", f"{param_2}_{values_2[ith0]}")
        ] = f"{base_dir}/r3-ky{iky}-th0{ith0}-0.in"

# Dictionary of param and values
param_dict = {param_1: values_1, param_2: values_2}

# Create PyroScan object
pyro_scan = PyroScan(
    pyro,
    param_dict,
    value_fmt="d",
    value_separator="",
    parameter_separator="-",
    file_name="",
    base_directory=base_dir,
    runfile_dict=runfile_dict,
)

# Add in path to each defined parameter to scan through
pyro_scan.add_parameter_key(
    param_1, "gk_input", ["data", "kt_grids_single_parameters", "n0"]
)

# Load outputs
pyro_scan.load_gk_output(output_convention=output_convention, tolerance_time_range=0.9)

# Create saturation object
saturation = SaturationRules(pyro_scan)

# Inputs for QL model
alpha = 2.5
Q0 = 25

# Must match convention
gamma_exb = 0.04380304982261718

gk_output = saturation.mg_saturation(
    Q0=Q0,
    alpha=alpha,
    gamma_exb=gamma_exb,
    output_convention=output_convention,
    gamma_tolerance=0.3,
    equal_arc_theta=True,
    theta0_dim="th0",
)

print("GK output", gk_output)
print("Heat flux calculation", gk_output["heat"])
print("Gamma flux calculation", gk_output["particle"])

```

~Currently this just returns an `xr.DataArray` with the heat flux as a function of species but this could/should be extended to be a general `GKOutput` object.~
Now the output is an xarray Dataset, currently with the heat flux and particle flux for each species

```
GK output <xarray.Dataset>
Dimensions:   (species: 2)
Coordinates:
  * species   (species) <U8 'electron' 'ion1'
Data variables:
    heat      (species) float64 [([tref] / [mref])**(0.5)·([tref] / [mref])**(0.5) * [mref] / [bref_B0])²·nref_electron·tref_electron/lref_minor_radius²] ...
    particle  (species) float64 [([tref] / [mref])**(0.5)·([tref] / [mref])**(0.5) * [mref] / [bref_B0])²·nref_electron/lref_minor_radius²] ...
Heat flux calculation <xarray.DataArray 'heat' (species: 2)>
<Quantity([0.09994335 0.17957855], 'nref_electron * rhoref_pyro ** 2 * tref_electron * vref_nrl / lref_minor_radius ** 2')>
Coordinates:
  * species  (species) <U8 'electron' 'ion1'
Gamma flux calculation <xarray.DataArray 'particle' (species: 2)>
<Quantity([0.01806394 0.01806394], 'nref_electron * rhoref_pyro ** 2 * vref_nrl / lref_minor_radius ** 2')>
Coordinates:
  * species  (species) <U8 'electron' 'ion1'

```

@arkabokshi FYI

### DIFF
--- a/docs/examples/example_saturation.py
+++ b/docs/examples/example_saturation.py
@@ -1,0 +1,77 @@
+from pyrokinetics import Pyro, PyroScan
+from pyrokinetics.diagnostics.saturation_rules import SaturationRules
+import numpy as np
+
+# Choose convention
+output_convention = "pyrokinetics"
+
+# Base file
+base_dir = "/common/CSD3/ir-giac2/t3d/resolution/nth128/restart6/outputs"
+base_file = "r3-ky0-th00-0.in"
+
+pyro = Pyro(gk_file=f"{base_dir}/{base_file}")
+
+base_ky = pyro.numerics.ky.to(pyro.norms.pyrokinetics).m / 2
+
+# Set up ky and theta0 grid
+param_1 = "ky"
+param_2 = "th0"
+
+kys = np.array([2, 3, 5, 10, 20, 30, 40, 50, 70, 100, 120, 140]) * base_ky
+th0s = np.array([0, 0.1, 0.2, 0.4, 1.2, 3.14])
+
+values_1 = kys
+values_2 = th0s
+
+# Create dictionary mapping input files to parameters in scan
+runfile_dict = {}
+for iky, ky in enumerate(kys):
+    for ith0, th0 in enumerate(th0s):
+        runfile_dict[
+            (f"{param_1}_{values_1[iky]}", f"{param_2}_{values_2[ith0]}")
+        ] = f"{base_dir}/r3-ky{iky}-th0{ith0}-0.in"
+
+# Dictionary of param and values
+param_dict = {param_1: values_1, param_2: values_2}
+
+# Create PyroScan object
+pyro_scan = PyroScan(
+    pyro,
+    param_dict,
+    value_fmt="d",
+    value_separator="",
+    parameter_separator="-",
+    file_name="",
+    base_directory=base_dir,
+    runfile_dict=runfile_dict,
+)
+
+# Add in path to each defined parameter to scan through
+pyro_scan.add_parameter_key(
+    param_1, "gk_input", ["data", "kt_grids_single_parameters", "n0"]
+)
+
+# Load outputs
+pyro_scan.load_gk_output(output_convention=output_convention, tolerance_time_range=0.9)
+
+# Create saturation object
+saturation = SaturationRules(pyro_scan)
+
+# Inputs for QL model
+alpha = 2.5
+Q0 = 25
+
+# Must match convention
+gamma_exb = 0.04380304982261718
+
+qflux = saturation.mg_saturation(
+    Q0=Q0,
+    alpha=alpha,
+    gamma_exb=gamma_exb,
+    output_convention=output_convention,
+    gamma_tolerance=0.3,
+    equal_arc_theta=True,
+    theta0_dim="th0",
+)
+
+print("Heat flux calculation", qflux)
--- a/docs/examples/example_saturation.py
+++ b/docs/examples/example_saturation.py
@@ -64,7 +64,7 @@ Q0 = 25
 # Must match convention
 gamma_exb = 0.04380304982261718
 
-qflux = saturation.mg_saturation(
+gk_output = saturation.mg_saturation(
     Q0=Q0,
     alpha=alpha,
     gamma_exb=gamma_exb,
@@ -74,4 +74,6 @@ qflux = saturation.mg_saturation(
     theta0_dim="th0",
 )
 
-print("Heat flux calculation", qflux)
+print("GK output", gk_output)
+print("Heat flux calculation", gk_output["heat"])
+print("Gamma flux calculation", gk_output["particle"])
--- a/src/pyrokinetics/diagnostics/saturation_rules.py
+++ b/src/pyrokinetics/diagnostics/saturation_rules.py
@@ -1,8 +1,9 @@
 from ..pyroscan import PyroScan
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
 import numpy as np
 from scipy.integrate import cumulative_trapezoid
+import xarray as xr
+
 
 class SaturationRules:
     r"""
@@ -10,24 +11,23 @@ class SaturationRules:
 
     Need a PyroScan object to apply the rule to
     """
-    
-    def __init__(self, pyro_scan):
+
+    def __init__(self, pyro_scan: PyroScan):
 
         self.pyro_scan = pyro_scan
 
     def mg_saturation(
-            self,
-            Q0: float = 25.0,
-            alpha: float = 2.5,
-            gamma_exb: float = 0.0,
-            output_convention: str = "pyrokinetics",
-            gamma_tolerance: float = 0.001,
-            equal_arc_theta: bool = True,
-            ky_dim: str = "ky",
-            theta0_dim: str = "theta0",
-            k_perp_in = None,
+        self,
+        Q0: float = 25.0,
+        alpha: float = 2.5,
+        gamma_exb: float = 0.0,
+        output_convention: str = "pyrokinetics",
+        gamma_tolerance: float = 0.001,
+        equal_arc_theta: bool = True,
+        ky_dim: str = "ky",
+        theta0_dim: str = "theta0",
     ):
-        if not self.pyro_scan.gk_output:
+        if not hasattr(self.pyro_scan, "gk_output"):
             self.pyro_scan.load_gk_output(output_convention=output_convention)
 
         data = self.pyro_scan.gk_output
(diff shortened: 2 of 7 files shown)


docs/examples/example_saturation.py
```diff
@@ -0,0 +1,79 @@
+from pyrokinetics import Pyro, PyroScan
+from pyrokinetics.diagnostics.saturation_rules import SaturationRules
+import numpy as np
+
+# Choose convention
+output_convention = "pyrokinetics"
+
+# Base file
+base_dir = "/common/CSD3/ir-giac2/t3d/resolution/nth128/restart6/outputs"
```
mgiacomin, line 9: This is probably fine because it provides just an example of usage, but most users wont be able to access to this folder.

src/pyrokinetics/diagnostics/saturation_rules.py
```diff
@@ -0,0 +1,247 @@
...
+        self,
+        Q0: float = 25.0,
+        alpha: float = 2.5,
+        gamma_exb: float = 0.0,
+        output_convention: str = "pyrokinetics",
+        gamma_tolerance: float = 0.001,
+        equal_arc_theta: bool = True,
+        ky_dim: str = "ky",
+        theta0_dim: str = "theta0",
+    ):
```
mgiacomin, line 29: I would suggest adding the following header to this function
```
"""
Please see doi:10.1017/S0022377824001107 for details on this quasi-linear model.
"""
```
or something similar.

src/pyrokinetics/diagnostics/saturation_rules.py
```diff
@@ -0,0 +1,247 @@
...
+            )
+
+            # Interp on to equal arc grid
+            if equal_arc_theta:
+                k_perp_interp = np.interp(theta, l_theta_long, k_perp_long)
+            else:
+                k_perp_interp = np.interp(theta, theta_long, k_perp_long)
+            for iky, ky in enumerate(kys.data):
+                # Technically k_perp / ky
+                k_perp[iky, itheta0, :, :] = k_perp_interp.m * ky * bunit_over_b0
```
mgiacomin, line 150: Why is `k_perp_interp` multiplied by `bunit_over_b0`? Is it required to convert it into pyro units?

mgiacomin (approved):
It looks good to me, thanks! I have added a couple of very minor comments, except maybe the bunit factor that multiplies kperp, which I do not understand. Probably it is related to how the Jacobian is defined in pyro?

src/pyrokinetics/diagnostics/saturation_rules.py
```diff
@@ -0,0 +1,247 @@
...
+
+        bmag = pyro.metric_terms.B_magnitude
+
+        if equal_arc_theta:
+            bmag_eq_arc = np.interp(theta_geo, theta_metric, bmag)
+            bmag_long = np.tile(bmag_eq_arc[:-1], 2 * nperiod - 1)
+            pyro_jacob = pyro.metric_terms.dpsidr * np.sqrt(g_tt_eq) / bmag_long
+            bmag_balloon = np.interp(theta, l_theta_long, bmag_long)
+        else:
+            g_tt_long = np.tile(g_tt[:-1], 2 * nperiod - 1)
```
d7919, line 160: Doesn't this `tile` approach mean the "long" grid isn't symmetric / doesn't fully cover a certain number of wraps. This might not matter, but suppose nperiod = 1 and `bmag` has three points (at theta = -pi, 0, pi) -- here we'll end up with bmag_long with just two points corresponding to theta =-pi and theta = 0. This is clearly an extreme example but if we're ever doing flux surface integrals / averages or similar with these extended arrays might we be losing a little bit of data that is actually needed for a complete answer? Again maybe these arrays are never used this way but it makes me a little nervous  (for example if someone introduces new uses of this pattern).

docs/examples/example_saturation.py
bpatel2107: Yeah I agree. I wonder if it is worth having data somewhere that we can pull and use as an example. This would be useful for basically a bunch of testing as we could have better quality testing.

src/pyrokinetics/diagnostics/saturation_rules.py
bpatel2107: Its more to do with the defn of `k_perp` and to match the thing used in GS2/CGYRO we need it. We could put it inside the actual `k_perp` method. I would need to delve into the exact way we have defined `k_perp` in our diagnostic but this is a little confusing I agree

bpatel2107: To elaborate:

In our CI testing, when we compare our calculated k_perp to the GS2 value we need the `bunit_over_b0`

https://github.com/pyro-kinetics/pyrokinetics/blob/unstable/tests/local_geometry/test_metric_terms.py#L199-L223

But when we compare to the `k_perp` in CGYRO we don't need the bunit_over_b0

https://github.com/gafusion/gacode/blob/29268b4b42b17cdc47376335e96bbde13e1857ee/cgyro/src/cgyro_equilibrium.F90#L260-L261

This is probably due to the definition of ky in the different codes. I have to manually add this factor into the QLGYRO implementation
https://github.com/gafusion/gacode/blob/6d8e83899c0c86fa4863f031aeeaa56eb6e0dceb/qlgyro/src/qlgyro_cgyro_cleanup.F90#L45

src/pyrokinetics/diagnostics/saturation_rules.py
bpatel2107: Technically this would be the case where we would always miss the very final point at the end of the field line. 

But I added 1 to nperiod here to avoid this issue
https://github.com/pyro-kinetics/pyrokinetics/blob/2ea9a1d7c445b9038a05b9af63eebb1fd548c9a1/src/pyrokinetics/diagnostics/saturation_rules.py#L78-L79

bpatel2107: This issue actually came up when building this in the interpolation I do later on mapping the geometric quantities onto the codes output theta grid.

d7919: So adding 1 to nperiod means you do cover the full extent of the "original" grid but it still means the total grid you end up with is missing a point so any integration over the full grid will not be quite as expected I guess? This tiling structure appears elsewhere as well.

bpatel2107: The aim of the tiling is to get the geometric quantities like $J$ and $|B|$ on the grid of the output simulation. These geometric quantities get calculated on a high resolution theta grid between $-\pi \rightarrow \pi$ (due to the way `MetricTerms` has been written) and then they get interpolated onto the `GKOutput` $\theta$ grid. All calculations including any sort of flux surface average is done on that grid, not the high resolution grid used to make the geometric quantities.

d7919: Ah OK that makes sense. I came across it looking at the k_perp metric terms method as Arka noticed the length of the output grid was not as expected. Scanning the code I think `k_perp` is actually the _only_ place using the tile structure apart from this new PR. Is there any danger about people misusing the k_perp output?

bpatel2107: Probably yes, its probably quite easy to fix to be honest.

To clarify what does GS2 do with its grid? If you have `ntheta` then from $-\pi \rightarrow \pi$ (inclusive of $\pi$) will have `ntheta+1` points? In total there are $\bigg((2n_{period} - 1) * n_{theta}\bigg) + 1$ points? 

That $+1$ is different to say CGYRO or GENE which just have $\bigg((2n_{period} - 1) * n_{theta}\bigg)$ points. I am happy to add on the final point but I am not sure what the best strategy is here...

bpatel2107: A note that GENE/CGYRO do include 0, but maintain the multiple of `ntheta` grid points by dropping the "final" theta grid point as we do here, meaning the final point is not an integer multiple of $\pi$ so $\theta_{final} mod 2\pi \ne 0$

d7919: Yes in GS2 we have `ntheta+1` points spanning the inclusive range from -pi to pi (and each subsequent 2 pi section). This means that in ballooning space we always have `theta(one_end) == - theta(other_end)`. This means that we always cover the poloidal section exactly an integer number of times. Provided the theta grid used in integrals is consistent then I guess it is unlikely to cause a major issue, but considering low resolution, low extent cases I could see there being slight errors in flux surface integrals when you drop the last point (without introducing any special handling).

bpatel2107: I've appended the final point on to the `k_perp` calculation now.

This won't help with cases like in CGYRO,GENE where the final point is missing so any integral over that domain will technically be slightly off but that's not really a pyro issue.

bpatel2107: If you want to match a code exactly so you model the same `theta` points then you may need to set up metric terms with one extra point i.e. to account for how the code sets up its theta grid. This would be the case in GS2 I think

```python
pyro.load_metric_terms(ntheta=pyro.numerics.ntheta + 1)
```

bpatel2107:
Will merge now! Maybe we can revisit a better way of "tiling" geometry outputs within metric terms itself. One can imagine wanting to get things like $|B|$ on an extended ballooning space grid.

PR #404, opened 2025-01-22
Add option to handle loading in adiabatic electrons in GS2/STELLA

bpatel2107:
Handles in adiabatic electrons cases in GS2/STELLA. The code was getting stuck on setting up the normalisation quantities as it couldn't find the electron species so couldn't make `tref_electron`.

The electron temp/density/mass is still set in the code so we can read them in, so we side step that here. 

We don't currently set them in the normalisations meaning you can't do `(1*tref_deuterium).to(units.tref_electron)` but I wonder if we should or not. Not sure to be honest.

This does mean we have to load it in the bespoke output convention and not convert to Pyro's default convention, see below

```python
from pyrokinetics import Pyro

pyro = Pyro(gk_file="cbc_linear_gs2.in")

pyro.load_gk_output(output_convention="gs2_bespoke")

print(pyro.gk_output.data)
```

So we can't convert the output to use say $T_{ref} = T_e$ or $m_{ref} = m_e$
```
<xarray.Dataset> Size: 18MB
Dimensions:                (theta: 73, kx: 1, ky: 15, time: 501, field: 1,
                            species: 1, energy: 16, pitch: 47, flux: 3)
Coordinates:
  * time                   (time) float64 4kB 0.0 0.5 1.0 ... 249.0 249.5 250.0
  * kx                     (kx) float64 8B 0.0
  * ky                     (ky) float64 120B 0.07071 0.1414 ... 0.9899 1.061
  * theta                  (theta) float64 584B -9.425 -9.076 ... 9.076 9.425
  * energy                 (energy) int32 64B 1 2 3 4 5 6 ... 11 12 13 14 15 16
  * pitch                  (pitch) float64 376B 0.001918 0.01006 ... 1.168 1.17
  * species                (species) <U4 16B 'ion1'
  * field                  (field) <U3 12B 'phi'
  * flux                   (flux) <U8 96B 'particle' 'heat' 'momentum'
Data variables:
    phi                    (theta, kx, ky, time) complex128 9MB [rhoref_gs2·tref_deuterium/e/lref_minor_radius] ...
    particle               (field, species, ky, time) float64 60kB [nref_deuterium·rhoref_gs2²·vref_most_probable/lref_minor_radius²] ...
    heat                   (field, species, ky, time) float64 60kB [nref_deuterium·rhoref_gs2²·tref_deuterium·vref_most_probable/lref_minor_radius²] ...
    momentum               (field, species, ky, time) float64 60kB [nref_deuterium·rhoref_gs2²·tref_deuterium/lref_minor_radius] ...
    growth_rate            (kx, ky, time) float64 60kB [vref_most_probable/lref_minor_radius] ...
    mode_frequency         (kx, ky, time) float64 60kB [vref_most_probable/lref_minor_radius] ...
    eigenvalues            (kx, ky, time) complex128 120kB [vref_most_probable/lref_minor_radius] ...
    eigenfunctions         (field, theta, kx, ky, time) complex128 9MB [] (0....
    growth_rate_tolerance  (kx, ky) float64 120B [] 0.02105 ... 9.304e-06
Attributes: (12/13)
    linear:            True
    gk_code:           GS2
    input_file:        \n\n\n\n&kt_grids_knobs\ngrid_option = 'range'\n/\n\n&...
    attribute_units:   {}
    title:             GKOutput
    software_name:     Pyrokinetics
    ...                ...
    object_type:       GKOutput
    object_uuid:       e421807e-5d8e-4cfa-8512-758c0324db93
    object_created:    2025-01-22 11:54:56.589850
    session_uuid:      5fb02907-87de-4cc4-8c15-9795abaac741
    session_started:   2025-01-22 11:54:50.631319
    netcdf4_version:   1.6.5
```

Any thoughts on this?

### DIFF
--- a/src/pyrokinetics/gk_code/gs2.py
+++ b/src/pyrokinetics/gk_code/gs2.py
@@ -624,7 +624,6 @@ class GKInputGS2(GKInput, FileReader, file_type="GS2", reads=GKInput):
             and self.data["dist_fn_knobs"]["adiabatic_option"]
             in adiabatic_electron_flags
         ):
-            adiabatic_electron = True
             found_electron = True
             nspec = self.data["species_knobs"]["nspec"] + 1
             species_key = f"species_parameters_{nspec}"
--- a/src/pyrokinetics/gk_code/gs2.py
+++ b/src/pyrokinetics/gk_code/gs2.py
@@ -12,7 +12,7 @@ import pint
 from cleverdict import CleverDict
 from scipy.integrate import trapezoid
 
-from ..constants import pi
+from ..constants import deuterium_mass, electron_mass, pi
 from ..file_utils import FileReader
 from ..local_geometry import LocalGeometry, LocalGeometryMiller, default_miller_inputs
 from ..local_species import LocalSpecies
@@ -625,12 +625,16 @@ class GKInputGS2(GKInput, FileReader, file_type="GS2", reads=GKInput):
             in adiabatic_electron_flags
         ):
             found_electron = True
-            nspec = self.data["species_knobs"]["nspec"] + 1
-            species_key = f"species_parameters_{nspec}"
 
-            electron_density = self.data[species_key]["dens"]
-            electron_temperature = self.data[species_key]["temp"]
-            e_mass = self.data[species_key]["mass"]
+            # Set from quasi-neutrality
+            qn_dens = 0.0
+            for i_sp in range(self.data["species_knobs"]["nspec"]):
+                species_key = f"species_parameters_{i_sp + 1}"
+                qn_dens += self.data[species_key]["dens"] * self.data[species_key]["z"]
+
+            electron_density = qn_dens
+            electron_temperature = 1.0 / self.data["knobs"].get("tite", 1.0)
+            e_mass = (electron_mass / deuterium_mass).m
 
         rgeo_rmaj = (
             self.data["theta_grid_parameters"]["r_geo"]
--- a/src/pyrokinetics/gk_code/gs2.py
+++ b/src/pyrokinetics/gk_code/gs2.py
@@ -635,6 +635,13 @@ class GKInputGS2(GKInput, FileReader, file_type="GS2", reads=GKInput):
             electron_density = qn_dens
             electron_temperature = 1.0 / self.data["knobs"].get("tite", 1.0)
             e_mass = (electron_mass / deuterium_mass).m
+            n_species = self.data["species_knobs"]["nspec"]
+            electron_index = n_species + 1
+
+            if np.isclose(electron_density, 1.0):
+                reference_density_index.append(n_species + 1)
+            if np.isclose(electron_temperature, 1.0):
+                reference_temperature_index.append(n_species + 1)
 
         rgeo_rmaj = (
             self.data["theta_grid_parameters"]["r_geo"]
--- a/src/pyrokinetics/gk_code/gs2.py
+++ b/src/pyrokinetics/gk_code/gs2.py
@@ -617,6 +617,22 @@ class GKInputGS2(GKInput, FileReader, file_type="GS2", reads=GKInput):
             temperatures.append(temp)
             masses.append(mass)
 
+        adiabatic_electron_flags = ["iphi00=2", "field-line-average-term"]
+
+        if (
+            not found_electron
+            and self.data["dist_fn_knobs"]["adiabatic_option"]
+            in adiabatic_electron_flags
+        ):
+            adiabatic_electron = True
+            found_electron = True
+            nspec = self.data["species_knobs"]["nspec"] + 1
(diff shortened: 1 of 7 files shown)


bpatel2107:
@physycola FYI

src/pyrokinetics/gk_code/gs2.py
```diff
@@ -617,6 +617,21 @@ def _detect_normalisation(self):
             temperatures.append(temp)
             masses.append(mass)
 
+        adiabatic_electron_flags = ["iphi00=2", "field-line-average-term"]
```
d7919, line 620: Whilst you're touching this part of the code I have some comments on the earlier if (line 604) which is trying to find the electron. Firstly can't electron_index and the reference indices just use `isp` rather than taking the `len` of  existing lists (we always add to these lists so the len at this point is always isp I think). Secondly, strictly GS2 users can choose a different reference charge if they really wanted to cause confusion. One could try to inspect the species' `type` inputs _or_ perhaps check that A) there are species opposite signs of the charge B) which species has the smallest mass. None of this is entirely general so perhaps fine to stick with what is there already.

src/pyrokinetics/gk_code/gs2.py
```diff
@@ -617,6 +617,21 @@ def _detect_normalisation(self):
...
+
+        if (
+            not found_electron
+            and self.data["dist_fn_knobs"]["adiabatic_option"]
+            in adiabatic_electron_flags
+        ):
+            found_electron = True
+            nspec = self.data["species_knobs"]["nspec"] + 1
+            species_key = f"species_parameters_{nspec}"
```
d7919: Just for parity with the earlier loop

```suggestion
            isp = self.data["species_knobs"]["nspec"] + 1
            species_key = f"species_parameters_{isp + 1}"
```

src/pyrokinetics/gk_code/gs2.py
```diff
@@ -617,6 +617,21 @@ def _detect_normalisation(self):
...
+
+            electron_density = self.data[species_key]["dens"]
```
d7919: The input file likely won't have a species_parameters namelist for the adiabatic species. Presumably we could get the density by enforcing quasi-neutrality. The temperature maybe from `tite` inputs? The mass just from scipy constants with normalisation applied?

physycola: Given that electrons will be adiabatic in these cases, I think it's probably safe to assume quasi-neutrality. Temperature could default to being the same as the ions unless `tite` is specified. And the mass assuming that the ions are deuterium would probably be a good default. 

Alternatively, you could demand that users need to specify a namelist for the electrons even if they are taken to be adiabatic to populate these values? Would probably be easier than the above modifications?

bpatel2107: Using `tite` with a default of 1 would be easy enough and I guess the mass doesn't really matter here so default to the scipy value. would be. Density from quasi-neutrality would also be straightforward too

src/pyrokinetics/gk_code/gs2.py
bpatel2107: This is now no longer relevant

src/pyrokinetics/gk_code/gs2.py
bpatel2107: Used `tite` and quasi-neutrality for GS2. Stella has something called `nine` which I used here.

src/pyrokinetics/gk_code/gs2.py
bpatel2107: 1) I think we can use `i_sp` so I can add that in
2) I guess they could and I would judge them for doing so. We could look at the type or mass but that could get complicated quickly. Not sure its worth handling unless we come across it.

bpatel2107: Fixed 1)

src/pyrokinetics/gk_code/gs2.py
bpatel2107: Now we allow for the electron to be the reference species for temp/density.

bpatel2107:
I think I'm happy with this now! I now have included the electron as a potential reference species (when things are equal to 1) meaning you can just do

```python
pyro.load_gk_output()
```

Without having to stick to a specific convention. We don't allow for automatic conversion between units when `Ti!=Te` as that would need more care. Currently we have something like `LocalSpecies`  which made up of all the kinetic `Species`, so adiabatic are not yet added here. We would need to add the adiabatic species to `LocalSpecies` which would then eventually allow for automatic conversions. I would want to do this carefully as I think any adiabatic `Species` object within `LocalSpecies` would need to be labelled as such. I.e have an attribute labelling it as much. To be honest I think a good way to split it up would be to have a property like `local_species.species_type` which could be `adiabatic`, `thermal` or `fast` as fast ion should be treated slightly differently too.

PR #407, opened 2025-01-24
Account for correct k_perp calc

bpatel2107:
Adding the final point to the `k_perp` calculation meant we needed to do that for the geometric quantities too.

Also included lambda is the output for those budding scientists!


mgiacomin:
It is working now, thanks!

mgiacomin (approved):

PR #408, opened 2025-02-04
Add option for kwargs in load_local and test

bpatel2107:
It was not possible to add kwargs for `load_local_geometry` and `load_local_species` to `pyro.load_local` before so we add that in now and include a test for this.

For example you can now do

```python
pyro.load_local(psi_n=psi_surf, 
                local_geometry='MXH',
                local_geometry_kwargs = {"n_moments": 6},
                local_species_kwargs = {"a_minor": 1.0 * pyro.norms.units.meter},)
```

which will get passed along appropriately.

### DIFF
--- a/src/pyrokinetics/pyro.py
+++ b/src/pyrokinetics/pyro.py
@@ -1771,8 +1771,8 @@ class Pyro:
         psi_n: float,
         local_geometry: str = "Miller",
         show_fit: bool = False,
-        local_geometry_kwargs={},
-        local_species_kwargs={},
+        local_geometry_kwargs: dict[str, Any] | None = None,
+        local_species_kwargs: dict[str, Any] | None = None,
     ) -> None:
         """
         Combines calls to ``load_local_geometry()`` and ``load_local_species()``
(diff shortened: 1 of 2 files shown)


LiamPattinson (approved):
Generally all looks good to me. The use of [mutable defaults](https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument) for the function arguments could be bug-prone, but for such a simple function it's unlikely to ever be a problem. I've left a suggestion in case you think it's worth patching up.

src/pyrokinetics/pyro.py
```diff
@@ -1767,7 +1767,12 @@ def load_local_species(self, psi_n: float, a_minor: Optional[float] = None) -> N
         self.local_species = local_species
 
     def load_local(
-        self, psi_n: float, local_geometry: str = "Miller", show_fit: bool = False
+        self,
+        psi_n: float,
+        local_geometry: str = "Miller",
+        show_fit: bool = False,
+        local_geometry_kwargs={},
+        local_species_kwargs={},
```
LiamPattinson: I didn't click through to submit the suggestion first time around...

```suggestion
        local_geometry_kwargs: dict[str, Any] | None = None,
        local_species_kwargs: dict[str, Any] | None = None,
```

And at the start of the function:

```python
if local_geometry_kwargs is None:
    local_geometry_kwargs = {}
# etc...
```

PR #409, opened 2025-02-06
Bugfix in gs2.py for normalisations value (qref-->zref)

d7919:
The GS2 input for the reference charge is zref.

### DIFF
--- a/src/pyrokinetics/gk_code/gs2.py
+++ b/src/pyrokinetics/gk_code/gs2.py
@@ -893,7 +893,9 @@ class GKInputGS2(GKInput, FileReader, file_type="GS2", reads=GKInput):
             self.data["normalisations_knobs"]["vref"] = (1 * convention.vref).to(
                 "meter/second"
             )
-            self.data["normalisations_knobs"]["zref"] = 1 * convention.qref
+            self.data["normalisations_knobs"]["zref"] = (1 * convention.qref).to(
+                "coulomb"
+            )
             self.data["normalisations_knobs"]["rhoref"] = (1 * convention.rhoref).to(
                 "meter"
             )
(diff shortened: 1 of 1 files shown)


src/pyrokinetics/gk_code/gs2.py
```diff
@@ -893,7 +893,7 @@ def set(
             self.data["normalisations_knobs"]["vref"] = (1 * convention.vref).to(
                 "meter/second"
             )
-            self.data["normalisations_knobs"]["qref"] = 1 * convention.qref
```
d7919, line 896: Should we also convert to Coulombs as we do for all other normalisations?

bpatel2107: If it should be then yes. I'm not sure what GS2 expects so it should be whatever that is

bpatel2107 (approved):
Looks good though if this should be in Coulombs please change.

Should we add something here for `zref`
https://github.com/pyro-kinetics/pyrokinetics/pull/409/files#diff-9cc4e33fc5abaf2716ce021a7408b0109f6e312ce3417903d8137cead938ec8cR508

src/pyrokinetics/gk_code/gs2.py
d7919: GS2 doesn't do anything with this figure other than writing it out to its netcdf output -- it's primarily to make it easier to keep track of the reference values used in generating the input file.

d7919:
> Looks good though if this should be in Coulombs please change.
> 
> Should we add something here for `zref` https://github.com/pyro-kinetics/pyrokinetics/pull/409/files#diff-9cc4e33fc5abaf2716ce021a7408b0109f6e312ce3417903d8137cead938ec8cR508

We could possibly do something here, but I'm not sure pyro supports anything other than qe for the reference charge?

src/pyrokinetics/gk_code/gs2.py
bpatel2107: I guess then to be consistent with the other references we should put it in Coulombs

PR #410, opened 2025-02-06
Reset local_geometry/species record after each load_local

bpatel2107:
Previously doing multiple `pyro.load_local` would work but would not reset the record of `local_species` and `local_geometry` meaning that if you then modify the object somehow (say a `merge_species`) due to a older record existing that change may not take when actually writing a file.  This PR fixes that by resetting the record for each call of `load_local` and adds a test.

### DIFF
--- a/src/pyrokinetics/pyro.py
+++ b/src/pyrokinetics/pyro.py
@@ -416,7 +416,6 @@ class Pyro:
         # Check that the provided gk_code is valid
         if gk_code is not None and gk_code not in self.supported_gk_inputs:
             raise ValueError(f"The gyrokinetics code '{gk_code}' is not supported")
-
         # If we've already seen this context before, or the new context is None, change
         # context and return.
         if gk_code is None or (
@@ -1211,9 +1210,9 @@ class Pyro:
                 "eigenvalues" in self.gk_output
                 and "time" in self.gk_output["eigenvalues"].dims
             ):
-                self.gk_output.data["growth_rate_tolerance"] = (
-                    self.gk_output.get_growth_rate_tolerance()
-                )
+                self.gk_output.data[
+                    "growth_rate_tolerance"
+                ] = self.gk_output.get_growth_rate_tolerance()
 
     # ==================================
     # Set properties for file attributes
@@ -1800,6 +1799,10 @@ class Pyro:
         Exception
             See exceptions for ``load_local_geometry()`` and ``load_local_species()``.
         """
+        if self._local_geometry_record:
+            self._local_geometry_record = {}
+        if self._local_species_record:
+            self._local_species_record = {}
 
         if local_geometry_kwargs is None:
             local_geometry_kwargs = {}
--- a/src/pyrokinetics/pyro.py
+++ b/src/pyrokinetics/pyro.py
@@ -1210,9 +1210,9 @@ class Pyro:
                 "eigenvalues" in self.gk_output
                 and "time" in self.gk_output["eigenvalues"].dims
             ):
-                self.gk_output.data[
-                    "growth_rate_tolerance"
-                ] = self.gk_output.get_growth_rate_tolerance()
+                self.gk_output.data["growth_rate_tolerance"] = (
+                    self.gk_output.get_growth_rate_tolerance()
+                )
 
     # ==================================
     # Set properties for file attributes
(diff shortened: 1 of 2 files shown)


src/pyrokinetics/pyro.py
```diff
@@ -1800,6 +1799,10 @@ def load_local(
         Exception
             See exceptions for ``load_local_geometry()`` and ``load_local_species()``.
         """
+        if self._local_geometry_record:
```
d7919, line 1802: I note that there's `force_overwrite` as an option to `_switch_gk_context` which sounds like it might be the sort of thing we would want here?

This is probably over complicated but could one keep the caching behaviour if the key was some combination of `gk_code`, `psin` and `local_geometry`? (e.g. a tuple of these or alternative nested dictionaries)?

Currently this wipes the entire cache, i.e. even for other gk_codes than the one currently active. Is that intentional (e.g. to force consistency) or just a by-product?

bpatel2107: This is a little brute force. The real issue I guess is as you say here

https://github.com/pyro-kinetics/pyrokinetics/blob/a2a962ea788a3999a3eb236a5f542e797ca47ef2/src/pyrokinetics/pyro.py#L419-L427

Maybe we can be a little more subtle.

Essentially we are in the position where

1. We have modified `local_species` and **then** set `gk_code` during the `write_gk_code`. This works the first time you write to file as when you try and switch from `gk_code = None` to `gk_code = GS2` there is nothing in `local_species_record` so we read in data from `_local_from_global` which has been modified by our merge. This also adds a `LocalSpecies` object to our record with `gk_code` as the key. 

https://github.com/pyro-kinetics/pyrokinetics/blob/a2a962ea788a3999a3eb236a5f542e797ca47ef2/src/pyrokinetics/pyro.py#L1422-L1428

2. But at the end of our `write_gk_file` we revert our `gk_code` back to `None`. The reason this happens is because before writing we originally had `gk_code = None` set, so later on it tries to switch back to `None`. This then makes us go through a `switch_gk_context` then go through the flag above which sets `gk_code` to `None`


https://github.com/pyro-kinetics/pyrokinetics/blob/a2a962ea788a3999a3eb236a5f542e797ca47ef2/src/pyrokinetics/pyro.py#L1104-L1105


3. However when you load, merge and then write the second time, when we first try and switch context we end up  using the `local_species` in the record as that dictionary now exists and is not modified when we do our merge.


So we have a few options I think

1. Delete the record as I have done for each `load_local` (easy)
2. Don't revert to `gk_code = None` at the end of the `write`. Will work in this instance as we won't go through a `_switch_gk_context` when writing. This means that when the 2nd `load_local` is done, we overwrite the entry in `_local_species_record` rather than writing to `_local_species_from_global` (easy but not sure if correct either). Once a `gk_code` is set from `None` should we ever revert back?
3. Modify the key to the records somehow to account for the different `load_local` Not sure how to do this. Could utilise the naming function we have in `pyro` and come up with a unique name each time we do a `load_local`?


The 3rd one is not impossible though a bit fiddily. Thoughts?

bpatel2107: So I had a go at making a bespoke tag each time a `load_local_geometry` or `load_local_species` gets called which works fine. Essentially the code will check the cache if an object already exists with the current key (currently `GKCODE_NAME` so `GS2_run0000` for example). If something already exists then it will make a new key and save it to that key.

The issue is that we use that key for all the other objects like `Numerics`, `GKInput` etc. So if we load a new `LocalGeometry` or `LocalSpecies`, that key will be updated and we won't be able to access the previously cached object with the original key. One could rename/copy all the other objects to use the new key but that doesn't feel good.

Here's a rough workflow as is

1. Create a pyro object with an original `LocalSpecies`, `LocalGeometry` and `Numerics`, `GKInput` mapped to `key1`
2. Load a new surface using `load_local` which creates a new `LocalSpecies`, `LocalGeometry` with `key2` 
3. When trying to access `Numerics`/`GKInput` we have updated the key to `key2` so we can't access the original objects (with `key1`) so we return a `None`
4. Everything falls off a cliff

Similarly this would happen if you just did `load_local_geometry` or `load_local_species` but in that case we'd have another object that couldn't be accessed.

I don't really see a nice way around this to be honest. I don't want bespoke keys for each object and its hard to say when we should automatically update keys for other objects. Thoughts here would be appreciated.

d7919: Yes it looks like it would be quite tricky to get something consistent and robust. Better to be robust (and correct) at the expense of resetting the cache.

bpatel2107: I think I agree, we can always revisit this if it becomes an issue.

PR #412, opened 2025-02-19
Bugfix/adiabatic electron cgyro local species

bpatel2107:
Fixes issue with adiabatic electron runs in CGYRO failing during `get_local_species` as the "electron's" `Te`, `ne` and `nu_e` are needed to get the collisionality for other species.

Also fixes an separate issue when loading box style runs without the nonlinear term (`N_TOROIDAL >1` and `NONLINEAR=0`) and when `THETA_PLOT!=N_THETA`


physycola:
Just to clarify. Is the intended behaviour that it will override the `THETA_PLOT` assignment?

bpatel2107:
This doesn't change the `THETA_PLOT` assignment, but just makes the `theta` variable a numpy array as there was an issue with the line later on with `theta = theta + theta0` not liking a mix of a list and float.

PR #413, opened 2025-02-19
Enhancements/stella update

physycola:
This PR bring brings stella.py up-to-date with stella at commit #7131981b62ea74ca9a4187ab197feb2a6b90a0ae on 20/02/25. 

The current stella.py script does not properly handle writing a new input file from an old one with multiple linear modes specifiied in the "range" grids layout, but without changing parts of pyro outside of stella.py, this cannot be resolved. Note that this issue is also the same for GS2.


bpatel2107:
Looks like the failing test is just from the new files in the templates. We store a netcdf file here

https://github.com/pyro-kinetics/pyrokinetics/blob/e6a304863fa7614733fab8b250d3b58cc0c79c20/tests/gk_code/golden_answers/stella_linear_output_cecf2584.netcdf4

That has the data stored and we check that the template data read in by pyro matches the data in these netCDF files. Given that you have changed the template its not surprising this is failing. We can either generate a new netCDF or rerun the same case as above. Probably makes more sense to make a new netCDF file and its very easy to do. I can push that to get the tests to pass.

A question I have is whether we want to preserve the ability to read in older STELLA runs as that seems useful. Looking at the changes I think it would be pretty straightforward to do. Something like

```python
self.stella_legacy = False
self.parameter_key = "parameters_physics"
self.numerical_key = "parameters_numerical"
if "knobs" in self.data.keys() and "physics_flags" in self.data.keys():
    self.stella_legacy = True
    self.parameter_key = "physics_flags"
    self.numerical_key = "knobs"
    
...

is_nonlinear = self.data[self.parameter_key]["nonlinear"]
```

This would let us preserve the ability to read in older runs. Let me know your thoughts.

src/pyrokinetics/gk_code/stella.py
```diff
@@ -510,16 +520,21 @@ def _detect_normalisation(self):
             temperatures.append(temp)
             masses.append(mass)
 
+        print("AAAAAAAAAAAAAAAAAAAAAAAAA")
+        print(found_electron)
+
```
bpatel2107:
```suggestion
```

Although informative we can probably remove these lines

src/pyrokinetics/gk_code/stella.py
```diff
@@ -529,6 +544,8 @@ def _detect_normalisation(self):
             if np.isclose(electron_temperature, 1.0):
                 reference_temperature_index.append(n_species + 1)
 
+        print(found_electron)
+
```
bpatel2107:
```suggestion
```

bpatel2107:
I was able to generate a new set of golden answers that almost passed all the tests but the netCDF output you have uploaded does not contain `phi(kx, ky, theta, time)` so I can't load in the fields. This causes the tests `test_amplitude` and `test_data_vars[phi]` to fail which both require `phi`

@physycola can you make run with all this information included in the output. To reduce the size of the file we can run it for fewer timesteps say 10. The point of the test is just to ensure that the data is being read correct, not to worry about the the convergence of that run.

physycola:
@bpatel2107 , I have uploaded the new data in both /src/pyrokinetics/templates/outputs/STELLA_linear and /tests/gk_code/golden_answers. These are from the new template file. The tests still fail, but I cannot figure out why. 

With respect to your suggestion for preserving backwards compatability, I think it will become too complicated to do this within pyro, as more may change. The stella team are writing a tool that can update old input files into the new format, so I think it's best we leave the backwards compatability to that.

bpatel2107 (commented):
I think this is goof to go, still feel like it is worth keeping backwards compatibility as 

1) I think it is quite easy to do
2) There are likely a lot of STELLA runs that already exist and dropping support for them seems like not a great idea
3) I hoping then input file structure is stable now as that is where most of the changes are
4) The only change to the output is the name of `omega` which again is easy to check

If the STELLA developers feel very strongly about dropping support for older input files then I think at least a warning should be raised when reading older input files telling the user to update their STELLA input files to account for this. Updating their output files may be much harder so I think a check in the omega reader is best there

bpatel2107 (approved):
Happy to merge this now

PR #416, opened 2025-02-24
Addition of GX

bpatel2107:
This PR adds GX to pyrokinetics!

It includes both the `GKInputReader` and `GKOutputReader` classes along with their appropriate tests. We have also included GX in several roundtrip tests. Given that GX only does box style runs we ended up finding a few bugs in that style of run in other codes so that is included here.

Big thanks to @physycola for doing the bulk of the work here!


bpatel2107:
So there are a few issues related to reading/writing adiabatic species here but I think that is a separate issue so I would like to get this merged and treat that as a separate PR if possible. Having a go at fixing it and its getting a little out of hand so deserves some proper treatment I think.

physycola:
Happy to treat the adiabatic species handling as a separate bugfix PR, given the amount of code-coverage required.

physycola (approved):
Happy for this to be merged. As discussed, it has introduced some bugs to do with handling adiabatic species, but this occurs across the code-base as a whole, and so should be handled as a separate PR.

bpatel2107:
Forgot the most important change!

PR #418, opened 2025-03-04
Allow for case without theta0 in MG sat and add legacy flux defn in GS2

bpatel2107:
Allow for applications of MG saturation rule without any theta0 dependence.


src/pyrokinetics/gk_code/gs2.py
mgiacomin, line 1: It looks fine to me, although these old diagnostics have been deprecated since a while. It might be okay for backward compatibility, but I let @d7919 to check.

mgiacomin (approved):
It looks good to me.

src/pyrokinetics/gk_code/gs2.py
d7919: It looks OK as a way to support the old diagnostics in case a user is using an old exe or has turned off the new diagnostics.

d7919 (approved):

PR #426, opened 2025-04-11
Bugfix/gene mxh map

### LINKED ISSUE #425: New issues with GENE tracer_efit geometry
Hi all, 
Encountering issues with datasets with geometry `tracer_efit` (These worked fine before recent merges)

I get the following error (accompanied by a few warnings):

```
/pyrokinetics/src/pyrokinetics/gk_code/gene.py:401: UserWarning: GENE dpdx_pm = -1.1908992 not set consistently with amhd = -0.0 - drifts may not behave as expected
  warnings.warn(
/pyrokinetics/src/pyrokinetics/pyro.py:1609: UserWarning: Explicitly set value of beta (0.02973882163703989 beta_ref_ee_B0) is inconsistent with value from physical reference values (0.029714536572720735 beta_ref_ee_B0)
  warnings.warn(

Cannot convert from 'tref_electron_parameters_00010000 * rhoref_custom_parameters_00010000 / elementary_charge / lref_magnetic_axis_parameters_00010000' ([mass] * [length] * [rhoref] / [current] / [time] ** 3) to 'tref_electron_parameters_00010000 * rhoref_pyro_parameters_00010000 / elementary_charge / lref_minor_radius_parameters_00010000' ([mass] * [length] ** 2 / [current] / [time] ** 3)

```

Have some test data [here](https://github.com/Sapientai/MGKDB/tree/develop/test_data/test_gene1_tracer_efit)

bpatel2107:
This fixes a few small bugs found in the GENE <-> IMAS mapping

1.  The sign of $I_p$ and $B_t$ are important when taking the metric terms from GENE and fitting $B_\theta$ so I added some signs when reading them in
2. When mapping to MXH we needed to be careful about the direction of $\theta$ and the fitting should span from $0 \rightarrow 2\pi$ so I added som interpolation around the end. Also missing a factor $0.5$ which only really mattered then the plasma was tilted. Also we are more careful/explicit about certain defn which can vary between `LocalGeometry` definitions.
~~3. GENE was not checking the default output convention like the other code so was assuming it was in the `pyrokinetics` or `gene` convention (note `gene` convention here does not mean it matches `tracer_efit`, rather that $L_ref=R_{major}$). So now we check the norm convention from the input like the other codes.~~ This was actually fixed in #364 
4. Finally when we define `rhoref_custom_physical` (which is needed for `tracer_efit` GENE runs) we were using simulation units (abstract) rather than physical units (which can be converted into SI). That has now been fixed.

This fixes #425 ~~and actually highlights a hidden bug as those run should have been loading in using `gene_bespoke` units from the beginning but were actually using `gene`. So the mapping to IMAS should have been failing from day one given issue 4.~~


bpatel2107:
Context for the $B_\theta$ defn before and after the fix:

Wrong defn of $g_{xy}$, $g_{yz}$
image

Correct defn of $g_{xy}$, $g_{yz}$
image

venkiteshayyar:
Tested it on datasets. More success, but a few runs still fail.

Here is a sample dataset that fails with the same error: 

`Exception: Least squares fitting in MXH::_set_shape_coefficients failed with message : The maximum number of function evaluations is exceeded.`



[sample_test_failure_gene.tar.gz](https://github.com/user-attachments/files/19710974/sample_test_failure_gene.tar.gz)

bpatel2107:
Easy mistake and fix. I was rotating my `R`,`Z` array to the point closest to `Z0` which for this case happens to be on the inboard side which will break this. I need the closest point on the outboard side. Pushed a fix for that

venkiteshayyar:
Alright, with these changes, the previous errors go away. 
But, I noticed an old error for a few GENE runs. 

Runs with a `rhostar` value in the parameters file like, `rhostar =  0.26326959E-02` had no problems 

However, for runs without such a line in the parameters file, the rho value is set to 0 by default and this causes an error here due to division by zero: 
``` return integral * self.Rmaj / (2 * pi * self.rho)```
https://github.com/pyro-kinetics/pyrokinetics/blob/unstable/src/pyrokinetics/local_geometry/local_geometry.py#L573

<strike> It's not clear where the value is being set to 0 either </strike> .Attaching call stack below.
Ok, I can see that it is set here : https://github.com/pyro-kinetics/pyrokinetics/blob/unstable/src/pyrokinetics/local_geometry/local_geometry.py#L110


Is it possible to set the default value of `rho` to something besides 0 or ensure that the input file contains a value for `rhostar` ?

I'm attaching the parameters file, just in case 
[parameters_0014.csv](https://github.com/user-attachments/files/19735534/parameters_0014.csv)
(remove suffix .csv after download) 

Call stack

bpatel2107:
Okay so that's progress! I would need the corresponding `tracer_efit_0014` file to check this as pyro is trying to read in that file to determine `rho` (`rhostar` isn't used here for the `LocalGeometry`. Can you upload that (also check that the file does exist in the correct directory)?

venkiteshayyar:
So, in this case, the geometry files were absent. 
This is more of an issue with data preparation that pyro or mgkdb.
All other runs work!

bpatel2107:
Added an error if the geometry file is not found

PR #427, opened 2025-04-14
Force the sign of ky to be positive when loading gk_output

physycola:
I thought it would be useful for various cases/parts of the code for the sign of `ky` to be forced to be positive when it happens to be negative in `CGYRO`. This does not cause any loss of information, given that it can be reconstructed from various inputs/outputs already saved by `CGRYO` and `pyro`.

### DIFF
--- a/src/pyrokinetics/gk_code/cgyro.py
+++ b/src/pyrokinetics/gk_code/cgyro.py
@@ -1255,6 +1255,7 @@ class GKOutputReaderCGYRO(FileReader, file_type="CGYRO", reads=GKOutput):
         pos += ntheta_ballooning
 
         ky = grid_data[pos : pos + nky] / bunit_over_b0
+        ky = np.abs(ky) # Force the sign of ky to be positive
 
         if gk_input.is_linear() and nky == 1:
             # Convert to ballooning co-ordinate so only 1 kx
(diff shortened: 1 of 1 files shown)


src/pyrokinetics/gk_code/cgyro.py
```diff
@@ -1255,6 +1255,7 @@ def _get_coords(
         pos += ntheta_ballooning
 
         ky = grid_data[pos : pos + nky] / bunit_over_b0
+        ky = np.abs(ky)  # Force the sign of ky to be positive
```
bpatel2107: Could just put the abs around the full statement so we have

`ky = np.abs(grid_data[pos : pos + nky] / bunit_over_b0)`

bpatel2107 (approved):
I think this is a good idea. Could do it without having another line but that is me being nitpicky

src/pyrokinetics/gk_code/cgyro.py
physycola: Yeah I wanted to give the user an opportunity to uncomment the second line if they wanted it for some unforseen purpose. Happy to put it into one line though.

bpatel2107: Nah they can just deal with it I think. Don't give them the option I think, they can change it in post as we shouldn't really offer changes in the source code

PR #428, opened 2025-04-15
Stella namelist backward compatibility

bpatel2107:
Stella recently changed (is changing?) its input file but given that many older runs still exists it seems appropriate to support this reading of older runs and raise a warning that we have a legacy file.

At the moment this just checks if the file is a legacy version or not and sticks with the relevant namelists but I guess we can leave it to the STELLA team to decide on how they want to handle the writing.

@HanneThienpondt FYI, couldn't get you as a reviewer, probably need to add you to the repo

### DIFF
--- a/src/pyrokinetics/gk_code/stella.py
+++ b/src/pyrokinetics/gk_code/stella.py
@@ -35,6 +35,10 @@ class GKInputSTELLA(GKInput, FileReader, file_type="STELLA", reads=GKInput):
     code_name = "STELLA"
     default_file_name = "input.in"
     norm_convention = "stella"
+    _parameters_physics = "parameters_physics"
+    _parameters_params = "parameters"
+    _parameters_numerical = "parameters_numerical"
+    _legacy_stella = False
 
     pyro_stella_miller = {
         "rho": ["millergeo_parameters", "rhoc"],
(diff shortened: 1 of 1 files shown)


physycola:
Having run this through all of my recent scripts (which use the new `STELLA` namelists), I can't see any obvious error with the implementation. I agree that maintenance of compatibility should be the responsability of the `STELLA` team; whether they want to handle this directly in `pyro` or do it on the `STELLA` end is up to them.

physycola (commented):
Having run this through all of my recent scripts (which use the new `STELLA` namelists), I can't see any obvious error with the implementation. I agree that maintenance of compatibility should be the responsability of the `STELLA` team; whether they want to handle this directly in `pyro` or do it on the `STELLA` end is up to them.

ZedThree:
One option we've talked about before is maintaining separate versions (with a common base implementation), `stella-0.5`/`stella-0.6` say, with `stella` defaulting to the latest version. This could also help cross-version comparisons. GS2 is also changing input parameters, so it would be good to get some consensus on how to deal with this in pyro.

bpatel2107:
Yes there does need to be a better way of handling this. I don't want to have a lot of repeated code here so if there is a way to have versions without having lots of copies for this then I am open to suggestions here.

What I've done here is just have a binary legacy flag which sets up certain namelist variables. We could switch that to an actual version number to track progress?

bpatel2107:
This sounds like a separate issue however so I may raise a bespoke issue here.

ZedThree:
I have some thoughts, but yes best for a separate issue!

PR #430, opened 2025-04-16
CGYRO signs of Bt/Ip/q

bpatel2107:
The ballooning transform in CGYRO was not being done correctly when q is negative leading to jumps in the eigenfunctions. You need to be careful when rearranging from a flux tube to ballooning space here which we were not doing. All the eigenvalues match here (signs of ion direction/electron direction can be important but we force ion = positive here)

When in doubt, do every permutation possible...

Here is a case at $\theta_0 = 0$ (KBM). You can see (sort of as they overlap a lot( that when $I_p$ flips sign, we essentially flip the eigenfunctions along $\theta$ (much easier to see in the odd parity eigenfunctions).
image
image

image
image

Here is a case with $\theta_0 = \pi$ (MTM)
image
image

image
image

### DIFF
--- a/src/pyrokinetics/gk_code/cgyro.py
+++ b/src/pyrokinetics/gk_code/cgyro.py
@@ -1255,7 +1255,7 @@ class GKOutputReaderCGYRO(FileReader, file_type="CGYRO", reads=GKOutput):
         pos += ntheta_ballooning
 
         ky = grid_data[pos : pos + nky] / bunit_over_b0
-        ip_ccw = gk_input.data.get("IPCCW", -1.0)
+
         if gk_input.is_linear() and nky == 1:
             # Convert to ballooning co-ordinate so only 1 kx
             theta = theta_ballooning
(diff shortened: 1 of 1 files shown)


bpatel2107:
CGYRO shifts the ballooning coordinate along by $\theta_0$ so you can see that it shifts the eigenfunctions so they are centered along $\theta_{ball}=0$ here.

I don't think any other code does this as it means $\theta$ here doesn't really corresponds to the geometric $\theta$ so I am not sure whether we should do this? @d7919 what does GS2 do here? 

Not massively important for eigenvalues but if we are doing things with Jacobians/B/other field varying terms we really ought to match the eigenfunctions to the right grid. Any thoughts on this?

d7919:
@bpatel2107 GS2 doesn't shift the grid in $\theta$ , although it is something that I've considered implementing previously . Because the parallel grid is connected to the trapped pitch angle handling for bounce points and some of the implicit assumptions in the code this is probably non-trivial for GS2.

bpatel2107:
Sorry to clarify CGYRO doesn't shift the angle internally in the code, it still uses the standard $\theta$. But when spitting out the "ballooning angle" it shifts $\theta$ by $\theta_0$

I'm leaning towards us not shifting the ballooning angle.

d7919:
Ah I see, so really just outputting a `theta_ball = theta - theta_0` type grid?

bpatel2107:
Will merge now

PR #433, opened 2025-04-22
GENE set beta_prime correctly if amhd=-1

bpatel2107:
When `amhd=-1` in GENE, $\beta'$ should be set consistently from beta/gradients but that wasn't being captured so that is fixed here.

Only relevant for reading in existing GENE runs where `amhd=-1` **and** `dpdx_pm=-2`.

Any files written by pyro from a global equilibrium, or any tracer_efit runs would have the correct $\beta'$


bpatel2107:
@dake0795  merging now btw

Issue #429, opened 2025-04-16
Handle different version of GK codes


bpatel2107:
As codes are developed input/output format can vary which can make their reading/writing a little complicated. We should have a clever way in which we can track the version of the code to determine its version and then set the appropriate values needed to read/write the data.

Related to #428 in which we do this for STELLA. Similar updates may be needed for GS2.

ZedThree:
It would also be nice to be able to select the version when writing input files. I can imagine a couple of different ways of doing it, either with a `version` keyword, or by parsing the code name.

PR #434, opened 2025-04-25
Bugfix/tglf pyroscan

fcasson:
Please merge this, Felix needs it.


src/pyrokinetics/gk_code/tglf.py
```diff
@@ -255,44 +255,43 @@ def get_local_geometry_miller(self) -> LocalGeometryMiller:
...
             self.data.get("p_prime_loc", 0.0)
-            * miller_turnbull_data["rho"]
-            / miller_turnbull_data["q"]
+            * mxh_data["rho"]
+            / mxh_data["q"]
             * (8 * np.pi)
         )
 
-        miller_turnbull = LocalGeometryMillerTurnbull.from_gk_data(miller_turnbull_data)
+        mxh = LocalGeometryMillerMXH.from_gk_data(mxh_data)
```
d7919: LocalGeometryMillerMXH --> LocalGeometryMXH

bpatel2107: How has this been running so far...

bpatel2107:
Okay so I've updated it to actually properly read/write MXH geometries which it wasn't doing before and I've added it to the roundtrip testing.

But there is a more fundamental issue here with TGLF (not really pyro specific so we can technically merge but this is relevant to #429).

The TGLF developers have modified the flux surface from `MillerTurnbull` to `MXH`. This means that the definition of `ZETA_LOC` and `S_ZETA_LOC` are now different so technically an old TGLF run will generate a different result if the sqaureness/derivative are non-zero.

In practice this is probably not a big deal as squareness tends to be small but the they are different now so integrated modelling codes that use squareness will be slightly different.

@hgd511 @tomneiser @ FYI

https://github.com/gafusion/gacode/commit/1d3daa68b5540ff6c83a58ca957291f283fc650e#diff-8ec26117cd5743de986a770e3e3a76fb0e394f3e7cbccac7a7460083bb8f16f1

bpatel2107 (approved):

PR #436, opened 2025-05-12
Enhancements/flux surface volumes

bpatel2107:
Add a method to `LocalGeometry` that calculates the toroidal and poloidal surface area as well as the volume enclosed by the flux surface using the analytical fits

$A_{toroidal} = \int_0^{2\pi}  R\frac{\partial L}{\partial \theta} d\theta$

where $\frac{\partial L}{\partial \theta} = \sqrt{\frac{\partial R}{\partial \theta} ^2 + \frac{\partial Z}{\partial \theta} ^2}$

$A_{poloidal} = \int_0^{2\pi}  R\frac{\partial Z}{\partial \theta} d\theta$

$V_{toroidal} = \pi \int_0^{2\pi}  R^2\frac{\partial Z}{\partial \theta} d\theta$

This also maintains units so we can convert to SI easily. Called by doing `area, surface, volume = local_geometry.get_flux_surface_area_volume`

These names here are a bit confusing I admit so happy to take suggestions

I've added a test using the template TRANSP netCDF which stores these as outputs.

### DIFF
--- a/src/pyrokinetics/local_geometry/local_geometry.py
+++ b/src/pyrokinetics/local_geometry/local_geometry.py
@@ -612,23 +612,29 @@ class LocalGeometry:
         Calculate the poloidal and toroidal area of the flux surface
         and the toroidal volume in units of lref
 
-        :math:`f = \frac{2\pi q}{\oint \frac{dl}{R^2 B_{\theta}}}`
+        :math:`A_{toroidal} = 2\pi \int_0^{2\pi}  R\frac{\partial L}{\partial \theta} d\theta`
 
-        See eqn 97 in Candy Plasma Phys. Control. Fusion 51 (2009) 105009
+        :math:`A_{poloidal} = \int_0^{2\pi}  R\frac{\partial Z}{\partial \theta} d\theta`
+
+        :math:`V_{toroidal} = \pi \ int_0^{2\pi}  R^2\frac{\partial Z}{\partial \theta} d\theta`
 
         Returns
         -------
-        f : Float
-            Prediction for :math:`f_\psi` from B_poloidal
+        poloidal_area : float, units [lref**2]
+            Calculation of the poloidal surface area  :math:`A_{poloidal}`
+        toroidal_area : float, units [lref**2]
+            Calculation of the toroidal surface area  :math:`A_{toroidal}`
+        toroidal_volume : float, units [lref**3]
+            Calculation of the poloidal volume :math:`V_{toroidal}`
         """
 
         lref = self.R.units
 
         # Calculate using Green's theorem
         def poloidal_surface_integrand(theta):
-            R, Z = self.get_flux_surface(theta)
+            R, _ = self.get_flux_surface(theta)
             (
-                dRdtheta,
+                _,
                 _,
                 dZdtheta,
                 _,
@@ -679,11 +685,11 @@ class LocalGeometry:
         def toroidal_volume_integral():
             return np.pi * quad(toroidal_volume_integrand, 0.0, 2 * np.pi)[0]
 
-        area = poloidal_surface_integral()
-        surface = toroidal_surface_integral()
-        volume = toroidal_volume_integral()
+        poloidal_area = poloidal_surface_integral()
+        toroidal_area = toroidal_surface_integral()
+        toroidal_volume = toroidal_volume_integral()
 
-        return area, surface, volume
+        return poloidal_area, toroidal_area, toroidal_volume
 
     def test_safety_factor(self):
         r"""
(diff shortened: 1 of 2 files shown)


src/pyrokinetics/local_geometry/local_geometry.py
```diff
@@ -607,6 +607,84 @@ def get_integral():
...
+
+        :math:`f = \frac{2\pi q}{\oint \frac{dl}{R^2 B_{\theta}}}`
+
+        See eqn 97 in Candy Plasma Phys. Control. Fusion 51 (2009) 105009
+
+        Returns
+        -------
+        f : Float
+            Prediction for :math:`f_\psi` from B_poloidal
+        """
```
hgd511: This looks like the descriptor of a different function

hgd511 (approved):
Looks good. 
There's a factor of 2 pi missing in your first PR comment for toroidal area, but it's in the code. Only thing to change is the function descriptor.

physycola:
Looks good from what I can tell. Is there a reason why the volume isn't being found by integrating the Jacobian directly? This should be the same, but may be easier to adapt if one ever changes the definitions of R and Z.

bpatel2107:
Yeah I realised that to get the volume via the Jacobian we need to do

$2\pi \int_0^{r_0} dr \int_0^{2\pi} d\theta  J(r,\theta)$

But we only have $J(r_0, \theta)$ on the local flux surface not $J(r,\theta)$ so we can't get the volume through the Jacobian locally.

src/pyrokinetics/local_geometry/local_geometry.py
bpatel2107: I've updated the docstring

physycola:
> Yeah I realised that to get the volume via the Jacobian we need to do
> 
> 2 π ∫ 0 r 0 d r ∫ 0 2 π d θ J ( r , θ )
> 
> But we only have J ( r 0 , θ ) on the local flux surface not J ( r , θ ) so we can't get the volume through the Jacobian locally.

Makes sense. Am I correct that this restriction will also mean that we cannot compute $dV/d\psi$ locally either? Would be another useful quantity to have for ballooning stability etc.

hgd511:
We should be able to evaluate it for a given surface $r_0$ via

$\frac{\partial V} {\partial r} \left(r = r_0 \right)= 2 \pi \int_{- \pi}^{\pi} \mathcal{J} \left(r_0, \theta \right) \mathrm{d} \theta$

where to make it $\partial V / \partial \psi$ we multiply by $1 / \left(\mathrm{d} \psi / \mathrm{d} r\right)$. 
Is this what you meant? Might have my wires crossed in terms of what constitutes computing locally.

physycola:
Good point @hgd511, my brain clearly wasn't working first thing in the morning. I think it would be useful to output at least $dV/dr$ at the flux surface alongside these quantities. If the user wants $dV/d\psi$, they can multiply by $(d\psi/dr)^{-1}$, as you suggest.

bpatel2107:
Happy to add this in. Do we want derivatives of all the areas too?

`get_flux_surface_area_volume_derivatives` method?

physycola:
I don't currently see a use-case for the derivatives of the other quantities, but there may be in the future. If it's straightforward to do for the areas without constructing derivatives that do not already exist, I'd say it would be worth including.

hgd511:
Differentiating the expressions in Bhavin's comment with respect to $r$, I'm getting

$\frac{\mathrm{d} A_{\mathrm{pol}}}{\mathrm{d} r} = \int_{-\pi}^{\pi} \frac{J}{R} \mathrm{d} \theta$

and (putting in the $2 \pi$)

$\frac{\mathrm{d} A_{\mathrm{tor}}}{\mathrm{d} r} = 2 \pi \int_{-\pi}^{\pi} \left( \frac{\partial R}{\partial r} \sqrt{g_{\theta \theta}} + \frac{R}{\sqrt{g_{\theta \theta}}} \left( \frac{\partial R}{\partial \theta} \frac{\partial^2 R}{\partial r \partial \theta} +  \frac{\partial Z}{\partial \theta} \frac{\partial^2 Z}{\partial r \partial \theta} \right) \right) \mathrm{d} \theta$

where $g_{\theta \theta} = \left(\partial R / \partial \theta \right)^2 + \left(\partial Z / \partial \theta \right)^2$, and we already have $\mathrm{d} V / \mathrm{d} r$.

I think Pyro should be calculating all of the quantities being integrated over already, so could be worth including.

bpatel2107:
Ive added the derivatives and some tests so will merge now

PR #438, opened 2025-05-22
Fix to setting of omprimfac in gs2.py

physycola:
`omprimfac` was being improperly set; should be default of `1.0` irrespective of the sign of the rotation, given that it is a muliplier that gets applied after the PVG term is calculated. @d7919 will correct me if I have read `GS2`'s source code wrong.

### DIFF
--- a/src/pyrokinetics/gk_code/gs2.py
+++ b/src/pyrokinetics/gk_code/gs2.py
@@ -275,7 +275,7 @@ class GKInputGS2(GKInput, FileReader, file_type="GS2", reads=GKInput):
 
             # Without PVG term in GS2, need to force to 0
             species_data.domega_drho = (
-                - self.data["dist_fn_knobs"].get("g_exb", 0.0)
+                -self.data["dist_fn_knobs"].get("g_exb", 0.0)
                 * self.data["dist_fn_knobs"].get("omprimfac", 1.0)
                 * self.data["theta_grid_parameters"]["qinp"]
                 / self.data["theta_grid_parameters"]["rhoc"]
@@ -789,7 +789,7 @@ class GKInputGS2(GKInput, FileReader, file_type="GS2", reads=GKInput):
         else:
             self.data["dist_fn_knobs"]["omprimfac"] = (
                 (
-                    - local_species.electron.domega_drho
+                    -local_species.electron.domega_drho
                     * local_geometry.rho
                     / local_geometry.q
                     / numerics.gamma_exb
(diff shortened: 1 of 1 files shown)


bpatel2107:
Ah okay I think I misunderstood your message before but there may be a sign error though @d7919 can check this.

So I think the original implementation was actually (mostly) correct. My understanding is that the PVG term is calculated from `g_exb` in the GS2 input file. If the PVG and ExB shearing terms are consistently set then this should be 1.

But if you have a case where they are different then you need to be able to change these independently which is what this code is doing.

In general pyro tries to use the values in the `Numerics` object to set  the ExB shear and the values in the `LocalSpecies` object to set the PVG terms.

To clarify


The ExB term is generally defined as 

$\gamma_{E\times B} = \frac{r}{q} \frac{\partial \omega}{\partial r}$

This is exactly `numerics.gamma_exb`


The PVG term is generally defined as 

$\gamma_P = R_{maj} \frac{\partial \omega}{\partial r}$

Which for us is `local_species.domega_drho`


There are many cases where you may want ExB shear but not the PVG term for example.

Essentially if you want to set the ExB shear you should change `numerics.gamma_exb`. 
If you want to set the PVG shear you should change `local_species.domega_drho`

physycola:
I understand the original implementation (and should definitely not have removed the functionality, whoops). What I do not understand, however, is what the correct implementation is. As it stands, the sign convention that passes the tests will result in the equivalent of `omprimfac = -1.0`, but this should not be the case if you are setting everything self-consistently. The use-case that I was considering was simply writing a `GS2` input file from a `TRANSP` one, which was setting `omprimfac = -1.0`, even though being "self-consistent" would suggest `omprimfac = 1.0`. I have reverted my changes to allow for more investigation on this branch.

bpatel2107:
So I think we are just missing a minus sign here. 

I was not careful before as the definitions should be 

$\gamma_{E\times B} = -\frac{r}{q} \frac{\partial \omega}{\partial r}$

$\gamma_P = -R_{maj} \frac{\partial \omega}{\partial r}$

The minus sign is in `numerics.gamma_exb` but is just missing from this. Probably should be


```
       if numerics.gamma_exb == 0.0:
            self.data["dist_fn_knobs"]["omprimfac"] = 1.0
        else:
            self.data["dist_fn_knobs"]["omprimfac"] = (
                (
                    - local_species.electron.domega_drho
                    * local_geometry.rho
                    / local_geometry.q
                    / numerics.gamma_exb
                )
                .to_base_units()
                .m
            )
```

physycola:
Such a change fails the `[gs2-cgyro]` and `[tglf-gs2]` due to a sign error (effectively the same test failures as simply setting `omprimfac = 1.0`), so I think it's not quite as simple as this?

bpatel2107:
This is a case of a two errors cancelling out, both in the getting and setting of it.

There should be a minus sign here too

https://github.com/pyro-kinetics/pyrokinetics/blob/aa3974aff6d73790db96a51ef2bc9c368016208c/src/pyrokinetics/gk_code/gs2.py#L276-L282

bpatel2107 (approved):
Looks good now! @d7919 any other thoughts?

d7919 (approved):

PR #439, opened 2025-05-29
Add in splines for psi_tor and rho_tor

bpatel2107:
Here we add in a calculate for the toroidal flux and the normalised root toroidal flux into the `Equilibrium` object

$\psi_{tor} = \int_{\psi_{axis}}^{\psi_{lcfs}} q d\psi$

With

$\rho_{tor} = \frac{\psi_{tor} }{\psi_{tor}^{lcfs}}$

They are added as Unit Splines so one can do

```python
psi_n = 0.5
psi_tor_mid = eq.psi_tor(psi_n)
rho_tor_mid = eq.rho_tor(psi_n)
```

Is this the best place/time to add in functionality where one can choose flux surface to load using either `rho` = $r/a$, `psi_n` = $\psi_n$ or `rho_tor` = $\rho_{tor}$ ? At the moment we can only load from `pyro.load_local(psi_n=0.5)` for example. 

Would it be useful to have `pyro.load_local(rho=0.5)` and `pyro.load_local(rho_tor=0.5)`


bpatel2107:
Will merge and close now. If a request comes in for the above functionality then we can add it in

Issue #335, opened 2024-05-10
Feature request: Clever Ion naming


bpatel2107:
We should mostly be able to determine which ion we are dealing with from the charge/mass alone. This get complicated with partial ionisation states but maybe we can handle this in a clever way. See conversation in #315 

@arkabokshi FYI

ZedThree:
Might be worth looking at PlasmaPy's [`Particle`](https://docs.plasmapy.org/en/stable/particles/index.html#introduction) module, or the [`mendeleev`](https://mendeleev.readthedocs.io/en/stable/) package to see if there's something that already does this?

ZedThree:
PlasmaPy sort of:

```python
import plasmapy.particles as pp

particle = pp.ionic_symbol(1, mass_numb=3, Z=1)
print(particle)
# 'T 1+'

pp.element_name(particle)
# 'hydrogen'
```

Note that it does need to know the atomic number, and I don't see an obvious way to get the name "tritium" although it does understand that `element_name("tritium") == "hydrogen"`.

I think without the atomic number, this can only ever be a "best guess"

arkabokshi:
Thanks Bhavin.  As we have been discussing on the sidelines, these are my current thoughts on the main issues:

1. How do we distinguish between He-ash and fast-alphas when converting from one code to another?  Current thought is to label the species with the higher temp as fast (and the other is then ash)?

2. About naming the species, I think it may be possible to identify _all_ the tokamak relevant species purely on the basis of the mass number (with may be the exception of He3 and Tritium).  Consider the following code which tries to open up gaps in the periodic table based on 2 exclusionary criteria:

- What are the likely elements we can expect in the tokamak?
- With the exception of T, we do not expect any other element to be produced in-situ and hence if the relative abundance is ~0, they do not appear in nature.

```python
import numpy as np
from periodictable import elements
likely_elements = ['H','He','Li','Be','C','N','O','Ar','Ne','W']
for Z,element in enumerate(elements):
    if str(element) in likely_elements:
        for isotope in element:
            if isotope.abundance > 0.0:
                print(f'{element},{isotope.mass:.0f}')
```

With this the only possible candidates are below and we needn't use another criterion such as `Z` which could complicate things due to partial ionization etc.

```H,1
H,2
He,3
He,4
Li,6
Li,7
Be,9
C,12
C,13
N,14
N,15
O,16
O,17
O,18
Ne,20
Ne,21
Ne,22
Ar,36
Ar,38
[...]

PR #440, opened 2025-06-04
Bugfix to prevent AttributeError

physycola:
Loading a pfile with impurity species was leading to an `AttributeError` as it was trying to call `charge.m` on a `UnitSpline` object. Changed to be the same format as the other species as I didn't see some fundamental reason why the `UnitSpline` construction was done before calling `ion_species_selector`.

### DIFF
--- a/src/pyrokinetics/kinetics/pfile.py
+++ b/src/pyrokinetics/kinetics/pfile.py
@@ -85,8 +85,6 @@ class KineticsReaderpFile(FileReader, file_type="pFile", reads=Kinetics):
         profiles = data["profiles"]
         species = data["species"]
 
-        print(species)
-
         # Interpolate on psi_n.
         te_psi_n = profiles["te"]["psinorm"] * units.dimensionless
         electron_temp_data = profiles["te"]["data"] * 1e3 * units.eV
--- a/src/pyrokinetics/kinetics/pfile.py
+++ b/src/pyrokinetics/kinetics/pfile.py
@@ -46,6 +46,8 @@ def ion_species_selector(nucleons, charge):
             return "tritium"
         if charge.m == 2:
             return "helium3"
+    elif nucleons == 6 and charge.m == 3:
+        return "lithium"
     else:
         return "impurity"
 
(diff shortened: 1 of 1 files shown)


physycola:
Note that if a 50/50 DT mix is treated as a single species, then this has the properties `{'N': 1.5, 'Z': 1.0, 'A': 2.5}`. Is it worth adding this as a named species, or should we stick with this showing up as an impurity species? I've gone for the latter, but I thought it could be worth thinking about?

bpatel2107:
Ah good spot there. 

Fine to add it in. I've heard people call it tokamakium so we could add that in? I do think hydrogenic species shouldn't be called impurities as that is misleading.

arkabokshi:
I think tokamakium is a more exotic mix of everything (including wall material etc) that deposits on the tokamak surface. See https://www.sciencedirect.com/science/article/pii/S0022311596001444. 

Probably best to avoid naming altogether.

src/pyrokinetics/kinetics/pfile.py
```diff
@@ -46,6 +46,8 @@ def ion_species_selector(nucleons, charge):
             return "tritium"
         if charge.m == 2:
             return "helium3"
+    elif nucleons == 6 and charge.m == 3:
```
arkabokshi, line 49: What about Lithium7 (do we care)?

bpatel2107: Yeah probably worth including. This relates to #335 and how far down the rabbit hole we want to go

physycola:
> I think tokamakium is a more exotic mix of everything (including wall material etc) that deposits on the tokamak surface. See https://www.sciencedirect.com/science/article/pii/S0022311596001444.
> 
> Probably best to avoid naming altogether.

The only "issue" with not naming it is that it shows up as an impurity, so if there are any routines that bundle impurities together, it'll throw the baby out with the bath water on those.

arkabokshi:
I guess just call it "DT" (or something that says equal DT mix) then?  I do not understand why it gets classed as an impurity -- maybe that is the issue to fix?

physycola:
I think the reason it's classed as an impurity is because everything that is not already named gets classed as an `impurity`. I've updated the PR to include the identification of "DT_5050".

bpatel2107:
I'm fine with the idea but if I'm being honest I really don't like that name but that is just a personal preference. If I think of a better name I'll let you know

bpatel2107 (approved):
Still don't like the name but is technically fine

PR #442, opened 2025-06-04
Possibly erroneous calculation of ntheta in gs2_geometry_terms

physycola:
When using the `gs2_geometry_terms` diagnostic on a given input file, the `ntheta` that is used to calculate `theta_even` is calculated assuming `nperiod = 1`, while `theta_max` takes into account `nperiod > 1`. This means that there can currently be a mismatch between the theta grid in runs generated using the same `pyro.numerics` and what is displayed by the diagnostic.


physycola:
I couldn't run the tests locally for some reason, so I couldn't test before push. I had not realised that the definition of `ntheta` used appears to be crucial for some of the tests. I will leave it up to @bpatel2107 to reject this PR.

d7919:
The failures appear to be in golden answer tests and the absolute disagreement is relatively small so the failures might not be indicative of something wrong with the new approach, just that the new code has led to slightly different `ntheta`. The GS2 geometry test fails -- this case uses ntheta_multiplier = 1 anyway so the only difference comes because `nperiod=2` here. I suspect the issue is that GS2's theta grid will have `ntheta_total = (2 * nperiod - 1) * ntheta_input_file + 1` whilst here we have either `ntheta_even = (2 * nperiod - 1) * ntheta_input_file` or `ntheta_even = (2 * nperiod - 1) * )(ntheta_input_file + 1)`

bpatel2107:
In linear runs `ntheta` is the number of point in a single $2\pi$ period so it wouldn't match the length of the full ballooning space coordinate so it does look like this is a bug. Only a few points in the test are wrong so my guess is that it is the interpolation issue as mentioned above.

physycola:
Updating `ntheta`  from `ntheta = ((2 * nperiod) - 1) * self.pyro.numerics.ntheta * ntheta_multiplier` to be `ntheta = (((2 * nperiod) - 1) * self.pyro.numerics.ntheta + 1) * ntheta_multiplier` still causes the tests to fail, though with a smaller relative difference.

bpatel2107:
This is currently failing on `cvdrift` close to where it is close to 0

Blue is pyro, orange is GS2 and the green is the relative error
image


Increasing `atol = 1e-2` seems to work

The issue is now that the raw GS2 test data (orange) for `aplot` isn't very good due to an old bug in GS2
image
image


@d7919 should we just make new test data with a new GS2 output file?

d7919:
@bpatel2107 yes I think regenerating the test data is probably sensible given that plot!

bpatel2107:
This is actually failing now due to a new release of `pint-xarray` which now supports units on the dimensions (I think).

Will make a new issue and pin in the mean time

bpatel2107:
Failing tests now addressed in #446 so will merge and close now.

Issue #443, opened 2025-06-05
Incorrect calculation of a_minor with GEQDSK and easy fix


jdominsk:
Hi,

Many quantities (normalized gradients,...) are incorrectly computed, because of an incorrect estimate of the minor radius. There is a quick fix. 

The minor radius computed by pyrokinetics from a GEQDSK file is incorrect: 58cm instead of 45cm, which I estimate as half the diameter:  (2.987-2.071)/2 =0.4580. 

It seems that the automatic contour algorithm "_flux_surface_contour" follows the left leg of the separatrix until the limit of my geqdsk domain. From this observation I can recover pyro estimation of a_minor to be close to 58cm: (2.987-1.809)/2 = 0.5890

![Image](https://github.com/user-attachments/assets/e9a3f553-6fb7-4597-b0dc-914c9c3fdb35)

I did a simple fix by imposing minimal Z_min and R_min in geqdsk.py (could you do this as an optional input of pyro?)
            _Z_min = max(Z_min, -0.636996* units.m)
            R_min = max(R_min, 2.0708* units.m)_

        for idx, psi in enumerate(psi_grid[1:], start=1):
            Rc, Zc = _flux_surface_contour(R, Z, psi_RZ, R_axis, Z_axis, psi)
            R_min, R_max = min(Rc), max(Rc)
            Z_min, Z_max = min(Zc), max(Zc)
            Z_min = max(Z_min,-0.636996* units.m)
            R_min = max(R_min,2.0708* units.m)
            R_major[idx] = 0.5 * (R_max + R_min)
            r_minor[idx] = 0.5 * (R_max - R_min)
            Z_mid[idx] = 0.5 * (Z_max + Z_min)

        a_minor = r_minor[-1]

bpatel2107:
Oof that is a big one!

This is a consequence of the contour fitting as you say with the contour finding the inner leg. This is a little harder to fix with the current implementation in a general way but I will have a think. This sometimes can happen with MAST-U data.

A quick (somewhat hacky) fix is to bring in the LCFS a little by setting `eq_kwargs={psi_n_lcfs:0.995}` or a number close to one as this usually avoids in divertor legs.

You ideally want to know how to cut out the divertor legs in a clever way. Would you mind attaching the GEQDSK file for us to have a play. Adding a flag is not the cleanest way to do this so I believe should be left as a last resort.

jdominsk:
[...]

PR #444, opened 2025-06-06
Add check for LCFS a minor in case of divertor legs

bpatel2107:
An attempt to patch (not necessarily fix) an issue raised in #443 

Essentially you can have cases where the contour fitting of the equilibrium finds the divertor legs in the fit to the flux surface. Our current algorithm looks for all connected contour at a fixed value of $\psi$ and then chooses the contour where the summed normalised distance to the magnetic is the smallest. 

Inside the LCFS this works really nicely without issues from the divertor legs as they aren't really connected to the flux surface. Here is an example from MAST-U and pyro will choose the red one by default.
image

The issue is that on the LCFS if divertor legs are connected according to the contour fit then the algorithm either chooses the wrong contour (in the case below it chooses the blue one) or as in the issue raised it can get $R_{min}$ wrong

image

The work around here essentially extrapolates the minor radius outwards from the second to last point and then checks to see if the calculated $a_{minor}$ is close to this within a specified tolerance. If not a error is raised suggesting `psi_n_lcfs` is lowered a little. Usually setting it to something like `0.999` is fine. See below
image

If someone can think of a better way of getting that contour right then I am open to suggestions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,6 +48,7 @@ dependencies = [
     "periodictable",
     "toml",
     "sympy",
+    "shapely",
 ]
 
 [project.optional-dependencies]
--- a/src/pyrokinetics/equilibrium/flux_surface.py
+++ b/src/pyrokinetics/equilibrium/flux_surface.py
@@ -124,13 +124,12 @@ def _flux_surface_contour(
                         ]
                     )
                     intersection = contour_ls.intersection(line)
-
                     if intersection.geom_type == "Point":
                         # Ensure intersection is in original contour
                         if np.isclose(point[1], intersection.x) and np.isclose(
                             point[0], intersection.y
                         ):
-                            new_contour.append([intersection.x, intersection.y])
+                            new_contour.append([intersection.y, intersection.x])
                     elif intersection.geom_type == "MultiPoint":
                         min_distance_idx = np.argmin(
                             [distance(inter, origin) for inter in intersection.geoms]
@@ -141,8 +140,7 @@ def _flux_surface_contour(
                         if np.isclose(point[1], min_distance_R) and np.isclose(
                             point[0], min_distance_Z
                         ):
-                            new_contour.append([min_distance_R, min_distance_Z])
-
+                            new_contour.append([min_distance_Z, min_distance_R])
                 new_contour = np.array(new_contour)
                 new_contours.append(new_contour)
             contours = new_contours
--- a/src/pyrokinetics/equilibrium/flux_surface.py
+++ b/src/pyrokinetics/equilibrium/flux_surface.py
@@ -7,7 +7,7 @@ import numpy as np
 from contourpy import contour_generator
 from numpy.typing import ArrayLike
 from shapely import distance
-from shapely.geometry import LineString, MultiLineString, MultiPoint, Point
+from shapely.geometry import LineString, Point
 
 from ..dataset_wrapper import DatasetWrapper
 from ..units import ureg as units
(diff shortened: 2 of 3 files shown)


bpatel2107:
@jdominsk FYI

LiamPattinson:
I'm not sure about how to return the correct contour when looking exactly on the LCFS, but a way to detect if it's going along a divertor leg would be to check that each subsequent point is clockwise/anticlockwise relative to the previous one as you travel around the flux surface. It wouldn't be a fix, but it would at least allow us to reliably raise an error if the fitting has gone wrong.

d7919:
I think I've done something similarly in the past, code dump here to see if helpful.

~~~python
   def flux_surface(self, psi = None, psiN = None, isgnBpol = 1):
        the_psi = self._get_psi(psi, psiN)
        the_psiN = self._get_psiN(the_psi, None)

        if the_psiN < 1.0e-3:
            print("Warning : Requested psiN very small - likely no contours available so making psiN = 1.0e-3")
            the_psiN = 1.0e-3
            the_psi = self._get_psi(None, the_psiN)

        if abs(isgnBpol) != 1:
            raise ValueError("ERROR: isgnBpol should have magnitude of exactly 1.")

        # We need to find the contours of constant psi as a funcion of R,Z
        # Probably getting data out of a matplotlib contour is the best bet.

        fig = plt.figure()
        cnt = plt.contour(self['R'], self['Z'], self['psirz'].T, levels=[the_psi])
        path = cnt.collections[0].get_paths()
        plt.close(fig)

        # Now we need to work out which path is correct, probably the one with the most points?
        # More strictly it's the one where all points are <= bdry/limitr
        # We can pick any point on the path an draw the line to the origin. If the
        # number of times this intersects the bdry is even then that point is outside
        # the boundary and hence that path can't be a valid one. But then we need to work out
        # how to find out if these things intersect (and if so how many times)
        maxR = self['R'].max() ; minR = self['R'].min()
        maxZ = self['Z'].max() ; minZ = self['Z'].min()

        results = []
        for i, case in enumerate(path):
            rr = case.vertices[:,0]
            zz = case.vertices[:,1]
            if (rr > maxR).any() or (rr < minR).any() or (zz > maxZ).any() or (zz < minZ).any():
                #Can't be this case as part of contour outside boundary
                pass
            else:
                results.append(case)

        if len(results) == 1:
            result = results[0]
        else:
            # Otherwise find the path with the least average distance from rmaxis, zmaxis
            distances = [ np.mean( np.sqrt((x.vertices[:,0] - self['rmaxis'])**2 +
                                           (x.vertices[:,1] - self['zmaxis'])**2)) for x in results]
            result = results[np.array(distances).argmin()]

        # Now we have the geometry let's save it.
        R, Z = result.vertices[:,0], result.vertices[:,1]
~~~

bpatel2107:
@d7919 We essentially have the same thing as the second part of your code where we check the average distance from the magnetic axis (see below)

https://github.com/pyro-kinetics/pyrokinetics/blob/5a933a4993c9fbbb12eb5a00e171cba1141116de/src/pyrokinetics/equilibrium/flux_surface.py#L98-L105


And I think the first section with the intersection with the limiter would not quite work here as the contour found that includes the physical LCFS does intersect with the boundary. It just so happens the the contour goes around the X point.

But your comment about the intersection with the boundary would work, i.e. for each point go inwards to the axis and see if you intersect with another point and if so then drop that point?

bpatel2107:
So i found a library called `shapely` that looks can find intersections between points that _almost_ works.

Here is a example of it now for the case above
image

Essentially from each point I find all the intersections I remove the furtherest point if there are multiple and then continue along.

Here is what I get following that
image

Right at the X point it is a little dodgy
image


Not sure it is worth doing much more beyond this. Thoughts?

I also do a loop over the points. I wonder if I can do all of them at once but I couldn't figure it out immediately

bpatel2107:
There was a small bug in the way that I was appending the new contour we found (I appended `R,Z` rather than the expected `Z,R`) but that fix solves the issues outlined in #443 

Will merge and close now

PR #447, opened 2025-06-11
Bugfix/hydrogenic species

bpatel2107:
Add in ability to reading "hydrogenic" species which have a charge of 1 and a mass between 1 and 3 $m_{proton}$ for the `pFile` format (not sure if any other codes do this).

We label this as hydrogenic and would be used in cases where you have a combined DT, HT, HD main ion mix.

Added a test for this by modifying the template `pFile` as we need to create a new reference mass `mref_hydrogenic`


physycola (approved):
Looks good to me; thanks for spotting the species fix @bpatel2107. Seems to behave as expected in the use-cases that I have for it (e.g., outputting the correct mass/charge in `pyro.local_species`), and also passes all of the automatic tests when run locally. 

Happy for this to be merged.

Issue #445, opened 2025-06-06
No units when calling `load_gk_output` on `PyroScan` object


physycola:
As the title said, any `gk_output` loaded from a `PyroScan` object will not have any units/dimensions due to having to use `np.reshape` to broadcast to the correct shape for the `xarray` objects. The dimensions (e.g., `time`, `ky`, etc.) also do not have any dimensions stored as attributes (as is the case when loading from a normal `Pyro` object). Thought it was worth flagging here so that people are aware, and in case someone has an easy fix to this.

PR #450, opened 2025-06-12
Add units to PyroScan output

bpatel2107:
Adds units and a `to` method to PyroScan output, addresses #445 

This will likely need changing when we update to the latest version to `pint` (requires another release so may be a while)

### DIFF
--- a/src/pyrokinetics/pyroscan.py
+++ b/src/pyrokinetics/pyroscan.py
@@ -16,6 +16,7 @@ import xarray as xr
 from .gk_code import GKInput
 from .normalisation import ConventionNormalisation
 from .pyro import Pyro
+from .units import ureg
 
 
 class PyroScan:
@@ -101,6 +102,12 @@ class PyroScan:
                 self.pyroscan_json = json.load(f)
 
             for key, value in self.pyroscan_json.items():
+                # Add units if stored
+                if key == "parameter_dict":
+                    for param_key, param_value in value.items():
+                        if param_value[-1] in ureg:
+                            value[param_key] = param_value[0] * ureg(param_value[-1])
+
                 setattr(self, key, value)
         else:
             self.pyroscan_json = {attr: getattr(self, attr) for attr in self.JSON_ATTRS}
@@ -606,7 +613,7 @@ class NumpyEncoder(json.JSONEncoder):
         if isinstance(obj, np.floating):
             return float(obj)
         if isinstance(obj, pint.Quantity):
-            return obj.m
+            return [obj.m, str(obj.units)]
         return json.JSONEncoder.default(self, obj)
 
 
--- a/src/pyrokinetics/pyroscan.py
+++ b/src/pyrokinetics/pyroscan.py
@@ -13,6 +13,8 @@ import pint
 
 from .gk_code import GKInput
 from .pyro import Pyro
+from .units import PyroQuantity as Quantity
+from .normalisation import ConventionNormalisation
 
 
 class PyroScan:
@@ -296,7 +298,16 @@ class PyroScan:
         import xarray as xr
 
         # xarray DataSet to store data
-        ds = xr.Dataset(self.parameter_dict)
+        dimensionless_parameter_dict = {
+            key: Quantity(value).m for key, value in self.parameter_dict.items()
+        }
+        ds = xr.Dataset(dimensionless_parameter_dict)
+
+        coord_units = {
+            coord: Quantity(value).units for coord, value in self.parameter_dict.items()
+        }
+        for coord, units in coord_units.items():
+            ds[coord] = ds[coord].assign_attrs(units=units)
 
         # TODO Need to add property to GKCode checking if it is an eigensolver
         # or initial value run and then set nmodes accordingly
@@ -304,6 +315,7 @@ class PyroScan:
             nmode = self.base_pyro.gk_input.data.get("nmodes", 2)
             nmode_coords = {"nmode": list(range(1, 1 + nmode))}
             ds = ds.assign_coords(nmode_coords)
+            ds["nmode"] = ds["nmode"].assign_attrs(units="dimensionless")
         else:
             nmode = np.nan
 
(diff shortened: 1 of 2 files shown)


physycola (commented):
Not sure whether I am doing something wrong, but if I try 

```
pyro_scan.load_gk_output()
gk_data = pyro_scan.gk_output
gk_data.to(pyro.norms.cgyro)
```
I get the error that `AttributeError: 'Dataset' object has no attribute 'to'`. I am unsure as to how, given that the `.to` method is explicitly there.

bpatel2107:
So this is because the `PyroScan.gk_output` attribute is just an xarray Dataset, not our bespoke `GKOutput`

My temporary workaround is you need to do

`pyroscan.to(pyro.norms.cgyro)`

Which will transform it correctly.

I agree that we ought to make it consistent with the other method. Would require a thin wrapper around the Dataset which should be trivial. Could be added here.

physycola:
Yeah I tried that when I got the other error, but `pyro_scan.to(pyro.norms.cgyro)` leads to `AttributeError: 'PyroScan' object has no attribute 'gk_output'` if called before `pyro_scan.load_gk_output()` (obviously), but also gives `AttributeError: 'NoneType' object has no attribute 'items'` when called after (less obviously)...

bpatel2107:
Can you send the full error?

physycola:
For the record, the error was

```
Plotting pyroscan outputs.
────────────────────────────────────────────────────────────────────────────────────────────────────
{'electron': 'electron', 'ion1': 'deuterium', 'ion2': 'carbon'}
D:\Users\tobya\source\pyrokinetics\src\pyrokinetics\gk_code\gk_output.py:729: UserWarning: Mode frequency may not be accurate due to low temporal resolution
  warnings.warn(
D:\Users\tobya\source\pyrokinetics\src\pyrokinetics\gk_code\gk_output.py:729: UserWarning: Mode frequency may not be accurate due to low temporal resolution
  warnings.warn(
C:\Users\tobya\AppData\Local\Programs\Python\Python311\Lib\site-packages\xarray\core\variable.py:363: UnitStrippedWarning: The unit of the quantity is stripped when downcasting to ndarray.
  data = np.asarray(data)
  File "D:\Users\tobya\source\gk_turbulence\python_tools\PYRO_compare_pyroscan.py", line 283, in <module>
    compare_pyroscans(filepaths, species_to_plot, iky_plot=-5, tolerance=0.1)
  File "D:\Users\tobya\source\gk_turbulence\python_tools\PYRO_compare_pyroscan.py", line 144, in compare_pyroscans
    pyro_scan.to(pyro.norms.cgyro)
  File "D:\Users\tobya\source\pyrokinetics\src\pyrokinetics\pyroscan.py", line 504, in to
    ).to(norms)
      ^^^^^^^^^
  File "C:\Users\tobya\AppData\Local\Programs\Python\Python311\Lib\site-packages\pint\facets\plain\quantity.py", line 536, in to
    magnitude = self._convert_magnitude_not_inplace(other, *contexts, **ctx_kwargs)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\tobya\AppData\Local\Programs\Python\Python311\Lib\site-packages\pint\facets\plain\quantity.py", line 480, in _convert_magnitude_not_inplace
    return self._REGISTRY.convert(self._magnitude, self._units, other)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\tobya\AppData\Local\Programs\Python\Python311\Lib\site-packages\pint\facets\plain\registry.py", line 1050, in convert
    return self._convert(value, src, dst, inplace)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\tobya\AppData\Local\Programs\Python\Python311\Lib\site-packages\pint\facets\context\registry.py", line 405, in _convert
    return super()._convert(value, src, dst, inplace)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\tobya\AppData\Local\Programs\Python\Python311\Lib\site-packages\pint\facets\nonmultiplicative\registry.py", line 251, in _convert
    dst_offset_unit = self._validate_and_extract(dst)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\tobya\AppData\Local\Programs\Python\Python311\Lib\site-packages\pint\facets\nonmultiplicative\registry.py", line 166, in _validate_and_extract
    (u, e) for u, e in units.items() if not self._is_multiplicative(u)
                       ^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'items'
```

bpatel2107:
So I went a bit wild here but the issue above was occurring due to the parameter in the scan not having any units.

I had to be careful here in the scenario where the user does not specify any units but pyro has to add it on. In this case I now raise a warning saying that a unit is being added and ensure that unit is added to the `GKOutput`.

What I have also done is written a small wrapper around the `xarray.Dataset` where you can do `pyro_scan.gk_output.to(pyro.norms.gs2)` to convert units, which is consistent to the regular `Pyro` object.

This should now hopefully work.

physycola:
I will test this locally on Monday. But I thought I'd note here that the reason it was failing for me (as @bpatel2107 pointed out) was because I was loading the scan using the 'pyroscan.json' file, which does not record the normalisation convention originally used. Is there a reason why this isn't included as, say, a string?

bpatel2107:
When specifying the values to be used in the scan do you have units cos if not then maybe the json file wouldnt have it. Not tested this though.

physycola:
I did actually specify the units in that case, viz., 
```
values_1 = np.geomspace(1e-2, 1e2, 20) * pyro.numerics.ky.to(pyro.norms.pyrokinetics)
```
but this did not seem to be stored in the `.json` file.

bpatel2107:
I think this is done now so I will merge!

PR #451, opened 2025-06-16
Bugfix/pfile divide by zero

physycola:
Minor change to allow for proper handling of cases where the density of a species is zero due to the finite precision of the pfile data (was leading to the temperature being all `nan` even if the density was only zero at, e.g., the edge).


physycola:
Closed PR as did not branch from unstable (whoops!).

PR #452, opened 2025-06-16
Bugfix/pfile divide by zero

physycola:
Minor change to allow for proper handling of cases where the density of a species is zero due to the finite precision of the pfile data (was leading to the temperature being all `nan` even if the density was only zero at, e.g., the edge).

### DIFF
--- a/src/pyrokinetics/kinetics/pfile.py
+++ b/src/pyrokinetics/kinetics/pfile.py
@@ -5,6 +5,7 @@ Reads in an Osborne pFile: https://omfit.io/_modules/omfit_classes/omfit_osborne
 """
 
 import re
+import warnings
 from contextlib import redirect_stdout
 from textwrap import dedent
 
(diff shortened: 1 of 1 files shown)


bpatel2107 (approved):
Looks good to me

physycola:
Is there any reason to set the elements that would error to `nan` instead of `0.0`? The particular case in which this occurs is when the pressure is finite but the density is zero.

bpatel2107:
Well I'm not sure under what scenario that would happen (zero density but non-zero pressure). That is nonphysical so should we even support that?

physycola:
It's due to the finite-precision used in the `pfile` (partially due to the normalisations). See, e.g., the attached `pfile` as an example. The fast ions have zero `nb` in the last few entries but nonzero (but very small) `pb` due to the fact that the density cannot be smaller than `0.000001`. It's not `unphysical` (the pressure profiles are a solution to GS), it's just that not enough precision has been used to store the data. 

[pFile.zip](https://github.com/user-attachments/files/20776985/pFile.zip)

bpatel2107:
Sure but that just feels like not a pyro issue to be honest and a formatting issue in the data.

You can't reasonably set a "fast ion temperature" without the density if the pressure is non-zero.

This would just highlight that the user needs to reformat the output.

I guess a NaN is more appropriate than a 0.0 here as 0.0 is definitely wrong

physycola:
Unfortunately, masking with `nan` causes `UnitSpline` to return all `nan`'s when interpolating onto some grid. I do not see another "safe" way of doing this, so it's probably best that we don't try and support such `pfile` data.

PR #455, opened 2025-06-23
Added writing of adiabatic electron flags in .set method

physycola:
If one was converting input files from other GK codes to `CGYRO`, the `CGYRO`-specific adiabatic electron flags required by `PYRO` were not being set, so I have added these. These changes work on a branch before the implementation of `check_quasineutrality`, which currently causes all adiabatic electron runs/conversions to fail @bpatel2107 .

### DIFF
--- a/src/pyrokinetics/gk_code/cgyro.py
+++ b/src/pyrokinetics/gk_code/cgyro.py
@@ -859,10 +859,10 @@ class GKInputCGYRO(GKInput, FileReader, file_type="CGYRO", reads=GKInput):
             )
 
             # Set adiabatic flags
-            self.data['AE_FLAG'] = 1
-            self.data['DENS_AE'] = ne
-            self.data['TEMP_AE'] = te
-            self.data['MASS_AE'] = me
+            self.data["AE_FLAG"] = 1
+            self.data["DENS_AE"] = ne
+            self.data["TEMP_AE"] = te
+            self.data["MASS_AE"] = me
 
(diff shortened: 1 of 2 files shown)


physycola:
Though it perhaps warrants a different PR, the loading of `CGYRO` data with adiabatic electrons appears is bugged, as the logic for getting `rho_star` from the `.equillbrium` file fails for `nspecies=1`.  I can't quite figure out which is the correct condition/index for `rho_star` in this case, however.

bpatel2107 (approved):
This looks good to me! I'm a little surprised about the `rho_star` as the species gets written after the `rho_star` value... What does pyro think it is?

physycola:
@bpatel2107, the length of the file does not seem to fall into any of the values given by the logic in `cgyro.py`, and so I thinks it's at index `23`, which is just zero. I don't quite understand why, though...

bpatel2107:
My guess is that the `n_species` may not match for some reason. Can you send the input and equilibrium file to double check?

physycola:
Files are attached below @bpatel2107 . `N_SPECIES = 1` in the input file, and I think the appropriate `AE` flags are set, but the length of `out.cgyro.equilibrium` still isn't one of the cases in `cgyro.py`.

[out.cgyro.zip](https://github.com/user-attachments/files/20884574/out.cgyro.zip)

physycola:
Will merge as the only tests that are failing are the ones for the poincare plot, which this PR has not touched.

PR #456, opened 2025-06-24
Ordering of local species when writing to GX input file

physycola:
`GX` does not like an adiabatic species being before others in the species list, and so I've moved the electron species to the end of the list when writing to the input file to make it easier to make them adiabatic.


physycola:
The rountrip tests fail because the resultant `gx` file obviously differs from the initial one. Do we want to proceed by updating the initial file?

bpatel2107 (approved):
Made a tweak to the test but looks good now

PR #457, opened 2025-06-25
Adding back in lines removed in previous commit.

physycola:
Lines were removed in commit `de65a2f6bc9b49eaaa9de46e8982ccf01920a15b` that also removed backwards compatability for older `CGYRO` runs. For example, `CGYRO` runs on their commit `05e3c07fb9b2ddb350ffef2c2c170433fdeaa861` would currently fail, as these are actually formatted using `55 + 10 * nspecies`.


bpatel2107 (changes requested):
Looks fine but fails on flake8 because of the unused f string in LocalSpecies. Surprised this hasn't come up before but is easy enough to fix.

Can you remove the "f" so it will pass the test?

physycola:
It passes the flake8 tests now, but is still failing due to the problem with the poincare tests highlighted in #462

bpatel2107 (approved):
Looks good to me

PR #463, opened 2025-07-01
Bugfix/gs2 geom units

### LINKED ISSUE #460: Possible bug in `gs2_geometry_terms`
I think I have possibly found a bug in the `gs2_geometry_terms` diagnostic.

 If I plot `(cvdrift - gbdrift)/gbdrift` either output by the `pyro` diagnostic, or by `GS2` itself, the `pyro` result seems to disagree quite a lot with what `GS2` is outputting, as below:

Image

If I load the exact same geometry but instead from `CGYRO`, it seems that it also disagrees with the `GS2` result, but in a different way (in particular, near zero), as below:

Image

Does this mean it is a bug with the `gs2_geometry_terms` diagnostic, or something more fundamental is going on? I have attached the input files for both codes, and I've uploaded the `gs2` output file to a [google drive](https://drive.google.com/file/d/1HzKlHr5YXyxvJh4wBIgyfjP9LRRogYvK/view?usp=drive_link).

[input_files.zip](https://github.com/user-attachments/files/20957135/input_files.zip)

### LINKED ISSUE #462: Scipy v1.16 breaking poincare test
All the latest PRs are failing the `test_poincare.py` test for python >=3.11 when installing `scipy==1.16`. Given that it is not available for 3.10, those tests run fine.

The issue is likely in the `RectBivariateSpline` with the only change being an additional kwarg `maxit` which changes the number of iterations used when smoothing.

Probably easier to either pin or regenerate the test data. That bit of code in getting refactored in a branch anyway so its not really worth trying to "fix"

Thoughts?

bpatel2107:
Fixes #460 as `beta_prime` in `MetricTerm` was dropping its units which meant conversions weren't being done

Also fixed #462 as this is breaking the `test_poincare.py` test by pinning `scipy <= 1.15.3`


physycola:
This appears to give the correct agreement between the two cases, but it fails a lot of the tests (see #460 for details of the comparison).

image

bpatel2107:
So there was a bug in the setting of `local_geometry.dpsidr` when calculating it from an existing gk input file which led to a few issues which eventually permeated down to `MetricTerms`.

Essentially we do a calculation where `dpsidr = bunit_over_b0 * rho / q` which is accurate but by definition is in units of `bref_B0 * lref`. However, this gets calculated before units are assigned and the units are dependent on the code. A correction for `bref_GKCODE / bref_B0` needs to be applied which is done here.

This becomes important here as `dpsidr` didn't have the right units in `MetricTerms` so it snuck through the testing. I have fixed this and made new test data for poincare as this was impacting it a little.

PR #464, opened 2025-07-02
Fixed comments at L1906

ILexiconI:
Fixed a mix up in minor/major radius.

### DIFF
--- a/src/pyrokinetics/pyro.py
+++ b/src/pyrokinetics/pyro.py
@@ -1903,8 +1903,10 @@ class Pyro:
             Electron density
         bref_b0: [tesla] pint.Quantity
             Toroidal magnetic field at centre of flux surface
-        lref_major_radius: [meter] pint.Quantity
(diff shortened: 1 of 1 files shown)


bpatel2107 (commented):
So the major radius here is actually the major radius of the local flux surface, not the magnetic axis so that would need to be tweaked here

ILexiconI:
Oh, I was aware that that normalisation was sometimes used but figured since the minor radius was normalised to a global parameter that the major radius would be too, I'll make another pull request.

bpatel2107:
You can just modify this one and that is a fair point. We in theory do support the magnetic axis for the reference length but it isn't often used in pyro as its isn't a local flux surface value. Neither is the minor radius but choices were made.

ILexiconI:
Just pushed the changes to my fork, has that automatically updated the pull request here?

bpatel2107 (approved):
This is failing tests due to a different issue fixed somewhere else. This PR is good to go

PR #465, opened 2025-07-17
Feature/eliteinp io

bpatel2107:
Add in reader for ELITEINP files which provide high resolution equilibrium and kinetics data.

Also added in test for the different readers.

@ILexiconI FYI


src/pyrokinetics/equilibrium/eliteinp.py
```diff
@@ -0,0 +1,197 @@
...
+        data_dict["p_prime"] = (
+            data_dict["dp/dpsi"] * units.pascal / (units.weber / units.radian)
+        )
+
+    if "ffp" in data_dict:
+        data_dict["ff_prime"] = data_dict["ffp"] * (
+            units.tesla**2 * units.meter**2 * units.radian / units.weber
+        )
+
+    if "Bt" not in data_dict and "fpol" in data_dict and "R" in data_dict:
```
d7919, line 72: Should we abort if expected keys are not found here, e.g. if both Bt and fpol / R are not available then we should abort here?

src/pyrokinetics/equilibrium/eliteinp.py
```diff
@@ -0,0 +1,197 @@
...
+        R = data_dict["R"]  # shape (npsi, npol)
+        data_dict["Bt"] = fpol[:, None] * R  # shape (npsi, npol)
+
+    if "p" not in data_dict:
+        if "ne" in data_dict and "Te" in data_dict:
+            data_dict["p"] = (
+                data_dict["ne"] * (data_dict["Te"] * electron_charge.m)
+            ) * units.pascal
+        if "nMainIon" in data_dict and "Ti" in data_dict:
+            data_dict["p"] += (
```
d7919: Can we use `+=` here if `ne` and `Te` were not in the file as well?

src/pyrokinetics/equilibrium/eliteinp.py
```diff
@@ -0,0 +1,197 @@
...
+                data_dict["nMainIon"] * (data_dict["Ti"] * electron_charge.m)
+            ) * units.pascal
+        if "nZ" in data_dict and "Ti" in data_dict:
+            data_dict["p"] += (
+                data_dict["nZ"] * (data_dict["Ti"] * electron_charge.m)
+            ) * units.pascal
+        else:
```
d7919: The message in this else looks like it should be on the first `if` in this block. Should there be a message and raise for each of these if statements?

src/pyrokinetics/equilibrium/eliteinp.py
```diff
@@ -0,0 +1,197 @@
...
+            I_p=Ip,
+            clockwise_phi=clockwise_phi,
+            cocos=cocos,
+            eq_type="ELITEINP",
+        )
+
+    def verify_file_type(self, filename: PathLike) -> None:
+        with open(filename, "r") as f:
+            head = f.read(50).upper()
+        if "HELENA GENERATED INPUT" not in head:
```
d7919: Other codes _can_ write this format I think?

src/pyrokinetics/kinetics/eliteinp.py
```diff
@@ -0,0 +1,189 @@
...
+        data_dict["Ti"] = data_dict["Te"]
+    elif "Ti" in data_dict and "Te" not in data_dict:
+        data_dict["Te"] = data_dict["Ti"]
+
+    if "p" not in data_dict:
+        if "ne" in data_dict and "Te" in data_dict:
+            data_dict["p"] = (
+                data_dict["ne"]
+                * (data_dict["Te"] + data_dict["Ti"])
+                * 1.602e-19
```
d7919: Should this use electron charge constant rather than raw value?

src/pyrokinetics/kinetics/eliteinp.py
```diff
@@ -0,0 +1,189 @@
...
+            )
+
+            result["impurity"] = impurity
+
+        return Kinetics(kinetics_type="ELITEINP", **result)
+
+    def verify_file_type(self, filename: PathLike) -> None:
+        with open(filename, "r") as f:
+            head = f.read(50).upper()
+        if "HELENA GENERATED INPUT" not in head:
```
d7919: As earlier, this might fail on some valid files as other codes _can_ write this format.

src/pyrokinetics/kinetics/eliteinp.py
bpatel2107: We actually don't need `p` here so this can be removed entirely.

src/pyrokinetics/equilibrium/eliteinp.py
bpatel2107: If `ne` and `Te` aren't in the file we should probably abort

src/pyrokinetics/equilibrium/eliteinp.py
bpatel2107: Hmm I guess? We could look for more keys but that does involve reading more of the file. We could also check for `Psi:` which is written near the top? Not sure if that is strict enough...

d7919: SCENE writes `Input data for ELITE, shape =eqbm format` in the header. Maybe it is just a case of building up a set of valid header checks?

src/pyrokinetics/templates/test.eliteinp
ZedThree, line 1: This is a pretty large file! Does it make sense to use a cut-down version instead?

bpatel2107: That was the cut down version 😢

src/pyrokinetics/equilibrium/eliteinp.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+def read_eqin(filename_or_file):
+    def token_generator(fp):
+        for line in fp:
+            for tok in line.split():
+                yield tok
+
+    if isinstance(filename_or_file, (str, Path)):
+        with open(str(filename_or_file), "r") as fh:
```
ZedThree: No need to use `str` here, `open` can handle `Path`.

I'm also a little confused by the logic flow here. Where can call `read_eqin` with a file handle that isn't here? I'm not super keen on the lone `f.close()` below. If you do really need to sometimes call with a file handle, maybe if you pull out actually reading the data, then it could look just like:

```py
def read_data(file):
   ...

def read_eqin(filename_or_file):
    if isinstance(filename_or_file, (str, Path)):
        with open(filename_or_file) as f:
            data = read_data(f)
    else:
        data = read_data(filename_or_file)

    # check units etc with `data`
```

bpatel2107: I can modify to use a `with` instead. Much cleaner overall

src/pyrokinetics/equilibrium/eliteinp.py
bpatel2107: Added a check for all keys at the end now

src/pyrokinetics/equilibrium/eliteinp.py
bpatel2107: Fixed now

src/pyrokinetics/templates/test.eliteinp
bpatel2107: Found a smaller version now

src/pyrokinetics/equilibrium/eliteinp.py
bpatel2107: I've cleaned this up now

src/pyrokinetics/equilibrium/eliteinp.py
bpatel2107: I've added another small check for the HELENA output but looking at the SCENE output the format/variable names are different (`psi` instead of `Psi:`, `pprime` instead of `dp/dpsi`). This was ran like 10 years ago so idk if much has changed with SCENE but it seems unlikely

This could become a little convoluted if we aren't careful. Maybe we need a dictionary for each key for given source of eliteinp...

d7919: I can't find a hard specification for the file format but I guess if different codes use different field names then these are probably not specified in the standard, which might suggest that only the order of fields is required to match expectations. Someone with access to ELITE might be able to check the source for what is accepted in this format. Here's the most recent SCENE code for writing out (https://github.com/PlasmaFAIR/SCENE/blob/master/src/elite_data.f90#L257).

Maybe we don't worry yet and just add a comment to the code or something to the error message so that if someone shouts in the future that their eliteinp file doesn't work then we have some guidance for what might have gone wrong?

bpatel2107: I have updated the error messages so if someone wants a different format a useful message will pop up.

src/pyrokinetics/equilibrium/eliteinp.py
```diff
@@ -0,0 +1,212 @@
...
+
+    if "p" not in data_dict:
+        if "ne" in data_dict and "Te" in data_dict:
+            data_dict["p"] = (
+                data_dict["ne"] * (data_dict["Te"] * electron_charge.m)
+            ) * units.pascal
+        else:
+            raise ValueError(
+                "No electron density and temperature data found in ELITEINP"
+            )
```
ZedThree: I usually recommend flipping checks like this

```suggestion
        if "ne" not in data_dict or "Te" not in data_dict:
            raise ValueError(
                "No electron density/temperature data found in ELITEINP"
            )

        data_dict["p"] = (
            data_dict["ne"] * (data_dict["Te"] * electron_charge.m)
        ) * units.pascal
```

bpatel2107: Done

PR #468, opened 2025-07-29
Bugfix for PyroScan loading with theta0

bpatel2107:
The recent release broke the loading of `PyroScan` when $\theta_0!=0$ because we selected $k_{x,min}=0.0$ rather than the smallest $k_{x,min}$, so this fixes that.

As a side note using `PyroScan` wouldn't load all the $k_x$ and $k_y$ data for simulation with multiple values. Not sure this is a big problem or a common use case.

### DIFF
--- a/src/pyrokinetics/pyroscan.py
+++ b/src/pyrokinetics/pyroscan.py
@@ -377,28 +377,30 @@ class PyroScan:
                         if 0.0 in pyro.gk_output.data.ky:
                             pyro.gk_output.data = pyro.gk_output.data.isel(ky=[1])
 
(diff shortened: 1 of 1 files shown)


physycola (commented):
The changes appear to give the intended behaviour after a small bugfix. I do not think that we need to support multiple kx's for a given ky in a pyroscan.

Is it worth adding some tests for pyroscan.py, or should this be a separate PR?

bpatel2107:
Separate PR as we need a clever way of storing simulation data online so we can do the testing properly

physycola (approved):
Seems to work well from all the cases that I have tested. Ready to merge!

PR #471, opened 2025-08-04
Removed setting of FIELD_PRINT_FLAG and MOMENT_PRINT_FLAG in cgyro.py

physycola:
The previous behaviour was to set both `MOMENT_PRINT_FLAG` and `FIELD_PRINT_FLAG` in `cgyro.py`, which would overwrite any values that were specified by the user. 

Given that the `'input.cgyro` template has `MOMENT_PRINT_FLAG = 1` and `FIELD_PRINT_FLAG = 1`, the only case where this may cause `pyro` to give unexpected behaviour is if the user supplies an input template that does not specify these flags (the `CGYRO` default is `0`).


bpatel2107 (approved):
This is fine I think given that the template has these set to 1 as you say

Issue #475, opened 2025-08-12
Issue with loading GENE data


physycola:
I have a case where I have two almost identical GENE [input files](https://github.com/user-attachments/files/21742342/parameters.zip) differing only by the regularity of saving the data (as well as the parallelisation for the run). These, however, do not give the same growthrate when loading them with PYRO, as can be seen below, despite GENE reporting the same growthrate. The QL particle and heat fluxes are identical, however, as are the fields. This suggests that something is doing wrong with the handling of the time data, but I have not been able to track it down yet. 

<img width="1166" height="475" alt="Image" src="https://github.com/user-attachments/assets/d0e5aefc-64f7-4203-9b69-a820cdb18d78" />

Issue #473, opened 2025-08-08
Failure to warn with low temporal resolution


physycola:
I have a case ([GENE input file](https://github.com/user-attachments/files/21685361/parameters.zip)) where PYRO fails to warn about the sparse temporal resolution of the data when calculating the growthrate and frequency. If one were to believe the results given by PYRO, they would get potentally suspect values, as is clear from the case below.

<img width="637" height="473" alt="Image" src="https://github.com/user-attachments/assets/f73a2ab3-6ddb-439c-b32f-264413f31bf4" />

Is there a clean way of handling such cases, given we do not know _a prior_ what the temporal resolution _should_ be for a given run?

Issue #472, opened 2025-08-07
PYRO fails to load GENE input file for certain Miller specifications


physycola:
It appears that PYRO currently fails when trying to load from a GENE input file that does not contain `minor_r` as a parameter. The attached input file runs with GENE without any issues (
[parameters_KBM.zip](https://github.com/user-attachments/files/21666952/parameters_KBM.zip)). However, one gets a `divide by zero` error when PYRO tries to calculate `aspect_ratio`. I am not sure what the best way of handling this is?

PR #476, opened 2025-08-22
Multiple bugfixes addressing issues #472, #473, and #475

physycola:
Have addressed multiple outstanding issues, as follows:

- #472 : PYRO was checking for `minor_r` (in order to calculate `aspect_ratio`) in more places than it needed to, and so would fail if `minor_r` was not specified consistently. This is now fixed
- #473 : Added checking of `mode_frequency` when calculating the growth rate tolerance; it will now warn if there is poor frequency data over the last portion of the simulation time.
- #475 : This was due to some legacy code resetting the `self.norm_convention` inside `get_local_geometry`, which has now been removed. 

I think this means that PYRO should once-again interact well with the GENE edge cases reported in the bug reports.

### DIFF
--- a/src/pyrokinetics/gk_code/cgyro.py
+++ b/src/pyrokinetics/gk_code/cgyro.py
@@ -156,7 +156,9 @@ class GKInputCGYRO(GKInput, FileReader, file_type="CGYRO", reads=GKInput):
         3: "Fourier",
     }
 
-    def read_from_file(self, filename: PathLike, detect_norm: bool = True) -> Dict[str, Any]:
+    def read_from_file(
+        self, filename: PathLike, detect_norm: bool = True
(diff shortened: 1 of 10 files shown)


bpatel2107:
Had to make a tweak here as there was a subtle bug here. Essentially the test was failing due to the `_detect_normalisation` method not being called when reading in the input file from a GENE IMAS IDS file. Doing that called the `read_dict` method which did not automatically check for the normalisation. The previous code checked later on the in the code which when removed broke the testing.

I've now changed it such that any way you read a `GKInput` (dict/string/file) the `_detect_normalisation` method will get called unless explicitly stated otherwise. Hence all the changes to the other `GKInputReaders`

bpatel2107 (approved):
Good to merge now

PR #477, opened 2025-08-22
Fixing GENE momentum fluxes

physycola:
The momentum fluxes that were being loaded by PYRO were the default fluxes output by GENE, the toroidal _parallel_ momentum fluxes. However, all other codes really compute the toroidal _angular_ momentum fluxes, and so I have changed the code to preferentially load those if they are present (including diagnostic outputs that are currently being added to GENE by another user).  

I have confirmed that both code paths (for the `.nrg` and `.h5` files) agree with one another, though I do not have a benchmark case for the outputs (currently being worked on).

The only thing I was wondering is as to whether we should warn the user when PYRO is loading the parallel momentum fluxes from GENE, so that they are not confused if they see disagreement with other codes?

### DIFF
--- a/src/pyrokinetics/gk_code/gene.py
+++ b/src/pyrokinetics/gk_code/gene.py
@@ -2168,7 +2168,7 @@ class GKOutputReaderGENE(FileReader, file_type="GENE", reads=GKOutput):
 
                 if nfield == 3:
                     logging.warning(
-                        "GENE combines Apar and Bpar fluxes, setting Bpar fluxes to zero"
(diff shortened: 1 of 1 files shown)


physycola:
This should not be merged quite yet, as things are still being changed in GENE. Will comment again when it is ready for merge.

physycola:
I believe that there will not be any more changes to the GENE outputs for now. I did have to use a somewhat "hacky" solution for loading from the `nrg.dat.h5` file, because there was an issue with the time grids. Perhaps there is a cleaner way of doing this @bpatel2107 ?

bpatel2107 (approved):

PR #481, opened 2025-09-08
Modifying behaviour of downsize in cgyro.py

physycola:
I was getting OOM errors due to `downsize` only being used after processing the data. I've moved this to be done directly after loading the raw data, such that we don't perform operations on data that we will later discard.


bpatel2107 (approved):
LGTM

PR #482, opened 2025-09-11
Use correct rho when setting domega_drho from GENE

bpatel2107:
When loading `local_species.domega_drho` (used to set PVG term) we were using the wrong definition of `rho= r / Lref`. Originally we had

`rho = trpeps * major_R / minor_r`
where 
`trpeps = r / R`
`major_R = R/Lref`
`minor_r = a/ Lref`

We should have had

`rho = trpeps * major_R`

So this would only become an issue when `a != Lref` (using GENE's default convention). This would mean that the PVG term when loading in a GENE input file with `minor_r !=1` would be incorrect, which doesn't get covered by the integrated testing. We could add this in?

### DIFF
--- a/src/pyrokinetics/gk_code/gene.py
+++ b/src/pyrokinetics/gk_code/gene.py
@@ -1281,6 +1281,12 @@ class GKInputGENE(GKInput, FileReader, file_type="GENE", reads=GKInput):
 
         if code_normalisation == "pyrokinetics":
             self.data["geometry"]["minor_r"] = 1.0
+        elif code_normalisation == "gene":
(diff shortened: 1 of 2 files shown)


physycola:
Looks like this is working as intended for me now, as I am getting the correct `pfsrate` given `exbrate` when doing a manual roundtrip from `GENE -> CGYRO -> GENE`. Is it worth adding in a test to cover this in future, as you suggested @bpatel2107 ?

bpatel2107:
I've added a test and caught a small bug in the setting of `minor_r` when using the `gene` normalisation. Will merge now

PR #483, opened 2025-09-11
Add function to set consistent beta_prime and a test

bpatel2107:
Add a function `pyro.enforce_consistent_beta_prime()` that ensures that `pyro.local_geometry.beta_prime` is consistent with `numerics.beta` and the `local_species` object. Written a test for this too


physycola (approved):
Looks good to me. I can't see any obvious way in which this can fail, and it is the correct formula for `beta_prime`.

PR #484, opened 2025-09-23
Downsample CGYRO data before loading into memory

bpatel2107:
Previously Pyro would try to load all the data in the binary files using `np.fromfile`. When working with nonlinear simulations where the output can be 100GB you can often run out of memory and the process will fail.

This PR switches to using `np.memmap` in CGYRO which just puts the location of the data in memory. This allows for downsampling before the data is loaded, which significantly reduces the time taken to read in the data.

To do this you need to add a dictionary `downsample` as a `kwargs` when calling `pyro.load_gk_output`

```python
from pyrokinetics import Pyro

gk_file = "input.cgyro"
pyro = Pyro(gk_file=gk_file)

downsample = {
"time_idx": slice(-100, None, None),
"ky_idx": slice(1, 10, None),
"theta_idx":  16,
"kx_idx": None,
}
pyro.load_gk_output(downsample=downsample)
```

The above example will only load the final 100 timeslices, the kys going from index 1 to 10,  the 16th theta and all the kx. One can specify a single index using an integer or a range of indices using `slice`

I've added a unit test for this which seems to work. Note I have remove the option `downsize` when loading CGYRO output as that is now redundant.

We could try and implement this for other codes. GKW does use `numpy.fromfile` so a similar strategy could be followed but I am not sure about the other codes. Lazy loading of netCDF files (GS2/STELLA) is possible but I am not sure if we do that. GENE output is quite messy so this could be a bit more difficult to implement. Not sure if we just implement as we go?

Also should we change the keys in the downsample dict from `time_idx` to `time`? Its less clear but cleaner...

### DIFF
--- a/src/pyrokinetics/gk_code/cgyro.py
+++ b/src/pyrokinetics/gk_code/cgyro.py
@@ -1024,7 +1024,7 @@ class GKOutputReaderCGYRO(FileReader, file_type="CGYRO", reads=GKOutput):
 
         for key, value in downsample.items():
             if isinstance(value, int):
-                downsample[key] = [value]
+                downsample[key] = slice(value, value + 1)
 
         bunit_over_b0 = (1 * norm.cgyro.bref).to(norm.gs2).m
         coords = self._get_coords(raw_data, gk_input, bunit_over_b0, downsample)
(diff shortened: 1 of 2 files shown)


daviddickinson:
I think using the names of the dimensions as the keys makes sense

bpatel2107:
Okay changed to this

```python
from pyrokinetics import Pyro

gk_file = "input.cgyro"
pyro = Pyro(gk_file=gk_file)

downsample = {
"time": slice(-100, None, None),
"ky": slice(1, 10, None),
"theta":  16,
"kx": None,
}
pyro.load_gk_output(downsample=downsample)
```

PR #485, opened 2025-09-25
Changed the assigned units for nmode in tglf from dimensionless to none. This solved an error with reading tglf pyroscans

FelixWattsYork:
In line 668 of the Pyroscan file, the "to" function in the PyroScanGkOutput object checks if the coord's units is None. If it is not None it attempts to multiply the data by the correct unit which isn't possible with dimensionless. Therefore the code throws the error:  
`new_coord = (self[coord].data * self[coord].units).to(norms)
                 ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~
TypeError: The 'out' kwarg is necessary. Use numpy.strings.multiply without it.`
Setting the unit type to None bypasses this steps and prevents the error being thrown


d7919:
Could you include details about the error you obtained from the original code?

bpatel2107 (approved):

PR #487, opened 2025-10-05
Specifying types

ILexiconI:
Must specify bt_ccw & it_ccw as int for GENE to be able to parse parameters files.  May want to check this doesn't break other codes if they expect floating point values instead.


bpatel2107 (approved):
Looks good

PR #488, opened 2025-10-07
Feature/neoclassical diagnostics

bpatel2107:
This PR adds in the ability to calculate the neoclassical bootstrap current using two different models

1. O. Sauter et al Physics of Plasma 6, 2834 (1999)
2. A. Redl et al Phys. Plasmas 28, 022502 (2021)

I have benchmarked it against SCENE, JETTO and TRANSP runs and it works reasonably well for a few different cases (see below).

The equation roughly looks like the second term on the RHS of the following equation (from Redl 2021)

image

With the coefficients (L31, L32 etc.) being fitted functions to NEO data.

It does not match the TRANSP run in the repo but matches other TRANSP runs but I think that is an older runs so hard to say what happened there.

There is an example and a unit test using the SCENE output data. We don't have the matching equilibrium for JETTO and as I've mentioned the TRANSP data is odd.

The biggest question I have is mostly about the inclusion of fast ions. The papers do include main ions and impurities in the "total ion pressure" but whether fast ions should be treated separately I am not sure.

image

The options I see are

1. Ignore them entirely
2. Include them in the calc of $p_\sigma$ and use the main ion $\frac{\partial T_i}{\partial \psi}$
3. Include them in the calc of $p_\sigma$ and use their equivalent gradient $\frac{\partial T_{fast}}{\partial \psi}$ which would tweak the sum above

I have included the first two as an option, set using a kwarg `ion_type=thermal` or `all` but am not sure about the third.

SCENE
image

JETTO
image

TRANSP
image

### DIFF
--- a/docs/examples/example_modify_shear.py
+++ b/docs/examples/example_modify_shear.py
@@ -1,0 +1,132 @@
+from pyrokinetics import Pyro, template_dir
+from pyrokinetics.diagnostics.neoclassical import Redl2021
+
+# Solves <B^2> f' / μ0 + fp' = <Jtot.B> = <Jbs.B> + <Jext.B>
+# Scales beta/gradients keeping boostrap current consistent and
+# modifies external current to scale like T/n
+
+run = "TRANSP"
+
+if run == "TRANSP":
+    # Path to files
+    eq_file = template_dir / "transp.cdf"
+    kinetics_file = template_dir / "transp.cdf"
+
+    # Load up pyro object
+    pyro = Pyro(
+        eq_file=eq_file,
+        eq_type="TRANSP",
+        eq_kwargs={"time_index": -1},
+        kinetics_file=kinetics_file,
+        kinetics_type="TRANSP",
+        kinetics_kwargs={"time_index": -1},
+    )
+
+elif run == "SCENE":
+    # Path to files
+    eq_file = template_dir / "test.geqdsk"
+    kinetics_file = template_dir / "scene.cdf"
+
+    # Load up pyro object
+    pyro = Pyro(
+        eq_file=eq_file,
+        eq_type="GEQDSK",
+        kinetics_file=kinetics_file,
+        kinetics_type="SCENE",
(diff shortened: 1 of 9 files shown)


bpatel2107:
Okay so this is now ready to merge!

Regarding the treatment of fast ions, we now treat their contribution to the neoclassical bootstrap using an equivalent thermal species, as suggested by `Belli, Emily A., et al. "Limitations of bootstrap current models." Plasma Physics and Controlled Fusion 56.4 (2014): 045006`. We don't allow for any other options as their overestimation can be quite significant. If someone wants to do a NEO run and compare then that could be interesting (an easy paper tbh).


I have also added in some additional functionality

- In `BootstrapModels` you can determine $\langle J_{total}.B\rangle$ and $\langle J_{external}.B\rangle$. I've updated the test to include a check on $\langle J_{total}.B\rangle$
- In `BootstrapModels` for a given $\langle J_{total}.B\rangle$, you can determine $F'$
- In `LocalGeometry` you can determine $F'$ from local geometry parameters only
- In `LocalGeometry` you can determine $\hat{s}$ for a given $F'$
- In `Pyro` you can now load a numerics object using `pyro.load_numerics` (a side thing but was helpful for my example script)

Finally with all of this together I have written an example script (`example_modify_shear.py`) that essentially allows you to scale up density, temperature and the temperature gradients and will then calculate the change in $\langle J_{bs}.B\rangle$ and scales $\langle J_{external}.B\rangle$ like $T/n$.

Then using this new modified total current and, assuming shaping and $q$ remain fixed, determine the new $F'$ needed to satisfy the Grad-Shafranov equation. From the new $F'$ you can calculate the new $\hat{s}$.

This method should in theory allow for more "consistent" scans in a sense. Will leave open for a bit for any discussion on this. If people have thoughts on simple CBC cases where we "know" how $\hat{s}$ should change then that could be helpful to add to testing.

EDIT - Got chatGPT to write some docs for me too!

PR #489, opened 2025-10-17
Writing of reference values

physycola:
I added some basic functionality to allow for the writing of local reference values to an external file. This, along with a `gk_input` file, will mean that one has complete information about the local surface, and so can convert local reference values to physical units. 

I am sure my implementation of `get_reference_values` is quite primitive, as I struggled to find a nice way of doing this that didn't lead to "cyclic dependencies" between units. I am sure you can do something better @bpatel2107. Also, is it worth adding in some round-trip tests to ensure this is working properly?

### DIFF
--- a/src/pyrokinetics/pyro.py
+++ b/src/pyrokinetics/pyro.py
@@ -14,12 +14,12 @@ the following objects:
 from __future__ import annotations
 
 import copy
+import json
 import warnings
 from collections import Counter
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
-import json
 import f90nml
 import numpy as np
 
@@ -1969,17 +1969,17 @@ class Pyro:
             magnetic-field reference (tesla), and minor-radius scale (m); the
             major-radius entry is returned as None when unavailable.
         """
-        
+
         reference_dict = {
             "tref_electron": (1.0 * self.norms.tref).to(self.norms.units.eV),
             "nref_electron": (1.0 * self.norms.nref).to(self.norms.units.m**-3),
             "bref_B0": (1.0 * self.norms.bref).to(self.norms.units.tesla),
             "lref_minor_radius": (1.0 * self.norms.lref).to(self.norms.units.m),
-            "lref_major_radius": None
+            "lref_major_radius": None,
         }
 
         return reference_dict
-    
+
     def write_reference_values(
         self,
         filename: PathLike,
@@ -2040,7 +2040,7 @@ class Pyro:
             if magnitude is None or unit_str is None:
                 kwargs[name] = None
                 continue
-            kwargs[name]  = np.asarray(magnitude).squeeze() * units(unit_str)
+            kwargs[name] = np.asarray(magnitude).squeeze() * units(unit_str)
 
         self.set_reference_values(**kwargs)
 
--- a/src/pyrokinetics/pyro.py
+++ b/src/pyrokinetics/pyro.py
@@ -1971,10 +1971,10 @@ class Pyro:
         """
 
         reference_dict = {
-            "tref_electron": (1.0 * self.norms.tref).to(self.norms.units.eV),
-            "nref_electron": (1.0 * self.norms.nref).to(self.norms.units.m**-3),
-            "bref_B0": (1.0 * self.norms.bref).to(self.norms.units.tesla),
-            "lref_minor_radius": (1.0 * self.norms.lref).to(self.norms.units.m),
+            "tref_electron": (1.0 * self.norms.pyrokinetics.tref).to("eV"),
+            "nref_electron": (1.0 * self.norms.pyrokinetics.nref).to("m**-3"),
+            "bref_B0": (1.0 * self.norms.pyrokinetics.bref).to("tesla"),
+            "lref_minor_radius": (1.0 * self.norms.pyrokinetics.lref).to("m"),
             "lref_major_radius": None,
         }
 
--- a/src/pyrokinetics/pyro.py
+++ b/src/pyrokinetics/pyro.py
@@ -49,7 +49,7 @@ from .normalisation import SimulationNormalisation
 from .numerics import Numerics
 from .templates import gk_templates
 from .typing import PathLike
-from .units import PyroQuantity, PyroNormalisationError
+from .units import PyroNormalisationError, PyroQuantity
 
 if TYPE_CHECKING:
     import xarray as xr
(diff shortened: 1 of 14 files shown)


bpatel2107:
This is a nice bit of functionality. I think this could trip up if the default convention in `self.norms` is not `pyrokinetics` (say when you read in a CGYRO input file) so I think it would be best to explicitly state the convention as pyrokinetics like below

` "tref_electron": (1.0 * self.norms.pyrokinetics.tref).to(self.norms.units.eV),`

I think a few roundtrip tests would be a good addition too.

You can also just pass a string as the units, but it doesn't really matter.

` "tref_electron": (1.0 * self.norms.pyrokinetics.tref).to("eV"),`

daviddickinson:
Looks good. In case this is helpful in testing,  GS2 can store the normalisations in its [input file ](https://gyrokinetics.gitlab.io/gs2/page/namelists/index.html#normalisations_knobs) and output file. I don't know if we would want to support using that as the first source of normalisations / a backup if the extra file isn't present?

bpatel2107:
Pyro will read/write normalisations from the `GKInput` if the code stores it (GS2/GENE store it, CGYRO does not) so it should take precedence. 

This function here would need to be actively called and likely only necessary when not the normalisation is not available. It may be worth add a warning in the scenario we try and set the normalisation when we already have the reference lengths?

src/pyrokinetics/pyro.py
```diff
@@ -1955,6 +1956,98 @@ def set_reference_values(
...
+            magnetic-field reference (tesla), and minor-radius scale (m); the
+            major-radius entry is returned as None when unavailable.
+        """
+
+        reference_dict = {
+            "tref_electron": (1.0 * self.norms.pyrokinetics.tref).to("eV"),
+            "nref_electron": (1.0 * self.norms.pyrokinetics.nref).to("m**-3"),
+            "bref_B0": (1.0 * self.norms.pyrokinetics.bref).to("tesla"),
+            "lref_minor_radius": (1.0 * self.norms.pyrokinetics.lref).to("m"),
+            "lref_major_radius": (
```
physycola, line 2031: This latest change to `get_reference_values` may not be correct; one probably needs to use the aspect ratio to convert from the minor radius to the major radius? @bpatel2107

physycola: That is, the method of extracting `lref_major_radius` may need to be changed to reverted to return `None`.

physycola:
There also may be something subtle going on here with normalisations that I do not quite understand. Suppose that I have a PYRO input file and some reference values written with the added `write_reference_values` method. Then, I would expect the following code to execute without error:

```
pyro = pk.Pyro(gk_file=<path to input file>)
pyro.read_reference_values(<path to reference values>)
a_minor = pyro.local_geometry.a_minor.to("m")
```

However, this gives the usual error that one encounters when trying to convert with only a local object loaded, i.e., 

```pint.errors.DimensionalityError: Cannot convert from 'lref_minor_radius' ([lref]) to 'meter' ([length])```

Changing the last line to 

```a_minor = pyro.local_geometry.a_minor.to("meter", pyro.norms.context)```

cures this error, though one still encounters issues with, e.g.,

```(1 / pyro.local_geometry.dpsidr * (1 * pyro.norms.bref)).to(pyro.norms.rhoref**-1, pyro.norms.context)```

However, it appears that the context should already be extablished in the `set_all_references` function called by `read_reference_values`, so I would expect these conversions to work. Can someone who understands the units a bit more than me shed some light on this? @bpatel2107 @ZedThree.

**Comment by bpatel2107** [comment]
Down the rabbit hole I went. Essentially pyro has "physical" units that can be converted to SI and "simulation" units which cannot. 

When loading data in from a GK input file all the units assigned will be given simulation units. It is not possible to convert between simulation and physical units unless you have a "context" defined where this transformation is known.

This often led to situations where you could not convert between them as you need to do something like 

`(1 * pyro.norms.units.lref_minor_radius).to("meter", pyro.norms.context)`

rather than

`(1 * pyro.norms.units.lref_minor_radius).to("meter")`

And the error in the latter case is often confusing.

I have modified the unit conversion to do the following

- Check if the input and output units are physical, simulation or a mixture
- If all units are simulation then proceed with the conversion
- If all units are physical then proceed with the conversion
- If there is a mixture then try to convert but if it fails raise an error with a more useful message

There were some issues with the conversion of `[beta_ref]` units as "physical"  units were not being generated (no `beta_ref_ee_B0_test0000`) so I've added that in.

I have also added a `.to()` method to convert the whole pyro object to the specified unit convention changing `LocalSpecies`, `LocalGeometry`, `Numerics` and the `GKOutput`.


So when setting all the reference values we now also convert everything from simulation units to physical units. This is optional and set by the flag `convert_pyro`. This then lets you convert to SI without needing a context. 

The code below highlights all of this and the final (expected) error you get is much more useful

`pyrokinetics.units.PyroContextError: Trying to convert between physical and simulation units '1.0 lref_minor_radius' -> 'm' without context. Please use a context here`

```python
import os
import pathlib
from typing import Union


def main(base_path: Union[os.PathLike, str] = ".", geometry_type: str = "Miller"):
    # Equilibrium file
    eq_file = template_dir / "test.geqdsk"

    # Kinetics data file
    kinetics_file = template_dir / "jetto.jsp"

    # Load up pyro object
    pyro = Pyro(
        eq_file=eq_file,
        kinetics_file=kinetics_file,
        kinetics_type="JETTO",
        kinetics_kwargs={"time": 550},
    )

    # Generate local parameters at psi_n=0.5
    pyro.load_local(psi_n=0.5, local_geometry=geometry_type)

    # Select code as CGYRO
    pyro.gk_code = "CGYRO"

    base_path = pathlib.Path(base_path)

    # Write CGYRO input file using default template
    pyro.write_gk_file(file_name=base_path / "test_jetto.cgyro")
    
    # Save and write initial reference values
    ref_values = pyro.get_reference_values()
    output_file = "references.cgyro"

    pyro.write_reference_values(output_file)

    new_pyro = Pyro(gk_file="test_jetto.cgyro")
    # Set to reference values - convert without context will work
    new_pyro.set_reference_values(**ref_values, convert_pyro=True)

    amin = new_pyro.local_geometry.a_minor

    print(amin)
    print(amin.to("m", new_pyro.norms.context))
    print(amin.to("m"))

    new_pyro_no_convert = Pyro(gk_file="test_jetto.cgyro")
    # Set to reference values - convert without context will not work
    new_pyro_no_convert.set_reference_values(**ref_values, convert_pyro=False)

    amin = new_pyro_no_convert.local_geometry.a_minor

    print(amin)
    print(amin.to("m", new_pyro_no_convert.norms.context))

    # This will fail!
    print(amin.to("m"))


if __name__ == "__main__":
    main()
```

bpatel2107 (approved):
I am now reasonably happy with this!

PR #490, opened 2025-10-20
Example script to generate global GENE profiles

bpatel2107:
Here is an example script to generate the profiles for GENE global simulations.

I am not yet convinced we should supply anything more than this for now as we don't have a good interface with global simulations yet but eventually we would want something like input file writers for each global code but we are a ways off that.


dake0795 (approved):
Looks good to me.

dake0795 (changes requested):
Will need changing to use a template

bpatel2107:
Switched to template JETTO profile and GEQDSK

PR #491, opened 2025-10-23
Flip if statement

bpatel2107:
Small bug in `load_numerics` breaking the `example_modify_shear.py' script


MantasA411 (approved):
This works and solves the issue.

PR #493, opened 2025-10-24
Enhancements/complex netcdf interface

bpatel2107:
This utilises the `auto_complex=True` feature in xarray Dataset so we don't need to do the hacky `ReIm` coordinate thing we were doing before. This also means we have use the base DatasetWrapper `to_netcdf` method for all the output objects now.

I've updated the test data as that originally had the `ReIm` axis we don't need anymore.


ZedThree (approved):
You'll want to also bump the `netCDF4` and `xarray` versions in `pyproject.toml` to those that support `autocomplex` (which I think are `1.7.6` and `v2024.10.0` respectively)

PR #494, opened 2025-10-24
Feature/fpsi attribute

MantasA411:
Added attributes Fpsi and Fprime to local_geometry.


bpatel2107:
So I think the `LocalGeometry` object stores `FFprime` rather than `Fprime` so it would make sense to switch to that.

bpatel2107 (approved):

PR #497, opened 2025-10-28
Update parameters

FelixWattsYork:
Broke out updating pyro objects in a pyroscan into a separate method so it can be called independently, for example for when you read a pyroscan object and what to be able to reference the input values for that pyroscan, for say, a Gaussian process model


bpatel2107 (approved):
LGTM

PR #498, opened 2025-10-29
Use correct defn of Cy

bpatel2107:
The calc of `k_perp` was using a slightly incorrect defn of $C_y = r/q$ where $y = C_y \alpha$, which is not consistent with what we assume which is $C_y= \frac{\partial \psi}{\partial r} \frac{1}{B_0}$ which led to factors of $B_{unit}/B_0$ all over the place where it shouldn't be so here we correct that. This comes up in the use of the MG saturation rule.


physycola (approved):
Looks good to me. The sign of `C_y` is consistent with the definition of `alpha` used in the metric terms.

PR #499, opened 2025-11-03
Add context to normalisation deepcopy and to consistent_beta_prime wi…

bpatel2107:
…th test

Previously using `deepcopy` on the `Pyro` object would go through and copy over all the contained objects but when copying over  the `Normalisation` `Context` was not being included in the copy so converting between `bref_B0` and `bref_Bunit` for example was not working.

Also adds in a missing context for `enforce_consistent_beta_prime`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ dependencies = [
     "freeqdsk ~= 0.5.0",
     "cleverdict >= 1.9.1",
     "xarray >= 2024.10.0, <=2024.11.0",
-    "pint ~= 0.23",
+    "pint >= 0.23, <=0.24.4",
     "pint-xarray == 0.4",
     "contourpy ~= 1.0",
     "xrft >= 1.0.0",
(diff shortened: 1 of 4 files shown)


bpatel2107:
So the check in `enforce_consistent_beta_prime` was too strict as you can have cases where `beta` is set in the `Normalisation` but not yet set in `Numerics` so now we check both.

Also now when we call `_load_local_geometry_species_dependency` actually sets `beta` in `Numerics` rather than leaving it as None. When `numerics.beta = None` then when writing the input files it would default back to `pyro.norms.beta` but that seems a bit dangerous if we have a `Numerics` object.

Also we added a method `load_template_file` which loads in a template file populating the `GKInput` and `Numerics` object (assuming `Numerics` doesnt already exist). When loading data from the template file, the `beta` and `gamma_exb` will still be taken from `LocalSpecies`/`LocalGeometry`/`Normalisation`

physycola (approved):
Appears to be working well. When constructing a `pyro` object from a combination of sources (equilibrium and GK numerics), I would recommend the following:

1. Load `equilibrium` object
2. Load `local_species` and `local_geometry`
3. Manipulate `local_species` and `local_geometry` in any way
4. Load `numerics` using `load_template_file()`
5. Call any consistency checks (e.g., `enforce_consistent_beta_prime()`)

PR #500, opened 2025-11-06
Add norms to get_coords and switch phase when using n0_global

bpatel2107:
Loading the `GKOutput` for GENE with a `tracer_efit` run was not working as the normalisation needed to be passed to the `_get_coords` so this addresses that.

Also there may be a incorrect factor in phase used to map to ballooning space but I'm not sure what it should be. @dake0795 any ideas.


ILexiconI (approved):
This solved the issue I was having with loading GENE outputs.  Thanks!

PR #501, opened 2025-11-06
Add to method to metric, force gs2 units for IBM and get_gs2_geometry

bpatel2107:
The IBM calculation uses `MetricTerms` which will change depending the choice of units. This propagates through to the IBM calc via `get_gs2_geometry`. In the IBM calc itself we drop the units so depending on which code you start from you will get a different answer (scaled by some constant factor). Here I force the units to be in GS2 units so we always get the same value.

We also force the `MetricTerms` calculation to use a completely consistent `dpsidr` calculate with `r`, `q` and `Jacobian` rather than the `dpsidr` that was put into `LocalGeometry` (which is set before things like `q` can be changed). This ensure that changes to `q` are appropriately calculated.

### DIFF
--- a/src/pyrokinetics/diagnostics/__init__.py
+++ b/src/pyrokinetics/diagnostics/__init__.py
@@ -268,6 +268,7 @@ class Diagnostics:
         self.pyro.load_metric_terms(theta=theta_even)
 
         metric = self.pyro.metric_terms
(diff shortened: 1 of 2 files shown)


MantasA411 (approved):
This has fixed the problem of changing q self consistently, changing q changes the dpsidr term (and the IBM calculation) self consistently now.

PR #503, opened 2025-11-07
Bugfix/gene adiabatic electrons

physycola:
Some bugfixes for loading CGYRO adiabatic electron cases. This isn't yet ready for merging though, as I seem to get failures when selecting a particular `output_convention` when loading the data. Not really sure what is causing this, as I believe all the references are set.

### DIFF
--- a/src/pyrokinetics/gk_code/gene.py
+++ b/src/pyrokinetics/gk_code/gene.py
@@ -1721,7 +1721,7 @@ class GKOutputReaderGENE(FileReader, file_type="GENE", reads=GKOutput):
             species_names = [gk_input.data["species"]["name"]]
         else:
             species_names = [species["name"] for species in gk_input.data["species"]]
-        
+
         geometry_type = gk_input.data["geometry"]["magn_geometry"]
         files = cls._get_gene_mom_geo_files(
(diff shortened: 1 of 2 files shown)


physycola:
This appears to be working correctly now, without the conversion bugs when trying to change norms.

src/pyrokinetics/gk_code/gene.py
```diff
@@ -384,18 +384,24 @@ def get_local_geometry(self) -> LocalGeometry:
...
-            amhd_beta_prime = (
-                -local_species.inverse_lp.m
-                * local_species.pressure.m
-                / local_geometry_data["B0"] ** 2
-            )
+            if local_geometry_data["B0"] is not None:
+                amhd_beta_prime = (
+                    -local_species.inverse_lp.m
+                    * local_species.pressure.m
+                    / local_geometry_data["B0"] ** 2
```
bpatel2107: Thinking about this, we could just multiple it by `beta` rather than dividing by `1/B0^2` as `beta` is always set even if not there

bpatel2107:
Will fix the `B0` issue elsewhere

PR #506, opened 2025-11-13
Added functionality to store the runfile_dict in a pyroscan in json, and load it back at a later point. no a breaking change but default dictionaty format for runfile_dict is now a string to allow for json saving

FelixWattsYork:
Previously if you set a custome runfile_dict for a pyroscan it wouldn't be scanned and therefore the pyroscan wouldn't correctly to the gk outputs. I changed the json saving to save the runfile_dict and load it. Since Json doesn't allow storage of tuples I created code to change the runfile_dict key to a string. This string can be given explicitly or the old tuple style format can be given and it is converted into a string

### DIFF
--- a/src/pyrokinetics/pyroscan.py
+++ b/src/pyrokinetics/pyroscan.py
@@ -112,9 +112,7 @@ class PyroScan:
                             value[param_key] = param_value[0] * ureg(param_value[-1])
                         else:
                             value[param_key] = param_value[:]
-                elif key == "runfile_dict":
-                    self.runfile_dict = value  # I think this logic is correct?
-                elif (
+                if (
                     key == "base_directory"
                     and base_directory != "."
                     and base_directory != value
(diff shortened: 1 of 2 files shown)


src/pyrokinetics/pyroscan.py
```diff
@@ -110,7 +112,9 @@ def __init__(
                             value[param_key] = param_value[0] * ureg(param_value[-1])
                         else:
                             value[param_key] = param_value[:]
-                if (
+                elif key == "runfile_dict":
+                    self.runfile_dict = value  # I think this logic is correct?
+                elif (
```
bpatel2107: I would think this would be covered by the `setattr(self, key, value)` later on?

bpatel2107 (approved):
I think this is fine but I think the `if` statement in the loop over the json file is unnecessary.

FelixWattsYork:
Added more tests and fixed issues that Bhavin raised. First proper PR yay!

PR #507, opened 2025-11-14
Fix GENE MXH setting of s_delta/s_zeta and correct Bgeo/B0 calc

bpatel2107:
Two GENE related bugs here

- Apparently the values of  `delta`, `s_delta`, `zeta` and `s_zeta` take precedence over the shape coefficients so now we treat them explicitly. The previous implementation had those derivatives subtly wrong as the internal definitions of `s_delta`/`s_zeta` do not match the GENE definitions. I have also removed them entirely from the defaults in from `pyro_gene_mxh` and `pyro_gene_mxh_default` given the explicit treatment
- When loading a `tracer_efit` run we would need to map from the `Bref` used in GENE and $B0 = f/Rmajor$. Previously I was using the output from the tracer_efit file to work out $f$ in the GENE units. However the output I was using had $|B|$ rather what I thought was $B_\phi$ so I was actually calculating $|B| R_{major}$ rather than $B_\phi R_{major}$ so I now work out the toroidal field and then use that for the determination of $f$.

### DIFF
--- a/src/pyrokinetics/gk_code/gene.py
+++ b/src/pyrokinetics/gk_code/gene.py
@@ -333,7 +333,7 @@ class GKInputGENE(GKInput, FileReader, file_type="GENE", reads=GKInput):
 
         trpeps = self.get_trpeps()
 
-        # Handle the fact that GENE defaults to 'minor_r = 1.0' in 
+        # Handle the fact that GENE defaults to 'minor_r = 1.0' in
         # 'parameters.dat' if 'minor_r' is not specified in 'parameters'
         minor_r = self.data["geometry"].get("minor_r", 0.0)
         if not (minor_r == 0.0 or np.isclose(minor_r, major_R)):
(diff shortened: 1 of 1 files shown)


physycola (approved):
These changes look good. Currently is able to handle the slightly-odd defaults that GENE uses for `parameters.dat`.

physycola:
I have not, however, looked at the `EFIT` stuff; can I leave that up to you @ILexiconI ?

ILexiconI:
> I have not, however, looked at the `EFIT` stuff; can I leave that up to you @ILexiconI ?

Sure, I've been looking and think it might be messing with setting the normalisation convention somehow, will update when I've figured out what's going on.

ILexiconI:
The issue I was encountering above originated from a different PR and is unrelated.

This PR fixes GENE misinterpreting the (self-conflicting) MXH geometry input files that pyrokinetics was producing previously.
Implementation of the f calculation looks fine.

PR #508, opened 2025-11-18
Added missing Bgeo option

ILexiconI:
This fixes a bug introduced in PR501 that was causing some of my codes to fail.  "Bgeo", in addition to "B0" and "Bunit", is a normalisation that can be available for Bref.

This fixes the bug and I think treats the Bgeo case correctly, but do correct me if I'm wrong on the latter @bpatel2107 .
I'm directing the review to d7919 as bpatel2107 is currently on Leave.

### DIFF
--- a/src/pyrokinetics/local_geometry/metric.py
+++ b/src/pyrokinetics/local_geometry/metric.py
@@ -145,7 +145,7 @@ class MetricTerms:  # CleverDict
         bref_units = local_geometry.dpsidr.units / local_geometry.rho.units
 
         # Needs to explicitly be 1 * B0 regardless of units
-        if "Bunit" in str(bref_units):
+        if "Bunit" or "Bgeo" in str(bref_units):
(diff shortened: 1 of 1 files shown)


d7919:
I'm not really familiar enough with this bit of the code to review I think. Perhaps @MantasA411 or @hgd511 might be able to comment?

MantasA411:
I also don't have a good enough understanding of the effect of normalisation here to review it, specifically whether `Bgeo` needs to be adjusted similarly to `Bunit ` by a factor of `B0/Bgeo`. 

The tests are failing for cases where `B0 ` is used because, I think, line 150 should be

`elif "B0" not in str(bref_units) and "Bgeo" not in str(bref_units):`

ILexiconI:
The second commit was more of a check really.  Yeah, I think I need to just add a second elif to the statement to handle the Bgeo case separately, should be doable.

ILexiconI:
Fixed in PR513

PR #511, opened 2025-12-03
More robust handling of psi_n_lcfs

lorenzozanisi:
- Mismatch between `psi_n_lcfs` from upstream and `psi_grid` and `psi_RZ` from IMAS was causing issues at psi close to 1, where `r_minor` would become non monotonic. This has been fixed by rescaling`psi_n_lcfs` 
- added try/except statement in case fast ion data is not available


bpatel2107:
This check was too aggressive and the adjustment to `psi_n_lcfs` is only needed if the new value is close to 1.0. Essentially this comes down to whether or not the equilibrium data outside the LCFS is valid or not. If it is valid then then check shift should not be done as the min/max `psi` could be anywhere. If the data outside the LCFS isn't valid then you may have the case where the data there has not been set well and you need to pull in the LCFS just a little bit to get a smooth contour.

This is now fixed and ready to merge

PR #513, opened 2025-12-10
Bugfix/local geometry b0

bpatel2107:
The `LocalGeometry` object has a variable `B0` which in the docs is defined as `f/Rmaj` (is a flux function).

However, due to an old implementations of reading `GKInput` from input files, a hacky `B0 = 1 / sqrt(beta)` was used which is not really needed anymore as we handle `beta` separately in `Numerics` (this was implemented before the `Numerics` object properly existed).

This means that the value of `B0` wasn't really `B0`. This is now fixed for each code. This results in the sometimes trivial with a definition of `local_geometr y.B0 = 1.0 bref_B0` but it is useful when using different normalisations and allows `MetricTerms` to work with any normalisation now which requires by `B0` (addressing #508). Added some simple test checking this and including `B0` (and `Fpsi`) in the roundtrip tests.

The value of `B0` wasn't really used anywhere so this won't affect functionality for 99% of cases but essentially makes `MetricTerms` much more robust against choice of normalisation.

### DIFF
--- a/docs/examples/example_gs2.py
+++ b/docs/examples/example_gs2.py
@@ -2,32 +2,9 @@ import pyrokinetics
 
 gs2_template = pyrokinetics.template_dir / "input.gs2"
 
-pyro = pyrokinetics.Pyro(gk_file=pyrokinetics.template_dir / "input.gs2")
-print(pyro.local_geometry.B0)
+pyro = pyrokinetics.Pyro(gk_file=gs2_template, gk_code="GS2")
 
-
-pyro = pyrokinetics.Pyro(gk_file=pyrokinetics.template_dir / "input.stella")
-print(pyro.local_geometry.B0)
-
-pyro = pyrokinetics.Pyro(gk_file=pyrokinetics.template_dir / "input.gx")
-print(pyro.local_geometry.B0)
-
-pyro = pyrokinetics.Pyro(gk_file=pyrokinetics.template_dir / "input.gkw")
-print(pyro.local_geometry.B0)
-
-pyro = pyrokinetics.Pyro(gk_file=pyrokinetics.template_dir / "input.cgyro")
-print(pyro.local_geometry.B0)
-
-pyro = pyrokinetics.Pyro(gk_file=pyrokinetics.template_dir / "input.gene")
-print(pyro.local_geometry.B0)
-
-pyro = pyrokinetics.Pyro(gk_file=pyrokinetics.template_dir / "input.tglf")
-print(pyro.local_geometry.B0)
-exit()
 flags = {
-
-
-
     "gs2_diagnostics_knobs": {
         "write_fields": True,
         "write_kpar": True,
--- a/docs/examples/example_gs2.py
+++ b/docs/examples/example_gs2.py
@@ -2,9 +2,32 @@ import pyrokinetics
 
 gs2_template = pyrokinetics.template_dir / "input.gs2"
 
-pyro = pyrokinetics.Pyro(gk_file=gs2_template, gk_code="GS2")
+pyro = pyrokinetics.Pyro(gk_file=pyrokinetics.template_dir / "input.gs2")
+print(pyro.local_geometry.B0)
 
+
+pyro = pyrokinetics.Pyro(gk_file=pyrokinetics.template_dir / "input.stella")
+print(pyro.local_geometry.B0)
+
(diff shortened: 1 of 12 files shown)


docs/examples/example_gs2.py
```diff
@@ -2,9 +2,32 @@
 
 gs2_template = pyrokinetics.template_dir / "input.gs2"
 
-pyro = pyrokinetics.Pyro(gk_file=gs2_template, gk_code="GS2")
+pyro = pyrokinetics.Pyro(gk_file=pyrokinetics.template_dir / "input.gs2")
+print(pyro.local_geometry.B0)
```
d7919: I can see why you would add this for checking the changes when developing, but I'm not sure it still makes sense to do this in the final PR state in an example file.

docs/examples/example_gs2.py
```diff
@@ -2,9 +2,32 @@
...
 
+
+pyro = pyrokinetics.Pyro(gk_file=pyrokinetics.template_dir / "input.stella")
```
d7919: I'm guessing and the below this was not intended to be committed / pushed as clearly not example_gs2 material! (and there's an exit below which is not what you want). Basically should undo all the changes to this file?

docs/examples/example_gs2.py
bpatel2107: Ah that's my bad, will revert the file!

src/pyrokinetics/gk_code/gs2.py
```diff
@@ -1244,8 +1234,9 @@ def _get_coords(
         raw_theta = raw_data["theta"].data
 
         if gk_input.data["theta_grid_eik_knobs"].get("equal_arc", True):
-
+            print(gk_input)
```
d7919: Remove?

src/pyrokinetics/gk_code/gs2.py
```diff
@@ -1244,8 +1234,9 @@ def _get_coords(
...
             local_geometry = gk_input.get_local_geometry()
+            print("LG after", local_geometry)
```
d7919: Remove?

src/pyrokinetics/local_geometry/metric.py
```diff
@@ -142,15 +142,7 @@ def __init__(self, local_geometry: LocalGeometry, ntheta=None, theta=None):
...
-        bref_units = local_geometry.dpsidr.units / local_geometry.rho.units
-
-        # Needs to explicitly be 1 * B0 regardless of units
-        if "Bunit" in str(bref_units):
-            bref_units *= 1.0 / local_geometry.bunit_over_b0
-        elif "B0" not in str(bref_units):
-            raise ValueError("Need to convert to a standard normalisation first")
-
-        self.dpsidr = self.Y * local_geometry.Rmaj / self.q * bref_units
+        self.dpsidr = self.Y * local_geometry.Rmaj / self.q * local_geometry.B0
```
d7919: Brackets around  mid-statement division to improve clarity of intent?

bpatel2107: Do you mean?

`self.dpsidr = self.Y * (local_geometry.Rmaj / self.q) * local_geometry.B0`

d7919: Yes. I know this isn't everyone's preferred style, but having marked a lot of student programming assignments I know that some people expect A / B * C to be equivalent to A / (B * C), or at least it is easy to not spot the difference.

bpatel2107: Changed!

src/pyrokinetics/gk_code/gs2.py
bpatel2107: Removed

src/pyrokinetics/gk_code/gs2.py
bpatel2107: Removed

docs/examples/example_gs2.py
bpatel2107: Reverted

docs/examples/example_gs2.py
bpatel2107: Reverted

ILexiconI:
Sorry forgot to actually merge this, but it did fix my issue thanks.

PR #514, opened 2025-12-10
PyroScan Error Handling

physycola:
Currently, the loading of pyroscan data may fail due to a `ValueError`, which is encountered if there is an array-size mismatch when, e.g., reshaping data, and is likely due to some I/O failure of the code. To guard against this, I've added handling of `ValueError`, as well an appropriate warning.


bpatel2107 (approved):
LGTM

PR #515, opened 2026-01-12
Downsample GENE data before loading into memory

bpatel2107:
Similar to #484 

Previously Pyro would try to load all the data from the fields in the binary files using np.frombuffer. When working with nonlinear simulations where the output can be 100GB you can often run out of memory and the process will fail. It was not as severe as the CGYRO as it would load data in each time step at a time.

This PR switches to using np.memmap in GENE, (similar to how CGYRO currently works) which just puts the location of the data in memory. This allows for downsampling before the data is loaded, which significantly reduces the time taken to read in the data.

To do this you need to add a dictionary downsample as a kwargs when calling pyro.load_gk_output

```
from pyrokinetics import Pyro

gk_file = "parameters"
pyro = Pyro(gk_file=gk_file)

downsample = {
    "time": slice(-30, None, None),
    "ky": slice(1, 10, None),
    "theta":  slice(13, 17, None),
    "kx": slice(20, 40, None),
}

pyro.load_gk_output(downsample=downsample)
```

The above example will only load the final 100 timeslices, the kys going from index 1 to 10, the thetas from index 13 to 17, and the kx from index 20 to 40. One can specify a single index using an integer or a range of indices using slice

Doing this I also found two bugs, both involving the `kx` grid.

Firstly, in nonlinear GENE I had thought that the `kx` grid with an even `nkx`was set up such that it would be 

`-3, -2 -1, 0, 1, 2`

but it was actually

`-2, -1, 0, 1, 2, 3`

The "extra" point is set to 0 by GENE but it meant before we were rolling the array when we didn't need to. Functionally this made no difference for the output of nonlinear runs.

Secondly for linear runs with an odd `nkx` we were accidentally dropping the final `kx` but that is not forced to 0 so is needed. The `kx` grid should be 

`-2, -1, 0, 1, 2`

We of course do a ballooning transform so we only get 1 `kx` but it meant we were missing the tail end of the eigenfunction.

The test happens to have an odd number for `nkx` so the results needed to change. 

For converged linear runs I expect this to make little difference but the test data isn't very good so it changed quite a bit...


bpatel2107:
Add some docs on this, made a lot easier with ChatGPT

I suspect many docs improvements may be on the way...

Issue #116, opened 2022-08-22
Add support for neoclassical codes?


d7919:
How difficult might it be to add support for neoclassical codes like neo and sfincs?

I'd expect that at least the local mode of neo shouldn't be too different from CGYRO, for example.

bpatel2107:
NEO would be trivial but I have never heard of sfincs. Depends on the inputs but if it has similar input then I don't see why not. 

This does stray away a little from the purpose of pyrokinetics as I imagine the outputs would look very different for normal GK runs. But its definitely possible

d7919:
Here's sfincs https://github.com/landreman/sfincs, it can be used with stella and we're working on using neo with GS2 (the old links between the codes don't work due to changes in the neo outputs) so it would be helpful to have everything under one interface so that we can more readily ensure consistency in the inputs etc.

d7919:
Bumping this as more people working with GK+NC in various forms.

PR #516, opened 2026-01-12
Write NEO input files

bpatel2107:
This PR adds the ability to read/write in NEO input files as it fits reasonably neatly into `GKInput` (even though it isn't "G"). Partially addresses #116.

I've added it to some roundtrip tests and that seems to work fine. In terms of more manual checks I compared the trapped particle fraction spit out by NEO and from Pyro which agree pretty well so I think at least the geometry is being set correctly!

image

The inclusion of `GKOutput` needs a bit more thinking. Things like neoclassical fluxes have a good enough equivalent but things like 1st order potentials/flows/currents may be more useful outputs. This is relatively straightforward to add but the structure may need thinking (does `xr.DataArray`) still make sense with so little data.. probably).

This could be merged as is but I leave it up to everyone on whether we want the outputs in now or deal with it later


bpatel2107:
Will merge now and deal with outputs as needed

PR #520, opened 2026-01-16
Bug fix/gene minor r check

FelixWattsYork:
Fixes Minor Radius if set incorrectly.
If Minor Radius is equal to trpeps * Major radius it is meant to be the reference value but is not set as one. This sets it to one in this situation whilst raising a warning


src/pyrokinetics/gk_code/gene.py
```diff
@@ -1136,6 +1135,10 @@ def _detect_normalisation(self):
         magnetic_axis_radius = None
         minor_radius = self.data["geometry"].get("minor_r", 0.0)
         major_radius = self.data["geometry"]["major_r"]
+        trpeps = self.data["geometry"]["trpeps"]
+        if trpeps * major_radius != minor_radius:
+            minor_radius = 1.0
+            raise warnings.warn("minor_r incorrectly set, setting to 1", UserWarning)
```
bpatel2107: This logic should be reversed

`
```suggestion
        if trpeps * major_radius == minor_radius:
            minor_radius = 1.0
            raise warnings.warn("minor_r incorrectly set, setting to 1", UserWarning)
```

Keep in mind that `trpeps = r/Lref` and `major_radius = Rmaj / Lref` so it is extremely unlikely that 

`r * Rmaj / Lref**2 = a_minor` which is what we are trying to catch here. This scenario is not impossible by very unlikely.

PR #521, opened 2026-01-27
Added ability to read Nonlinear Data in pyroscans

FelixWattsYork:
Updated pyroscans to allow them to read nonlinear flux data. For TGLF this works by simply taking the output flux data, however for Gyro kinetic simulations like GENE this works by averaging over a period of time steps. 

This PR also includes a number of other useful features:
- Ability to pass flags to load fluxes and fields through pyroscans. Whilst there is little point in passing fields you can now get field values over a range
- Ability to save pyroscan JSON with relative as opposed to absolute paths. This allows pyroscans to be moved between machines and still be easily readable, improving reproducibility. This can be toggled with the flag Relative_Path = True for the write method of a pyroscan. 
- Improved testing of both pyro reading this flags and reading pyroscans, although it could do with a increased test base this comes with the trade off of requiring more storage

Future work
- Quasi Linear weights reading from TGLF and propagation through pyroscans. I current have a branch that does this, but I need to properly test and merge it
- Ability to save base pyro object and read from it. This is currently a major problem with the reproducibility of pyroscans. You can't read a pyroscan without giving it a base pyro object which is uses for a large number of it operations. This base pyro object contains important information like the units. Using an input file from the pyroscan is not always a suitable replacement. There exists functionality to store data from a pyro object in JSON, but this needs to be implemented as an option for the base pyro in pyroscan.

---Update----

Now that branch is as ready for Merge as I can get it. I have made some addition test and implemented some more important features, namely
- Loading purely from pyroscan. Now by using the load_base_pyro flag, pyroscan will use the input file of the first point in the pyroscan as the new base pyro object.
- Improved testing of pyroscans with files missing
- Drastically reduced code size and improved readability
- Zipped TGLF nonlinear testing file

The main issue with this PR is that it lacks testing for reading NON-linear GK. It should work but I haven't tested it yet. Given however that the current branch of Pyro will definitely fail much less gracefully in the same situation, I think it best to merge now and then get to work on that feature when I have access to nonlinear test data

As a special plea I would request that If you use Pyro and conduct linear or nonlinear scans to please try and use Pyroscans. Now with relative paths and loading completely from the JSON, a pyroscan is fully reproducible and completely portable. This would make it incredibly easy to recreate analysis with different codes (making my life much easier)


src/pyrokinetics/pyroscan.py
```diff
@@ -190,7 +211,13 @@ def create_single_run(self, parameters: dict):
         new_run.run_parameters = copy.deepcopy(parameters)
         return name, new_run
 
-    def write(self, file_name=None, base_directory=None, template_file=None):
+    def write(
+        self,
+        file_name=None,
+        base_directory=None,
+        template_file=None,
+        Relative_Path=False,
```
bpatel2107: fix case

src/pyrokinetics/pyroscan.py
```diff
@@ -427,12 +471,18 @@ def load_gk_output(
...
+                        load_fluxes=load_fluxes,
+                        load_moments=load_moments,
+                        drop_nan=drop_nan,
+                        **kwargs,
+                    )
 
                     if "mode" in pyro.gk_output.dims:
                         growth_rate.append(pyro.gk_output["growth_rate"])
                         mode_frequency.append(pyro.gk_output["mode_frequency"])
-                        eigenfunctions.append(pyro.gk_output["eigenfunctions"])
```
bpatel2107, line 435: add back

src/pyrokinetics/pyroscan.py
```diff
@@ -565,7 +621,420 @@ def units_reshape(array, shape):
...
+                    if "bpar" in pyro.gk_output.data_vars:
+                        bpar.append(pyro.gk_output["bpar"])
+                    if "apar" in pyro.gk_output.data_vars:
+                        apar.append(pyro.gk_output["apar"])
+
+                # Remove GKOutput to conserve memory
+                ky_length = len(pyro.gk_output["ky"])
+                pyro.gk_output = None
+
+                # #except (FileNotFoundError, OSError, IndexError, RuntimeError, KeyError):
```
bpatel2107: try except

FelixWattsYork:
Fixed black issues

FelixWattsYork:
Added additional testing for reading nonlinear pyroscans, and improved the documentation surrounding reading and writing pyroscans. Changed pyroscans so that they create a base template input file which can be read to create the pyroscan. Json template file now automatically use a relative path to make generated pyroscans far more portable

docs/examples/example_create_pyroscan.py
```diff
@@ -0,0 +1,18 @@
...
+pyro = Pyro(gk_file=gk_file)
+
+param_key = "ky"
+param_value = np.array([0.1, 0.2, 0.3])
+param_dict = {param_key: param_value}
+
+# Create PyroScan object
+pyro_scan = PyroScan(pyro, param_dict, base_directory=base_directory)
+
+pyro_scan.write(file_name="input.cgryo", template_file=None)
```
bpatel2107: This typo is probably why there are `input.cgryo` files lying around

bpatel2107:
I think this needs a little tidying as there are several rogue files lying around (uv.lock, input.cgryo etc.) I can check functionality when that is fixed up

docs/examples/example_create_pyroscan.py
bpatel2107: Bump

tests/gk_code/test_gk_output_reader_gx.py
```diff
@@ -84,13 +84,9 @@ def test_infer_path_from_input_file_gx():
 
 @pytest.mark.parametrize(
     "load_fields",
-    [
-        True,
-        False,
-    ],
+    [True],
```
bpatel2107: Why has this been changed? A few of the tests on reading the outputs have been tweaked?

bpatel2107 (approved):
Happy with this!

tests/gk_code/test_gk_output_reader_gene.py
```diff
@@ -175,12 +175,9 @@ def test_data_attrs(self, attr):
 
 @pytest.mark.parametrize(
     "load_fields",
-    [
-        True,
-    ],
+    [True, False],
```
bpatel2107: @FelixWattsYork This is causing the test to fail. Take out the False here

PR #523, opened 2026-01-29
Add eq.upscale() to increase R–Z grid resolution via interpolation

MantasA411:
Adds a method on the equilibrium object `pyro.eq.upscale(R_upscale, Z_upscale)`, it increases the resolution of the equilibrium `psi_RZ`. The method constructs a finer R–Z grid (scaled by the requested factors) and interpolates the existing `psi_RZ` values onto it.


bpatel2107:
Worth adding a test in?

bpatel2107 (approved):
LGTM

PR #524, opened 2026-02-13
Feature/pyro net cdf

FelixWattsYork:
Simple PR to add the ability to load from netcdf to pyro gk outputs in the same manner you can load pyroscan gk outputs

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -38,6 +38,8 @@ jobs:
     - name: Run isort
       run: |
         isort src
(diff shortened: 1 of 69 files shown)


bpatel2107:
I've tweaked the netcdf reading/writing example so it doesn't write to the pyrokinetics directory by default. Also added some docs (generated by ChatGPT) on how to use the netCDF writing for linear/nonlinear runs

bpatel2107 (approved):

PR #525, opened 2026-02-16
Feature/flux surface integral

FelixWattsYork:
Load GK_Output now loads the Jacobian (in terms of R)  for each theta for Gyrokinetic simulations. This allows one to accurately integrate over a flux surface, as can be necessary for field data


bpatel2107 (commented):
I don't think we should pass an xarray dataset inside each individual gk_code file as the other data (fields, fluxes, moments) aren't treated like that.

I think we just pass it a numpy array of the Jacobian and let `gk_output.py` convert it to the appropriate format. Otherwise we add a bunch of xarray imports that aren't really necessary

src/pyrokinetics/gk_code/cgyro.py
```diff
@@ -1332,6 +1334,29 @@ def _get_coords(
...
+
+        # Strip units safely for interpolation
+        J_units = Jacobian_raw.data.units
+        J_mag = Jacobian_raw.data.m
+
+        # Interpolate magnitudes only
+        J_interp_mag = np.interp(theta, Jacobian_raw.theta.values, J_mag)
+
+        # Reattach units
+        J_interp = xr.DataArray(
```
bpatel2107: We should just pass `J_interp_mag` directly to the dictionary rather than a `Dataset`

We can also just call it Jacobian here. All this xarray stuff here seems unnecessary

bpatel2107:
Added in #535 so can close now

PR #528, opened 2026-02-17
Add back Z_EFF/Z_EFF_METHOD if legacy

bpatel2107:
Add `Z_EFF` and `Z_EFF_METHOD` when writing files in legacy format (this is the default at the moment)


physycola (approved):
Seems to work for me! The `set_cgyro_legacy` flag should be called before writing any files if one wants the new behaviour.

PR #529, opened 2026-02-21
Improving pyroscan error handling

physycola:
Modified error handling to cleanly handle the first point in a pyroscan not loading properly


bpatel2107 (approved):
Ooh this is clever, I like!

PR #531, opened 2026-03-03
Feature/real space plot 3d

bpatel2107:
A simple script to plot 3D visualisation of a flux tube. We do the fourier transform to real space and make the flux tube go from $-\pi$ to $\pi$ by appending on the end of the theta grid a phase shifted first value in theta. This makes the interpolation done for the mapping from field aligned to cartesian a lot easier.

The code currently plots at $\theta=0$, then a flux 3D torus of the data and then finally at a fixed toroidal angle $\zeta=0$ by interpolating onto that surface.

This does a simple mapping from the code's $y$ -> $\alpha$ -> $\zeta$ (at a given $\theta$) via

$\alpha = q \theta - \zeta$

However we currently don't account for the radial variation of $\alpha(r)$, $q(r)$ and the flux surface mapping with $\partial R/\partial r$ and $\partial Z/\partial r$. That is likely the only thing that is really missing. Essentially for each $x$ we need to expand the above expression to cover this. We do have $\partial \alpha/\partial r$ and $\partial \alpha/\partial \theta$ so it is possible.

Essentially the mapping from field aligned $(x,y,\theta)$ to cartesian $(X,Y,Z)$ is mostly right but could be improved

- Append data with final theta point (data doesn't include info `+pi` because its periodic)
- Fourier transform into real space using `xrft` $(k_x,k_y,\theta) \rightarrow (x, y, \theta)$
- Append $+L_y/2$ onto real space data (IFFT doesnt include final point back in)
- Map from $(x,\theta)$ to $(R,Z)$ or each radius $x$ - _doesn't include radial variation of flux surfaces_
- Map $y$ to corresponding toroidal angle $\zeta$ - _doesn't include radial variation of_ $\alpha$
- Interpolate onto requested points (using periodicity in $y$ to get any toroidal angle)

Added a ChatGPT generated docs on this. Accidentally branched from the wrong branch though so a kinetics doc is in there...


docs/howtos/real_space_plotting.rst
```diff
@@ -0,0 +1,387 @@
...
+   \rho = \rho_0 + x \, \rho_\mathrm{ref} \, \rho_*
+
+We then evaluate the flux surface:
+
+.. code-block:: python
+
+    >>> x = rs_phi.x.data
+    >>> theta = rs_phi.theta.data
+    >>> rho0 = pyro.local_geometry.rho
+    >>> rhostar = 0.005 * pyro.norms.lref / pyro.norms.rhoref
```
d7919, line 138: 0.005 being a placeholder rho* value?

bpatel2107: Yes, could clarify this or set it as an input

bpatel2107:
Will merge for now and fix problems later on!

PR #533, opened 2026-03-12
Updated field line following diagnostics

bpatel2107:
This PR updates/adds field line following diagnostics `FieldLine` which calculates all things related to field line integration stuff like 

- Poincare maps
- Linear field line tearing
- Radial displacement
- Parallel correlation lengths
- Radial diffusion coefficients

It does this in a unit aware method so works in any convention.

I have added standard field line integrator `follow_field_line` which all the diagnostics that need it use which has several integration schemes from central differences to RK4.

There are some docs that describe that the code does too.


bpatel2107:
Updated the tests so will merge now

PR #535, opened 2026-03-12
Feature/simdb database

bpatel2107:
This write a Yaml file that can be ingested by SimDB for a gyrokinetic database. This stores the following data in the yaml which should allow for a searchable database

- [x] GK input file names
- [x] GK output file names
- [x] IMAS IDS (generated by pyro)*
- [x] LocalGeometry
- [x] LocalSpecies
- [x] Numerics
- [x] Convention
- [ ] **Convergence**
- [ ] _GKOutput_

Note the last two need to worked on. I have added a linear diagnostic `Convergence` that does a series of convergence tests for a linear sim, which are basically one useful number to describe something (like the field end ratios which show the ratio of a field at the end to its peak, the number should be small). 

Something simulation should be done for nonlinear simulations though that should be discussed (like standard deviation from average, peaks of spectra, grid scale oscillations). These should all eventually be added.

Further what GKOutput data should be added to this. Maybe time averages (or final time slice for linear) of 

- Fluxes (broken down by field)
- Field amplitude (averaged/integrated)
- Eigenvalues (linear)
- Flux ratios

Many more could be added but the aim of this is for searchable data so we shouldn't go crazy.

@FelixWattsYork apparently I already added the jacobian into here for and forgot about it...

Note IDS is generated by `idspy-toolkit` and not `imaspy` so we do need to switch over at some point!


bpatel2107:
Will merge for now. If further SimDB support is needed then that can be updated. @FelixWattsYork this will push the jacobian into `GKOutput`

PR #536, opened 2026-03-13
Bugfix/nonlinear pyroscan coordinates

FelixWattsYork:
There was a bug in my recent PR, which meant that coordinates didn't have the associated values. Basically ky was there but you could only index it by position not value. I have added the proper coordinate adding into the code, and updated the test to check for the correct behaviour

### DIFF
--- a/src/pyrokinetics/pyroscan.py
+++ b/src/pyrokinetics/pyroscan.py
@@ -91,7 +91,6 @@ def select_kx_ky_time(
         time_mode=time_mode,
         tolerance_time_range=tolerance_time_range,
     )
-
     if sum_ky and "ky" in da.dims:
         da = da.sum(dim="ky")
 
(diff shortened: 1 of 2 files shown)


bpatel2107:
I think this is not preserving the units of the simulation. You might need to add the units in as an attribute

EDIT I mean the units of the coordinates

bpatel2107:
See these lines here

https://github.com/pyro-kinetics/pyrokinetics/blob/2e238643779756d57e93eac832d5a9600ea1a562/src/pyrokinetics/pyroscan.py#L405-L407

For older versions of xarray you couldn't assign units directly to the coordinate so you would have to put the unit as an attribute.

However in a newer version now you can so you could force a newer version of xarray where the coordinate has the unit or give it the attribute as above.

bpatel2107:
@FelixWattsYork I've pushed a temp fix that works for now. Feel free to match the implementation you currently have though and tweak away.

Would be worth adding a test to make sure the units are kept.

@dake0795 checkout this branch please and give it a go?

FelixWattsYork:
> @FelixWattsYork I've pushed a temp fix that works for now. Feel free to match the implementation you currently have though and tweak away.
> 
> Would be worth adding a test to make sure the units are kept.
> 
> @dake0795 checkout this branch please and give it a go?

Just adding the test now

bpatel2107 (approved):

PR #538, opened 2026-03-16
Feature/gs2 gp

FelixWattsYork:
### Implementation of Will's Gaussian Regression Model into Pyrokinetics

Allows for Wills GS2 surrogate to be loaded into an queried in Pyrokinetics https://arxiv.org/abs/2403.13006
It more generally allow any pytorch objects to be loaded and their output read, providing they have the same input list. However this is specifically aimed at will's GP surrogate, to allow for easier comparison with existing gyrokinetic simulations.
This model is designed to be used with Maurizio Giacomin's saturation rule, however this has not been implemented yet and will come later


bpatel2107:
@FelixWattsYork  
Can you merge unstable back into this branch to the Files Changed tab is tidier

PR #543, opened 2026-03-30
Fix: persist normalisation reference values in PyroScan

FelixWattsYork:
# AI TEST!
Test of Claude Codes ability to fix real world bugs in pyrokinetics, on recommendation of Francis. Will not be merged until Code changes have been reveiwed and real world impact has been assessed

## Summary

- **Bug**: When a `PyroScan` is created from a code that stores normalisations (e.g. GENE) and then converted to a code that cannot (e.g. TGLF), reloading the scan from disk via `load_base_pyro=True` lost the normalisation data. This caused failures when scan parameters carried units tied to those normalisations (e.g. `gamma_exb` in `rhoref` units).
- **Fix**: `PyroScan.write()` now saves `pyroscan_norms.json` alongside `pyroscan_base.input`, and `PyroScan.__init__(load_base_pyro=True)` automatically restores those reference values when the file is present. Existing scans without the norms file continue to work unchanged.
- **Tests**: Added `test_norms_persisted_across_write_load` which creates a pyro from equilibrium+kinetics data, converts to TGLF, writes a PyroScan, reloads it, and verifies all reference values match.

## Test plan

- [x] New test `test_norms_persisted_across_write_load` passes
- [x] All 15 existing `test_pyroscan.py` tests pass
- [x] Manual verification with a real GENE → TGLF scan workflow

🤖 Generated with [Claude Code](https://claude.com/claude-code)


FelixWattsYork:
# AI Review!!

After trying to use Claude Code to solve this bug, here is my review of it's progress

## The Good
- Quickly generated Code, to do as I explained, including tests
- Fixed my spelling mistakes
- Generated PR with documentation without any interaction on my part, easily integrated with Github
- Was flexible when error were pointed out

## The Bad
- Initial tests written contained excess checks, testing for implementation not outcome
- warnings were not included with exceptions in the code by default
- Comments were pretty verbose

## The Ugly
- Didn't spot the second error in the code, that the reference values in the pryoscan.json reference the original input file, not the pyroscan_base.input that gets given the normalization, so it still doesn't work. Had to be prompted pretty heavily to see this issue and was not picked up in the tests the AI constructed
- Ran out of usage before the bug was fixed. I'm only using the pro plan, not the max plan, but that is still a paid for plan. This is a pretty big issue, if the AI can't correctly identify the actual bug and fix it within the base usage limit, then it is both slow and rather expensive. 

## Ultimate Score
4/10

Whilst I think this has potential for things like documentation, and visualization tools, I think it's bug fixing ability of pre-exisitng code has limits. If the developer is able to write thorough tests including the error then the AI is still useful, but it therefore requires strong supervision. In addition any AI generated code comes with the issue of readability. Whilst the code Claude generated is very simple, it is very capable of generating more complex code. However this demonstrates that it can happily generate "fixes" that don't solve the full issue, but may seem too, introducing further bugs into the Code base.

Ultimately it is up to the new king of Pyro @MantasA411 to decide regulations regarding AI generated code an PRs but unfortunately I don't think it is going to speed up Pyro Development by orders of magnitude
:(

FelixWattsYork:
It goes without saying the bug isn't fixed, there is still and issue with the pyroscan.json referencing the units in the original file the prysocan was generated from, even if that file isn't the base file anymore

MantasA411:
I think you should know what is in the code you want to merge, so not just copy paste the AI answer, but if you find AI helpful then it's a good idea to use it to speed up your work, it's a new tool and potentially very useful. In general, whenever an update is made there is potential for introducing new bugs, but that's the beauty of tests, so just need to make sure we keep those up to date.

FelixWattsYork:
Ok, I think I have (claude has) implemented as fix. There is a new method in units which allows for conversion back from physical units to simulation units. There are hooks for numeric e.c.t that call this method. That might be a little overkill as this is currently only needed for this one specific use case, but I thought it best to leave in the hooks for metrics, local geometry as well just in case they are needed in the future. However they can be removed.  

The parameter dict that the pyroscan is given will now be converted to simulation units immediately on construction. This simplifies things, however if the user gives a physical unit only for the parameters they are scanning over it will be lost (however I don' t know how they would be able to do that anyway)

The parameter dict will also always now be in the Pyro convention, not the convention of the code that the base pyro is! 

This is a big change as it will mean different names for all the files generated, however it is more consistent and will allow for intuitive conversions between codes.

physycola:
I am retiscent to approve the PR as there is a lot here I haven't had time to look over, but the parts relevant to the reference units seem to look okay. What testing have you done of this functionality? Would it be worth writing some "roundtrip" tests to check whether the reference normalisations are correctly persisted?

FelixWattsYork:
> I am retiscent to approve the PR as there is a lot here I haven't had time to look over, but the parts relevant to the reference units seem to look okay. What testing have you done of this functionality? Would it be worth writing some "roundtrip" tests to check whether the reference normalisations are correctly persisted?

The test_norms_persisted_across_write_load is essentially a roundtrip test, converting GENE to TGLF and then checking that the resulting Reference values are preserved. The last tick box is essentially just to test this manually

FelixWattsYork:
Also, I think one the template files I used was a MAST file by accident containing MAST geometry. I think that is public anyway but too be sure would it be best to squash and merge so it is not in the git history?

physycola:
Not sure about this one. @MantasA411 , @dake0795 , what's the best practise here? 

@FelixWattsYork , does it work manually? What does codecov report?

FelixWattsYork:
> Not sure about this one. @MantasA411 , @dake0795 , what's the best practise here?
> 
> @FelixWattsYork , does it work manually? What does codecov report?

Having tested it a lot manually it works fine. It's gotten to the point now where the bugs I'm running into are in the main branch of pyro and not this pr so I'm pretty confident. I'm haven't looked at codecov yet. I would say that any attempt to have 100% code coverage with tests whilst useful it probably a bit too ambitious with the current team. 

Basically I think this branch is definitely ready to merge

bpatel2107 (approved):
Nice, this is a good feature and a cool use of Claude

PR #544, opened 2026-04-02
Created Enforce Consistent Parallel Velocity Gradeint Function

FelixWattsYork:
# Enforce Parallel Velocity Gradient Function for ExB Scans
The Parallel Velocity Gradient (PVG), or Parallel Velocity Shear is an important gyro kinetic drive term. The effects of PVG are analogous to the Kelvin Helmholtz instability. 

Since for large V_ExB velocity orderings V_ExB = V_toridial, the ExB shear rate is linked to the Parallel Velocity Gradient. Certain gyrokinetic codes, such as GENE, only have an input for ExB shear and calculate the PVG based on that, however other codes like TGLF have PVG as a separate input parameters. This means that if a user want to do an ExB scan with a code such as TGLF, they would have to update the PVG value in lock step with ExB value manually.

This PR adds a new function, Enforce PVG, which automatically updates PVG values in line with ExB. This is similar to the function, enforce beta prime which has a similar purpose when updating beta in simulations.

The equation for ExB Shear is 
$ExB = -SIGN(I_{tor})\frac{r}{|q|}\frac{\partial}{\partial r}(\frac{V_{ExB}}{R})\frac{a}{c_s}$
and the equation for Vpar is 
$PVG = -SIGN(I_{tor})R_{maj}\frac{\partial}{\partial r}(\frac{V_{tor}}{R})\frac{a}{c_s}$

This gives the relation 
$PVG = \frac{|q|R_{maj}}{r} ExB_Shear = \frac{q}{\rho} ExB$

using that
$\Omega = \frac{V_{tor}}{R_{Maj}}$
this can be rewritten as

$\frac{d \Omega}{d \rho} = \frac{-q}{\rho} ExB$

This code was validated by comparisons to GENE's internal PVG calculations, by comparing the PVG values from an ExB scan of TGLF simulations with enforce PVG to a ExB scan of GENE then converted into TGLF


FelixWattsYork:
This also adds gamma_exb as a parameter key for updating ExB shear in pyroscans, which is a very natural addition

MantasA411 (approved):
Looks good to me.

PR #545, opened 2026-04-10
Improved pyroscan error handling (again)

physycola:
The new implementation of `pyroscan.py` did not properly handle runs that failed to reshape correctly, which can sometimes happen with large batches of runs. I believe that this gives more robust error handling.

### DIFF
--- a/src/pyrokinetics/pyroscan.py
+++ b/src/pyrokinetics/pyroscan.py
@@ -752,9 +752,7 @@ class PyroScan:
             buffers[name] = [ref * np.nan if x is None else x for x in arrays]
 
         if all(all(x is None for x in arrays) for arrays in buffers.values()):
-            raise FileNotFoundError(
-                "Unable to load any gk_output files in this scan"
-            )
+            raise FileNotFoundError("Unable to load any gk_output files in this scan")
 
         ds = xr.Dataset(parameter_dict)
         for name, arrays in buffers.items():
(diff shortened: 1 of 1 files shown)


FelixWattsYork:
Ah I see, When I change Pyroscan I ran into issues with how to handle creating a buffer is the first Pyro input file wasn't readable. This seems a good solution.  My only request would be if you are able to create a test containing the original scenario that was failing. I wasn't aware of any good reason that you would have a pyroscan where the first item wasn't readable (hence the original error handling)

physycola:
@FelixWattsYork , the failure state is unrelated to the `pyroscan.py` logic. In my case, it was related to the failure of `cgyro.py` to reshape the loaded `gk_output` data properly, likely due to a cluster-related IO failure. I am not really sure how to artificially manufacture such data, but more importantly, I don't think that falls under testing for `pyroscan.py`?

bpatel2107:
Making a test for this although possible is a bit difficult given that you need broken data to begin with and we would need to construct that, which can get messy. What we really need is to have a separate repo with a bunch of test data which can then contain this broken data, rather than including it here.

Maybe a separate zenodo/github repo. This is a longer term thing and should not really hold up this PR being merged. Can discuss tomorrow

bpatel2107 (approved):
LGTM but we do need to consider a separate testing repo.

PR #546, opened 2026-04-20
Bugfix: Fixed pyroscan reading failing if pyro output doesn't include all expeted fields

FelixWattsYork:
Previously, pyroscan would try and read all values that it tries to load, e.g."growth_rate" frequency, and if these are not in the pyro gk output then it will fail. This fixes this bug by checking if the spec is in the gk_output before pyroscan trys to add it.


MantasA411 (approved):
Will need to check if there exists a similar issue anywhere else in the pyroscan but this looks good.

PR #547, opened 2026-04-20
Adding loading of new CGYRO momentum outputs

physycola:
This adds functionality to allow for the loading of the new `CGYRO` output files that add a significant breakdown of the momentum fluxes (see [this branch](https://github.com/gafusion/gacode/tree/momentum_output)) into parallel, perpendicular, and electromagnetic components. It appears to be working as intended (on both the `CGYRO` and `pyro` sides), which I've tried to illustrate below.

In all of the following plots, "Initial" refers to a run with `CGYRO` where the electromagnetic momentum fluxes have been _manually_ disabled in the code (for the standard momentum fluxes), while "Final" refers to the case with them turned restored. "New Outputs" are the results extracted from the new output file, which shows excellent agreement with "Final". One advantage of the new outputs is that they allow the "Initial" state to be reconstructed directly, which we denote with "Reconstructed". I have also included a figure showing the parallel/perpendicular slicing that can now be done with the outputs. 

Let me know what you guys think....

### Linear growthrate and frequency
growthrates_and_frequencies

### Eigenfunctions
eigenfunctions

### Fluxes
fluxes

### Momentum flux comparison
momentum_flux_comparison

### Momentum flux breakdown
momentum_component_breakdown


src/pyrokinetics/gk_code/cgyro.py
```diff
@@ -1847,3 +1853,103 @@ def _get_eigenfunctions(raw_data: Dict[str, Any], coords: Dict) -> np.ndarray:
...
+            mflux, **pyro.gk_output._downsample
+        )
+        mflux = np.swapaxes(mflux, 0, 2)
+
+        # Bootstrap normalisations from standard momentum-flux data
+        reference = pyro.gk_output["momentum"]
+
+        raw_sum = mflux[:, :2, ...].sum(axis=1)
+
+        scale = reference.pint.dequantify().data / raw_sum
```
bhavin-hash, line 1939: Sneaky, I like it!

bhavin-hash (approved):
LGTM

dake0795 (approved):
Looks great. I ran the branch I was writing in parallel and get the same answers.

PR #548, opened 2026-04-21
Bugfix/claude/pyroscan netcdf units

FelixWattsYork:
Draft, not quite ready yet


FelixWattsYork (commented):
Actually a very small dif from Pr #543 only changes are in pyroscan and pyroscan test, where there is extra code to convert all pyroscan units and load normalization before reading netcdf. The test basically constructs a new cgyro input file with normalization and then create a netcdf that the loading of is trialed.  I think this is also good to merge as I haven't had any problems with it when using it myself

src/pyrokinetics/pyroscan.py
MantasA411, line 767: Not really this PR but this seems to be an identical if statement to the if block below, so this will be overwritten, should this be removed?

PR #551, opened 2026-04-23
Add stella v1 input file support

dake0795:
▎ - Add stella v1 namelist format support (auto-detected read; v1 is the default on write)            
  ▎ - Retain backward-compatible reading of legacy and pre-v1 formats
  ▎ - Fix y0 sign bug (v1 requires y0 > 0 as box size; pre-v1 kept as-is)                               
  ▎ - Add CGYRO→stella example showing how to write each format                                         
  ▎ - Add howto documenting the three format versions                                                   
  ▎ - Add test cross-validating v1 reading against stella's own convert_inputFile.py                    
  ▎                                                                                                     
  ▎ Test plan                                                                                           
  ▎                                                                                                     
  ▎ - pytest tests/gk_code/test_gk_input_stella.py                                                      
  ▎ - pytest tests/gk_code/test_gk_output_reader_stella.py


MantasA411:
I will try and figure out why the tests are failing, they probably need to be updated.

MantasA411:
I think this is all fixed now and ready to merge unless anyone else sees any issues.

`input.stella` file in the templates is still the old 'PRE_V1' version and `input.stella_v1` is the new version. Would it be useful to make `input.stella` the new version and rename the old one as `input.stella_pre_v1 `?

MantasA411:
`input.stella` is now v1,`input.stella_pre_v1` is pre_v1. I added a warning to tell the user which version is used, just in case they have a script that uses input.stella but expects a pre_v1 version, but not really sure it's necessary

MantasA411:
@dake0795  do you know if `tend` parameter has to be an integer in stella files? AI says not necessarily

The tests starting failing (even going back a few commits and running tests they failed, no idea how). I managed to get it to pass but i had to change `self._set_numerical_var("tend", int(numerics.max_time.m))` to `self._set_numerical_var("tend", numerics.max_time.m)`, that is I removed the integer requirement. This is needed because the cgyro file used for the test has a time of `0.25`. Will push the commit soon.



Again, still a mystery to me why the tests would not pass when running pytest locally even when i created a worktree based on a commit, which according to the history seemed to pass CI.

PR #552, opened 2026-04-24
Bugfix/eliteinp pedestal

MantasA411:
Improved fitting of equilibrium for eliteinp files


hgd511 (approved):
Looks good, has been working well for me since end of Jan. The time to load up the pyro object from the eliteinp has increased, but the accuracy is spot on.

PR #553, opened 2026-04-24
Feature/gene flux spectra

dake0795:
Adds functionality to construct GENE fluxes as functions of kx and ky based on fields and moments data.


bpatel2107:
This looks good! But why do we need a new `FluxSpectra` object. Is this not already covered in the `Fluxes` object but with more dimensions? Would it not make more sense to have a flag like

`pyro.load_gk_output(kxkx_flux_spectra=True)`

Which would just run your file reader instead of the default one? A new object for this already existing quantity seems like overkill.

MantasA411:
Does the fact that `flux_time` used in `FluxSpectra` is different from time in `Fluxes ` make it useful to have a separate object? Also the new object has two quantities `particle_es `and `particle_em`, to make it consistent with fluxes we'd need to make it into a single quantity with the field coordinates (phi, apar, or create new ones) separating es and em contributions.

`load_flux_spectra ` already exists in this branch which loads these fluxes  and is False by default when loading gk output.

PR #554, opened 2026-04-28
Fix load_gk_output crash on unitless scan parameters

FelixWattsYork:
PyroScan.load_gk_output called .magnitude and .units on every entry of parameter_dict, raising AttributeError when a scan included a parameter without pint units (e.g. TGLF WIDTH/NBASIS_MAX, integer ntheta, or any parameter pyrokinetics did not auto-attach units to). Plain arrays now fall back to ureg.dimensionless.


bpatel2107 (approved):
LGTM

PR #559, opened 2026-05-08
Removed old field line following and use new one for tests

bpatel2107:
Removed old field line following algorithm and ensured tests for correlation/displacement use new method


MantasA411 (approved):
LGTM

PR #560, opened 2026-05-08
Draft:  Add MEGPy-backed MXH fitting

fcasson:
This MR adds a MEGPy-backed MXH surface-fit path to Pyrokinetics for GEQDSK equilibria via `load_local_geometry(..., local_geometry="MXH", fit_backend="megpy")`.

The practical benefit is a better reconstruction of the traced flux surface than the current native Pyro MXH fit, while keeping Pyro’s existing MXH object and downstream workflow. In the current JET 84793 case, the MEGPy-backed path reduces the surface-fit error from `RMSE/a = 9.19e-03` to `6.13e-03` at `rho_tor = 0.96`, and from `9.79e-03` to `7.31e-03` at `rho_tor = 0.98`, relative to native Pyro MXH.

The attached comparison plot shows the traced surface, native Pyro MXH, and the MEGPy-backed MXH result at both radii.  The MEGPy backend already improves the reconstructed MXH surface shape relative to native Pyro MXH. We also tested passing MEGPy-derived radial derivative information into Pyro, but on the JET test case that richer seed is not yet a better starting point for Pyro’s internal derivative fit, so Pyro falls back to its existing initialization. In practice, this MR improves the surface representation first; derivative-fit improvements remain future work.

The current implementation is not a “MEGPy direct” geometry path. It keeps Pyro’s MXH object and downstream workflow, but replaces the surface-decomposition stage with a MEGPy MXH fit. For the JET 84793 case this improves the reconstructed surface relative to native Pyro MXH, but it does not yet reach the accuracy of the MEGPy direct fit itself. In other words, the MR closes part of the gap to MEGPy direct while preserving Pyro compatibility, and the remaining error appears to come from the handoff into Pyro’s MXH representation rather than from MEGPy’s surface fit.

In the JET 84793 case, the MEGPy-backed MXH path improves the reconstructed surface relative to native Pyro MXH corresponding to 33% and 25% reductions in error. The result is still less accurate than the MEGPy optimized MXH fit itself so the current implementation closes part, but not all, of the gap to a “MEGPy direct” result. The geometry-only handoff and the full Pyro+MEGPy result are almost the same, which shows that the current improvement mainly comes from better surface coefficients, not from the derivative-fit step.

| rho_tor | Option | RMSE/a | Relative to Pyro native MXH | Comment |
| --- | --- | ---: | ---: | --- |
| 0.96 | Pyro Miller | 4.893648e-02 | 5.33x worse | Much worse than any MXH-based option |
| 0.96 | Pyro native MXH | 9.189222e-03 | 1.00x | Baseline |
| 0.96 | Pyro MXH + MEGPy backend | 6.134152e-03 | 0.67x | 33.2% lower than native Pyro MXH |
| 0.96 | MEGPy handoff geometry only | 6.130138e-03 | 0.67x | Essentially identical to full Pyro+MEGPy result |
| 0.96 | MEGPy optimized Miller | 3.651250e-02 | 3.97x worse | Better than Pyro Miller, but still far worse than MXH |
| 0.96 | MEGPy optimized MXH | 1.479244e-03 | 0.16x | Best fit; 83.9% lower than native Pyro MXH |

| rho_tor | Option | RMSE/a | Relative to Pyro native MXH | Comment |
| --- | --- | ---: | ---: | --- |
| 0.98 | Pyro Miller | 5.649571e-02 | 5.77x worse | Much worse than any MXH-based option |
| 0.98 | Pyro native MXH | 9.793984e-03 | 1.00x | Baseline |
| 0.98 | Pyro MXH + MEGPy backend | 7.306036e-03 | 0.75x | 25.4% lower than native Pyro MXH |
| 0.98 | MEGPy handoff geometry only | 7.300899e-03 | 0.75x | Essentially identical to full Pyro+MEGPy result |
| 0.98 | MEGPy optimized Miller | 4.235874e-02 | 4.33x worse | Better than Pyro Miller, but still far worse than MXH |
| 0.98 | MEGPy optimized MXH | 1.947325e-03 | 0.20x | Best fit; 80.1% lower than native Pyro MXH |

The RMSE metric is a surface-shape metric only: it measures agreement of the reconstructed flux surface trace with the reference traced surface. It does not validate the radial derivative terms directly. 

image


fcasson:
Current MHX fits inside PYRO after first commit

image

Megpydirect fits outside pyro

image

Next goal is to make Pyro as good as Megpy direct

fcasson:
After fixing #562, we have very little difference between pyro MXH and megpy in pyro MXH, so I don't know if this is really worth keeping.  Interesting though that the Megpy Miller Fit is much better, Pyro could copy that.

bpatel2107:
@fcasson can you post the updated comparison to compare? Miller is far less constrained as a representation so it is not too surprising that there is more wiggle room.

My understanding is that MegPy takes adjacent surfaces to calculate derivatives?

Pyro attempts a completely local approach. The issue is that we have two unknowns $\frac{\partial R}{\partial \rho}$ and  $\frac{\partial Z}{\partial \rho}$ but currently only 1 function to fit it to $B_\theta$. I am curious how well it matches the poloidal field? I guess technically there could be multiple solutions to this but because MXH is so constrained it generally works out. I don't know how to prove that mathematically

I can't really think of a better way to obtain another constraint on the equilibrium from local parameters but we likely would have come across it by now? I know we've discussed this before but @hgd511 thoughts?

hgd511:
I also don't know another quantity that could be used to constrain $\partial R / \partial \varrho$ and $\partial Z / \partial \varrho$ on a single surface other than $B_{\mathrm{pol}}$. 
I think a potential use case for MegPy here is that it may be able to better fit derivatives close to the LCFS, where the $B_{\mathrm{pol}}$ method is currently poor for all eq files other than eliteinp. Provided this is the case, it would be nice to see a correspondence between MegPy and Pyro on a given eliteinp close to the edge.

fcasson:
Can you provide an eliteinp for me to test on?

PR #561, opened 2026-05-09
Classify TGLF Miller-family geometry correctly

fcasson:
This MR fixes Pyrokinetics’ interpretation of TGLF geometry flag 1.

Native TGLF uses `GEOMETRY_FLAG = 1` for the full Miller-family path, including both ordinary Miller geometry and MXH-extended shaping. Pyro previously treated flag 1 too directly as “MXH”, with only a narrow fallback to Miller. This MR changes the reader logic so flag 1 is handled as “Miller-family”, and Pyro now distinguishes `LocalGeometryMiller` vs `LocalGeometryMXH` from the presence of extended shaping terms in the input.

This does not change the native TGLF input file generation; it fixes Pyro’s post-read classification, round-tripping, and related tests so the interpretation matches TGLF semantics.

Co-authored-by: Github Copilot (GPT-5.4)

### DIFF
--- a/src/pyrokinetics/gk_code/tglf.py
+++ b/src/pyrokinetics/gk_code/tglf.py
@@ -268,15 +268,11 @@ class GKInputTGLF(GKInput, FileReader, file_type="TGLF", reads=GKInput):
             convention = getattr(norms, self.norm_convention)
 
         tglf_eq_flag = self.data["geometry_flag"]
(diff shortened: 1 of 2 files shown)


fcasson:
I'm not sure about this one suggested by the AI.  It might be too much semantics...

bpatel2107 (approved):
LGTM

PR #558, opened 2026-05-08
Add ITERDB kinetics reader

fcasson:
Co-authored-by: Codex (GPT-5.5)

PR #562, opened 2026-05-09
Map ITERDB RHOTOR via equilibrium psi_n

fcasson:
Fixes error in #558 

This MR corrects ITERDB radial-coordinate handling when an equilibrium is provided.

Previously the reader assumed `psi_n = RHOTOR^2`. That is not generally correct for the supplied equilibrium, and it can shift the loaded kinetic profiles to the wrong radius. This change instead maps ITERDB `RHOTOR` onto the equilibrium `psi_n` grid using the equilibrium’s `rho_tor(psi_n)` relation, while keeping the previous fallback for cases where no equilibrium is passed.

A focused regression test is included for the equilibrium-mapped path.

Co-authored-by: Github Copilot (GPT-5.4)

### DIFF
--- a/src/pyrokinetics/kinetics/iterdb.py
+++ b/src/pyrokinetics/kinetics/iterdb.py
@@ -156,43 +156,41 @@ class KineticsReaderITERDB(FileReader, file_type="ITERDB", reads=Kinetics):
         vrot_profile = _profile(blocks, "VROT", time_index, time, required=False)
 
         if eq is None:
-            electron_psi_n = te_rhotor**2 * units.dimensionless
-            ion_temp_psi_n = ti_rhotor**2 * units.dimensionless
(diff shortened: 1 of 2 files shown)


bpatel2107 (commented):
Overall this look good to me! Do we want to still support loading iterDB files without Equilibrium objects? At least a warning should be raised

fcasson:
Yeah I think I agree, the fallback is too dangerous:

>  while keeping the previous fallback for cases where no equilibrium is passed.

MantasA411:
I think it should probably error without equilibrium, just like some other formats (pfile I think). Could consider some flag that a user can set explicitly, which  allows it to run with 'no equilibrium' . It would only be useful if running tests maybe. 

I am running the tests and checking some stuff on this, will merge it once that's done.

MantasA411 (approved):
Looks good.

PR #563, opened 2026-05-13
Update dimensionality of mass to mref

bpatel2107:
In pyro there are two classes of units, simulation (`vref`, `tref` `nref` etc) and physical units (all in SI) for each of the reference values. In general it is possible to convert between simulation units without issue but when converting to SI, you need to specify a context under which you could transform sat `tref` to `keV`

However, somehow the simulation reference masses were defined in `kg` rather than making up a an dimension of `mref`. This means that before it wasn't possible to do the following

`(1 * ureg.vref_nrl).to((ureg.tref_electron / ureg.mref_deuterium)**0.5).m`

This corrects that so we now have a proper `mref` dimension which can be used to convert velocities.

Also allows for converting `rhoref` to its constituent units.

### DIFF
--- a/src/pyrokinetics/normalisation.py
+++ b/src/pyrokinetics/normalisation.py
@@ -838,7 +838,7 @@ class SimulationNormalisation(Normalisation):
             add_deuterium = True
             hyrogenic_mass = kinetics.species_data["hydrogenic"].get_mass()
             deuterium_factor = (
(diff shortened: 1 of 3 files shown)


MantasA411 (approved):
This looks great. I'm surprised this never came up as an issue.

Also nice to have  the electron_mass defined explicitly in `self.define(f"electron_mass = {physical_constants['electron mass'][0]} {physical_constants['electron mass'][1]}"  )`

PR #565, opened 2026-05-20
initial commits to change averaging

FelixWattsYork:
Changes time avering


FelixWattsYork:
This is the branch that averages of time for the linear results

PR #568, opened 2026-05-28
Feature/gene flux spectra

MantasA411:
This is a fresh start for this feature. The previous PR (#553) was automatically closed due to a bad history force-push from another machine. This PR contains the correct history and is ready for review.


MantasA411:
Re structured it  so that it uses the already defined Flux object. The documentation example seems to work. Made use of AI, the results seem reasonable, but not sure if there are any better benchmarks to test this with.

bpatel2107 (approved):
This looks good to me!

PR #569, opened 2026-05-29
Bugfix/duplicate self update

MantasA411:
Fix double application of `update_self_parameters `causing incorrect `add_parameter_func `behaviour. 

`self.update_self_parameters()' calls user-defined functions, calling it twice results in double application of user defined functions, which may give an incorrect scan output.

Removed the duplicate and added a test to make sure a function is only applied once.

Thanks @physycola for reporting this.


physycola (approved):
Looks good to me. Thanks for adding a test to prevent such a future regression!

PR #570, opened 2026-06-01
Bugfix/claude/pyroscan netcdf units

FelixWattsYork:
Repeat of old one


FelixWattsYork:
This is the one that needs to be merged

MantasA411:
Original PR (#548) was closed due to a bad force-push.

MantasA411:
@FelixWattsYork The pyroscan example works without this fix. Can you make a small example where this fix is used or do you need a full NL simulation data set to do so?

FelixWattsYork:
> @FelixWattsYork The pyroscan example works without this fix. Can you make a small example where this fix is used or do you need a full NL simulation data set to do so?

I could but it would be quite tedious, it would involve saving, moving and reloading an example with units. I still need to go through the code to make sure there isn't any duplicate functions. But I've used it extensively and it does work

MantasA411 (approved):
I think this works, and fixes the bug that was introduced.

PR #571, opened 2026-06-02
Feature/pyroscan linear time average

FelixWattsYork:
This is an important bugfix for gs2 linear simulations, but probably very overbuilt


d7919:
Could you explain what the bug is that is being fixed here? The branch name, PR description and commit messages each seem to refer to different topics?

PR #572, opened 2026-06-02
Initial Commits by Claude, will be looked over, Fixes sum_ky not work…

FelixWattsYork:
…ing correctly with pyroscan

### DIFF
--- a/src/pyrokinetics/pyroscan.py
+++ b/src/pyrokinetics/pyroscan.py
@@ -78,13 +78,17 @@ def reduce_time(
 def select_kx_ky_time(
     da,
     *,
-    kx_min,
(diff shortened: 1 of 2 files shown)


MantasA411:
Origional PR (#550 ) was closed sue to a bad force push.

Original comment:

"In the Current implementation, ky is still incorrectly sliced even if sum_ky is put to false. This fixes this and also allows for sum_kx to be set to false. This is a pretty easy change and looks good but again hasn't been properly tested

IMPORTANT
What is branch fails to do is account for the Jacobin when summing over ky. If this is now available in gk_output this should be implemented before the branch is pushed"

PR #573, opened 2026-06-02
Feature/gs2 gp

FelixWattsYork:
Added will's gp, however definetly buggy and still needs work


MantasA411:
Original PR (#538) was closed due to a bad merge.

PR #574, opened 2026-06-02
Feature/gene restarting reading better

FelixWattsYork:
Automatically loads and concatenates the GENE reload outputs


MantasA411:
Original PR (#549 ) was closed due to a bad force push.
Original description form @FelixWattsYork :

A Branch to allow Pyrokinetics to automatically recognize and load multiple restarts of a single gyro kinetic simulation. The loaded results are automatically concatenated in the time dimension. Please not that if the simulations do not flow smoothly from one to another this will mean that the time indices will not be evenly spaced. This Functionality is toggle able with the flag load_restarts.

The Functionality I mostly made myself as it's pretty simple but this has not been extensively tested and the tests have not check yet, there this branch is still in the draft phase

PR #575, opened 2026-06-02
Bug fix/gene minor r check

FelixWattsYork:
Can't remember but a bug fix


MantasA411:
Previous PR (#520) was automatically closed due to a bad force push.

This PR:

Fixes Minor Radius if set incorrectly.
If Minor Radius is equal to trpeps * Major radius it is meant to be the reference value but is not set as one. This sets it to one in this situation whilst raising a warning

MantasA411:
@bpatel2107 I am not 100% sure the fix Felix implemented or the fix I attempted actually fix it the right way, there is maybe some simpler way to check that `minor_radius ` is representing what it should be representing. As far as  I understand `trpeps*major_radius` represents the position of flux surface `r/L_ref` and `minor_radius should be the machine minor radius `a`?

bpatel2107:
This is more of a result from an old bug in Pyro. `minor_r` in GENE doesn't actually do anything but we used to use it to represent `a / Lref` (and when Lref = a this is 1) which was helpful in terms of converting between different units. When Lref = Rmajor then this should be `a / R` which is just the inverse aspect ratio.

But at some point we accidentally set it to `rho / R` which is just wrong. 

Keep in mind `trpeps = r /Lref`and `rmaj = R / Lref`

The issue is that we may not know what the aspect ratio is and it is technically possible that `r / Lref` just so happens to equal `a / R` by sheer coincidence. A warning is probably fine here but in that unlucky scenario things may go wrong.'

This would only impact very old gene input files generated by pyro or gene input files from some other source.

None if it actually changes the core simulation, just if you want to convert between reference length.

MantasA411 (approved):
LGTM

PR #577, opened 2026-06-10
Bugfix/load shape terms

ILexiconI:
Fixes erroneous local geometry calculation when loading gene input files that haven't specified s_delta and s_zeta, as rho has already been divided out of dsndr and dcndr near line 350.


d7919:
This looks like the right fix to me. Prior to this fix you observed that a round trip reading in and writing out a parameters file without s_delta and s_zeta would lead to changes in the parameters due to this bug. I assume you've confirmed that this no longer occurs and the file is unchanged?

ILexiconI:
Yes, I have checked this eliminates the issue by loading and rewriting a parameters file that was previously coming out incorrect for the aforementioned reason.

MantasA411 (approved):
Looks good to me, I wrote a test to check it and it seems to pass fine, don't think adding a test to the repo for this is needed though, so I'll merge it the way it is.

PR #579, opened 2026-06-17
Bugfix/claude/pyroscan dirname rounding

FelixWattsYork:
If the rounding value is set to extremely, it can causing multiple runs to be set to the same file.  If the varying parameter dataset is finer than allowed with the rounding, this code reducing the rounding requirement


MantasA411:
Hey,

 I think the tests should use GENE as default not GS2 because then kys set are the actual kys used (not converted to pyro inputs). Also kys should be 0.123 and 0.124 as 0.126 rounds to 0.13 and gives no error.

More importantly the fix above doesn't actually pass the test, it only checks parameters within the run (so let's say ky and beta) not across different runs. 

Trying to understand it I (AI) wrote something which looks to work, I will push it, have a look to see if it passes any other tests, maybe I've misunderstood something.

PR #583, opened 2026-07-08
Feature/sonic rotation geometry

bpatel2107:
# Add sonic rotation: poloidally varying potential, centrifugal/Coriolis drifts and rotation-corrected geometry

## Overview

This PR adds support for *sonic* toroidal rotation to pyrokinetics. When the rotation
speed approaches the ion sound speed, the centrifugal force pushes ions to the outboard
side of the flux surface and a poloidally varying equilibrium potential $\phi_0(\theta)$
is generated to hold quasineutrality, pulling the electrons along with the ions
(Belli & Candy 2018 [1]; Casson et al. 2010 [2]). These effects are quadratic in the
Mach number and are neglected in the usual weak-rotation limit, but they matter at high
rotation and are essential for heavy impurity transport.

Three pieces are added:

1. `Diagnostics.calculate_sonic_quasineutrality` — solves for $\phi_0(\theta)$,
   $\partial\phi_0/\partial r$, $\partial\phi_0/\partial\theta$ and the effective
   pressure-gradient term $F_\rho$ on a flux surface.
2. `Diagnostics.generate_eik_geometry_terms` — produces the full set of eikonal
   geometry coefficients including the new rotation drifts, written in a form that slots
   directly into the gyrokinetic equation (see below).
3. `MetricTerms` — updated so the local equilibrium is consistent with the rotation
   modified (poloidally varying) effective pressure gradient when sonic rotation is on.

## Equilibrium potential $\phi_0$ and its derivatives

In a rotating frame the equilibrium density of species $s$ varies over the flux surface
(Belli & Candy Eq. 10):

```math
n_s(\theta) = n_{s0} \exp\left[ -\frac{z_s e\, \Phi_*(\theta)}{T_s}
+ \frac{m_s \omega_0^2 \left(R^2(\theta) - R_0^2\right)}{2 T_s} \right]
```

with $\Phi_*(\theta) = \phi_0(\theta) - \phi_0(0)$ and densities specified at the
outboard midplane $R_0 = R(\theta{=}0)$. The potential is found by enforcing
quasineutrality $\sum_s z_s n_s(\theta) = 0$ at every $\theta$, which for an arbitrary
number of species is a nonlinear algebraic equation solved here with Newton iteration
(as in CGYRO/NEO [1]).

The radial derivative $\partial\phi_0/\partial r$ is obtained from a second Newton solve
of the radially differentiated quasineutrality condition,
$\sum_s z_s\, \partial n_s/\partial r = 0$, which brings in the density, temperature and
rotation gradients. The poloidal derivative satisfies an equation *linear* in
$\partial\phi_0/\partial\theta$ and is evaluated in closed form:

```math
\frac{\partial \phi_0}{\partial \theta} =
\frac{\sum_s z_s n_s\, m_s \omega_0^2 R \,(\partial R/\partial\theta) / T_s}
     {\sum_s z_s^2 n_s / T_s}
```

For a pure plasma the solution reduces to the analytic result of Belli & Candy Eq. (9),
which is used to verify all three quantities in the new tests
(`tests/diagnostics/test_rotation.py`, scanning $\omega_0$ and
$\mathrm{d}\omega_0/\mathrm{d}\rho$).

Comparison of $\phi_0$, $\partial\phi_0/\partial r$ and $\partial\phi_0/\partial\theta$
against CGYRO at $M=3$ for GA-STD geometry:

image
image
image

### Poloidal density variation

The resulting equilibrium density asymmetry, for a pure hydrogenic plasma at $M = 3$ and for an
impurity plasma (carbon and tungsten, $n_{imp}/n_e = 1e-5$) at $M = 1$. The strong outboard
accumulation of tungsten ($\propto \exp(m_s \ldots)$) is the reason these terms dominate
heavy impurity dynamics [1, 2]:

image

image

## Rotation drift terms

`generate_eik_geometry_terms` now returns, alongside the standard
`gbdrift`/`gbdrift0`/`cvdrift`/`cvdrift0`/`gds*` coefficients, the sonic-rotation drifts:

| Term | Physics |
|---|---|
| `crdrift`, `crdrift0` | Coriolis drift, $\mathcal{O}(M)$ |
| `cfdrift`, `cfdrift0` | Centrifugal drift, $\mathcal{O}(M^2)$ |
| `phdrift`, `phdrift0` | Drift from the equilibrium potential $\phi_0$ ($\nabla\Phi_*$ modification to the $E\times B$ drift / electrostatic trapping) |

These are deliberately written in the same form as the normalised magnetic drifts of
Eq. (3.61), Section 3.7 of the Highcock thesis [3]
($\omega_{\nabla B N}$, $\omega^0_{\nabla B N}$, $\omega_{\kappa N}$,
$\omega^0_{\kappa N}$, i.e. `gbdrift`, `gbdrift0`, `cvdrift`, `cvdrift0`): each new
drift can simply be added to the total drift frequency multiplied by its respective
velocity-space moment, with the unlabelled/`0` pairs multiplying
$\nabla_N\alpha$ and $(k_{xN}/k_{yN}\hat{s})\nabla_N q$ respectively. This makes it
straightforward for any code already consuming GS2-style geometry coefficients to pick
up rotation support.

The returned terms can be written to a netCDF geometry file readable by other codes —
support for reading this file is being added to GX in [link to be added] , which will show the benchmark of linear growth rates etc. On the pyrokinetics side, `GKInputGX` now writes `geo_option = "nc"` and the `mach` parameter, and reads `mach`/`g_exb` back into `omega0`/`domega_drho`.

## Metric terms with rotation

With sonic rotation the magnetostatic equilibrium is modified: momentum balance replaces
$\nabla p$ with the effective pressure gradient
$(\nabla p)_\mathrm{eff} = \nabla p + \sum_s n_s m_s \mathbf{U}_0\cdot\nabla\mathbf{U}_0$
(Belli & Candy Eq. 13), which is *poloidally varying* because the equilibrium density
is. `MetricTerms` therefore now accepts `sonic=True` together with the flux-surface
resolved `mu0F_rho`, which replaces the flux-function `beta_prime / 2` in the local
Grad–Shafranov terms (Section 3.6 of the Dudding geometry paper <https://github.com/user-attachments/files/21363240/HDudding_Geometry_Paper.pdf>).
The pressure term in the poloidal-field integral is moved inside the $\theta$ integral
accordingly. With `sonic=False` (default) behaviour is unchanged.

## Tests

- `tests/diagnostics/test_rotation.py`: verifies `phi0`, `dphi0_dr` and `dphi0_dtheta`
  against the analytic pure-plasma solution (Belli & Candy Eq. 9, generalised to
  $T_i \ne T_e$ and including gradient terms) over a scan of
  $\omega_0 \in [0, 1]\, v_\mathrm{ref}/a$ and
  $\mathrm{d}\omega_0/\mathrm{d}\rho \in \{0, 0.7, -1.3\}$, agreeing to machine
  precision.
- $\omega_0 = 0$ is handled explicitly (all rotation terms identically zero) and the
  weak-rotation limit recovers the previous geometry output.

## References

[1] E. A. Belli and J. Candy, *Impact of centrifugal drifts on ion turbulent transport*,
Phys. Plasmas **25**, 032301 (2018). https://doi.org/10.1063/1.5020298

[2] F. J. Casson, A. G. Peeters, C. Angioni, Y. Camenen, W. A. Hornsby, A. P. Snodin and
G. Szepesi, *Gyrokinetic simulations including the centrifugal force in a rotating
tokamak plasma*, Phys. Plasmas **17**, 102305 (2010). https://doi.org/10.1063/1.3491110

[3] E. G. Highcock, *The Zero Turbulence Manifold in Fusion Plasmas*, DPhil thesis,
University of Oxford (2012), Section 3.7, Eq. (3.61). https://arxiv.org/pdf/1207.4419


bpatel2107:
Not 100% sure who should review this but its out there!

@fcasson @d7919 FYI may be of interest to you

fcasson:
I suggest you also read the often overlooked erratum to [2] and references therin.

* Casson Erratum: https://doi.org/10.1063/1.3491110
* Peeters Erratum: https://doi.org/10.1063/1.4751549
* Angioni and Veth:  https://doi.org/10.1063/1.4773051

bpatel2107:
Ah yes I did come across this when implementing in GX and was tripped up by it a little. 

The difference is that the Maxwellian gradient has to be taken at fixed total energy (thermal + potential + rotation which is conserved along a guiding centre orbit) but GK codes generally take derivatives with respect to fixed thermal energy (see defn below from Sugama 1998 - V0 here is actually omega0 in pyro...).

image

So you have to add extra terms for those additional gradients of the potential/rotation (see Sugama 1998 section 4) which gives the GK equation at in fixed total energy and then adjust that gradient to account for GK codes working with thermal energy.

The Belli paper gets the final correct result (equation 31) and below is a PDF covering this mapping from sugama to the belli formulae.

[sugama_belli_drive.pdf](https://github.com/user-attachments/files/29844787/sugama_belli_drive.pdf)

We could add this to pyro output but I actually added this to the GX source code. The issues is that this is a poloidally varying term that is species dependant. I could add another dictionary to the output if that is something people are interested in? Then in theory you could just read it in and just add it to your gradient term.

There is a benchmark of CGYRO and GX for this so I believe the implementation is correct (CGYRO has been benchmarked against GKW so hopefully all is well).

hgd511:
Looks good. As well as the density variation plots shown here, demonstration of agreement with say CGYRO for the metric term quantities, including the drifts, would be good.

Also as an extra feature, once the poloidal density variations have been calculated for inputted outboard midplane values, I think it would be valuable to include a calculation of the flux-surface average densities as these are typically what transport codes deal with.

PR #585, opened 2026-07-13
let me look through this

FelixWattsYork:
Attempt to make pyrokinetics compatible with the upcoming gyrofluid quasilinear flux model GFTM. This code is an evolution of TGLF, essentially just using GFS as the linear solver. As such that code reads input and outputs very similarly abide with very small syntax difference. This branch mostly just copies and renames the TGLF reading infrastructure (let me know @MantasA411 if this is the way you want to do it), with some small changes which I will detail in comments.

Importantly GFTM is not publicly available yet and therefore this branch can easily be checked. I will ask GFTM developer Dr Gary M Staebler, if it is possible to include example input and output files in the code, however the format is likely subject to change before the final release


FelixWattsYork:
Changes: 
src/pyrokinetics/gk_code/gftm.py line 1007:  takes the last float on the line rather than splitting the line based on ":" since GFTM uses ky = output, same for eingenvalues on line 1194
the helpers for this are at the end of the file

FelixWattsYork:
With Garry's Permission added example GFTM outputs for template input

